PR #1, opened 2015-08-27
Remove a number of #![feature] directives

alexcrichton:
Most had a simple workaround or weren't actually even used, so yay!

### DIFF
--- a/src/mem/epoch/garbage.rs
+++ b/src/mem/epoch/garbage.rs
@@ -1,7 +1,5 @@
 //! Data structure for storing garbage
 
-use alloc::heap;
-
 use std::ptr;
 use std::mem;
 use std::sync::atomic::AtomicPtr;
@@ -12,8 +10,7 @@ use std::sync::atomic::Ordering::{Relaxed, Release};
 /// Stores enough information to do a deallocation.
 struct Item {
     ptr: *mut u8,
-    size: usize,
-    align: usize,
+    free: unsafe fn(*mut u8),
 }
 
 /// A single, thread-local bag of garbage.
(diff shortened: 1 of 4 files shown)


src/mem/epoch/garbage.rs
```diff
@@ -23,10 +20,12 @@ impl Bag {
         if size > 0 {
             self.0.push(Item {
                 ptr: elem as *mut u8,
-                size: size,
-                align: mem::align_of::<T>(),
+                free: free::<T>,
             })
         }
+        unsafe fn free<T>(t: *mut u8) {
+            drop(Vec::from_raw_parts(t as *mut T, 0, 1));
```
cristicbz: This violates the `Vec` docs, doesn't it? (where `from_raw_parts` should only be passed the result of `into_raw_parts`) I think at least some kind of `// HACK` comment is in order, if you really want to get rid of `heap::deallocate`. Also the extra level of indirection introduced by calling `free` for each element may turn out to have perf impact, have you measured any decrease in the benchmarks? My 2c.

src/mem/epoch/mod.rs
```diff
@@ -406,7 +408,7 @@ thread_local!(static LOCAL_EPOCH: LocalEpoch = LocalEpoch::new() );
 
 #[must_use]
 pub struct Guard {
-    _dummy: ()
+    _marker: marker::PhantomData<Rc<()>>, // !Send and !Sync
```
cristicbz: Why `Rc` rather than, say, `*mut ()`? To avoid the extra `use`, if nothing else.

alexcrichton: Ah yeah that's work as well, forgot about that!

PR #5, opened 2015-09-14
Do not drop `Owned<T>` after store

whataloadofwhat:


aturon:
Gah, good catches, thanks!

PR #8, opened 2015-09-28
Docs

steveklabnik:
I want to start using scoped threads, but I also don't want to have to memorize the API :)

### DIFF
--- a/src/scoped.rs
+++ b/src/scoped.rs
@@ -40,6 +40,22 @@ pub struct ScopedJoinHandle<T> {
     thread: thread::Thread,
 }
 
+/// Create a new `scope`, for scoped concurrency operations.
+///
+/// Using crossbeam for concurrency involves creating a [`Scope`][scope]. That's the job of
+/// `scope`.
+///
+/// [scope]: struct.Scope.html
+///
+/// # Examples
+///
+/// Creating a scope:
+///
+/// ```
+/// crossbeam::scope(|scope| {
+///     // do stuff with scope here
+/// });
+/// ```
 pub fn scope<'a, F, R>(f: F) -> R where F: FnOnce(&Scope<'a>) -> R {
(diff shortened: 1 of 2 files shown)


aturon:
Thanks so much, @steveklabnik!

The only problem here is that the scoped spawning actually *does* still involve allocation under the hood: it dynamically tracks all of the cleanup it has to do when the scope is exited. But we can certainly make that more efficient than boxing.

steveklabnik:
Ah ha! Okay, let me remove that line then.

To be clear, it has to allocate _something_, but it doesn't need to box up the things it's closing over? It's the cleanup stuff that gets allocated?

aturon:
@steveklabnik That's the right idea, although the issue isn't so much the size as that the allocation can be amortized (e.g. the way we do with `Vec` where we alloc chunks in exponentially larger sizes as you run out of capacity).

Basically, the allocation is being managed by the library, which gives us leeway to be smarter about it.

steveklabnik:
@aturon removed the allocation comment

src/scoped.rs
```diff
@@ -78,6 +94,116 @@ impl<'a> Scope<'a> {
         });
     }
 
+    /// Create a scoped thread.
+    ///
+    /// `spawn` is similar to the [`spawn`][spawn] function in Rust's standard library. The
+    /// difference is that this thread is scoped, meaning that it's guaranteed to terminate
+    /// before the current stack frame goes away, allowing you to reference the parent stack frame
```
aturon: Can you mention that this guarantee is ensured by having the parent thread join on the child thread before exiting the scope?

steveklabnik: updated!

aturon:
Merged, thanks again @steveklabnik!

steveklabnik:
:confetti_ball:

PR #25, opened 2015-12-12
Add an implementation of the Chase-Lev deque

alexcrichton:
This adds an implementation of the Chase-Lev work-stealing deque which was found
in the standard library long ago. This implementation has been modified to use
crossbeam's memory reclamation scheme instead of a custom "buffer pool", so it
should actually free data now!

### DIFF
--- a/src/sync/mod.rs
+++ b/src/sync/mod.rs
@@ -9,3 +9,4 @@ mod atomic_option;
 mod ms_queue;
 mod treiber_stack;
 mod seg_queue;
+pub mod chase_lev;
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,3 +10,6 @@ license = "Apache-2.0/MIT"
 
 [features]
 nightly = []
+
+[dev-dependencies]
+rand = "0.3"
--- a/src/sync/chase_lev.rs
+++ b/src/sync/chase_lev.rs
@@ -1,0 +1,576 @@
+// Copyright 2013 The Rust Project Developers. See the COPYRIGHT
+// file at the top-level directory of this distribution and at
+// http://rust-lang.org/COPYRIGHT.
+//
+// Licensed under the Apache License, Version 2.0 <LICENSE-APACHE or
+// http://www.apache.org/licenses/LICENSE-2.0> or the MIT license
+// <LICENSE-MIT or http://opensource.org/licenses/MIT>, at your
+// option. This file may not be copied, modified, or distributed
+// except according to those terms.
+
+//! A lock-free concurrent work-stealing deque
+//!
+//! This module contains an implementation of the Chase-Lev work stealing deque
+//! described in "Dynamic Circular Work-Stealing Deque". The implementation is
+//! heavily based on the pseudocode found in the paper.
+//!
+//! # Example
+//!
+//! ```
+//! use crossbeam::sync::chase_lev;
+//! let (mut worker, stealer) = chase_lev::deque();
+//!
+//! // Only the worker may push/try_pop
+//! worker.push(1);
+//! worker.try_pop();
+//!
+//! // Stealers take data from the other end of the deque
+//! worker.push(1);
+//! stealer.steal();
+//!
+//! // Stealers can be cloned to have many stealers stealing in parallel
+//! worker.push(1);
+//! let stealer2 = stealer.clone();
+//! stealer2.steal();
+//! ```
+
+// FIXME: all atomic operations in this module use a SeqCst ordering. That is
+//      probably overkill
+
+use std::cell::UnsafeCell;
+use std::mem;
+use std::ptr;
+use std::sync::atomic::Ordering::SeqCst;
+use std::sync::atomic::AtomicIsize;
+use std::sync::Arc;
+
+use mem::epoch::{self, Atomic, Shared, Owned};
+
+// Once the queue is less than 1/K full, then it will be downsized. Note that
+// the deque requires that this number be less than 2.
+const K: isize = 4;
+
+// Minimum number of bits that a buffer size should be. No buffer will resize to
+// under this value, and all deques will initially contain a buffer of this
+// size.
+//
+// The size in question is 1 << MIN_BITS
+const MIN_BITS: u32 = 7;
+
+struct Deque<T> {
+    bottom: AtomicIsize,
+    top: AtomicIsize,
+    array: Atomic<Buffer<T>>,
+}
+
+// FIXME: can these constraints be relaxed?
+unsafe impl<T: Send + Sync> Send for Deque<T> {}
+unsafe impl<T: Send + Sync> Sync for Deque<T> {}
+
+/// Worker half of the work-stealing deque. This worker has exclusive access to
+/// one side of the deque, and uses `push` and `try_pop` method to manipulate it.
+///
+/// There may only be one worker per deque, and operations on the worker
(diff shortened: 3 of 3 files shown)


src/sync/deque.rs
```diff
@@ -0,0 +1,577 @@
...
+
+struct Deque<T> {
+    bottom: AtomicIsize,
+    top: AtomicIsize,
+    array: Atomic<Buffer<T>>,
+}
+
+// FIXME: can these constraints be relaxed?
+unsafe impl<T: Send + Sync> Send for Deque<T> {}
+unsafe impl<T: Send + Sync> Sync for Deque<T> {}
```
alexcrichton: I *think* that these are just the normal implications (e.g. Send => Send and Sync => Sync), but these concurrent/shared data structures often have some weird way over another so I ended up just making these overly restrictive and they can perhaps be relaxed in the future.

src/sync/deque.rs
```diff
@@ -0,0 +1,577 @@
...
+const K: isize = 4;
+
+// Minimum number of bits that a buffer size should be. No buffer will resize to
+// under this value, and all deques will initially contain a buffer of this
+// size.
+//
+// The size in question is 1 << MIN_BITS
+const MIN_BITS: u32 = 7;
+
+struct Deque<T> {
```
alexcrichton: Currently this isn't public, but in theory it can be so long as the `push` and `pop` methods are `unsafe` (as they only work if one thread is calling them).

For now I figured it could at least start out with the safe API that imposes an `Arc` allocation (the `deque` constructor) and those could perhaps all be deprecated in the future if necessary?

src/sync/mod.rs
```diff
@@ -9,3 +9,4 @@ mod atomic_option;
 mod ms_queue;
 mod treiber_stack;
 mod seg_queue;
+pub mod deque;
```
alexcrichton: If the only exported type from this module were `Deque` I'd want to just reexport it, but with a `deque` constructor, `Worker`, `Stealer`, and `Steal` types I figured it may be best for now to just expose the module.

Open to suggestions though!

alexcrichton:
Also in terms of where this implementation came from, I copied the deleted file from https://github.com/rust-lang/rust/commit/71d4e77db8ad4b6d821da7e5d5300134ac95974e (a real blast from the past) and then just brought it up to date and got it working with crossbeam instead of using a custom mutex inside.

src/sync/deque.rs
aturon: I agree with that call. In general, I expect to experiment with the question of safe APIs in this crate -- as we've talked about, providing things like iterators or `get` on a concurrent hashmap are somewhat problematic in safe APIs.

src/sync/deque.rs
aturon: I'm happy to revisit this later.

src/sync/deque.rs
aturon: Also note that the allocation here is unlikely to matter much; you shouldn't be spinning up/down work-stealing queues very often...

src/sync/deque.rs
```diff
@@ -0,0 +1,577 @@
...
+/// `steal` method.
+///
+/// Stealers can be cloned to have more than one handle active at a time.
+pub struct Stealer<T> {
+    deque: Arc<Deque<T>>,
+}
+
+/// When stealing some data, this is an enumeration of the possible outcomes.
+#[derive(PartialEq, Eq, Debug)]
+pub enum Steal<T> {
```
aturon: Hm, I might've gone for `Result<Option<T>, ()>`. But this is fine.

src/sync/deque.rs
```diff
@@ -0,0 +1,577 @@
...
+
+impl<T> Worker<T> {
+    /// Pushes data onto the front of this work queue.
+    pub fn push(&mut self, t: T) {
+        unsafe { self.deque.push(t) }
+    }
+
+    /// Pops data off the front of the work queue, returning `None` on an empty
+    /// queue.
+    pub fn pop(&mut self) -> Option<T> {
```
aturon: We may want to call this `try_pop` to match what we're doing with other queues. (Granted, this data structure will probably never have a blocking variant...)

src/sync/deque.rs
```diff
@@ -0,0 +1,577 @@
...
+    }
+
+    // Helper routine not mentioned in the paper which is used in growing and
+    // shrinking buffers to swap in a new buffer into place. As a bit of a
+    // recap, the whole point that we need a buffer pool rather than just
+    // calling malloc/free directly is that stealers can continue using buffers
+    // after this method has called 'free' on it. The continued usage is simply
+    // a read followed by a forget, but we must make sure that the memory can
+    // continue to be read after we flag this buffer for reclamation.
+    unsafe fn swap_buffer<'a>(&self,
```
aturon: This is kinda amazing...

src/sync/deque.rs
```diff
@@ -0,0 +1,577 @@
...
+
+
+impl<T> Drop for Deque<T> {
+    fn drop(&mut self) {
+        let guard = epoch::pin();
+
+        let t = self.top.load(SeqCst);
+        let b = self.bottom.load(SeqCst);
+        let a = self.array.swap(None, SeqCst, &guard).unwrap();
+        // Free whatever is leftover in the dequeue, and then move the buffer
```
aturon: This comment is out of date.

src/sync/deque.rs
```diff
@@ -0,0 +1,577 @@
...
+    unsafe fn maybe_shrink(&self, b: isize, t: isize, guard: &epoch::Guard) {
+        let a = self.array.load(SeqCst, guard).unwrap();
+        if b - t < (a.size() as isize) / K && b - t > (1 << MIN_BITS) {
+            self.swap_buffer(b, a, a.resize(b, t, -1), guard);
+        }
+    }
+
+    // Helper routine not mentioned in the paper which is used in growing and
+    // shrinking buffers to swap in a new buffer into place. As a bit of a
+    // recap, the whole point that we need a buffer pool rather than just
```
aturon: There's no longer a pool, huzzah! Comment out of date ;-)

src/sync/deque.rs
```diff
@@ -0,0 +1,577 @@
...
+                        remaining.fetch_sub(1, SeqCst);
+                    } else {
+                        panic!()
+                    }
+                }
+                None => {}
+            }
+        }
+
+        for thread in threads.into_iter() {
```
aturon: Could probably rewrite via scoped threads, but whatever.

src/sync/mod.rs
aturon: If only we had inherent associated types!

Exporting a module is fine. However, please rename to something like `cl_deque` or `chase_lev_deque`. For the moment, I'm trying to keep fairly specific about the algorithm being implemented. Later on, we may have aliases for our "preferred" queues, deques, etc.

aturon:
OK, this looks quite good! I've left a number of minor notes, most of which can be ignored for now. I didn't review the core algorithm very carefully, as I'd need to re-read the paper and this is code we already know works. But I did look at the bits related to memory management: :+1:

src/sync/deque.rs
alexcrichton: *swoons*

alexcrichton:
Updated, yay!

aturon:
Merged, thanks!

PR #27, opened 2015-12-16
Add blocking support to `MsQueue`.

### LINKED ISSUE #10: Blocking channels
It can be useful to have channels that can block when they're empty (or full, if they're size limited).

aturon:
This involves renaming the existing `pop` method in all queues to
`try_pop`, which is useable for non-blocking `pop` attempts.

The new `pop` method in `MsQueue` will block until data is available.

There's relatively little overhead on purely non-blocking use-cases, and
most uses of queues want blocking functionality anyway.

Closes #10.

r? @alexcrichton @huonw

### DIFF
--- a/src/sync/ms_queue.rs
+++ b/src/sync/ms_queue.rs
@@ -1,21 +1,55 @@
 use std::sync::atomic::Ordering::{Acquire, Release, Relaxed};
+use std::sync::atomic::AtomicBool;
 use std::{ptr, mem};
+use std::thread::{self, Thread};
 
-use mem::epoch::{self, Atomic, Owned};
+use mem::epoch::{self, Atomic, Owned, Shared};
 use mem::CachePadded;
 
-/// A Michael-Scott lock-free queue.
+/// A Michael-Scott lock-free queue, with support for blocking `pop`s.
 ///
 /// Usable with any number of producers and consumers.
+// The representation here is a singly-linked list, with a sentinel
+// node at the front. In general the `tail` pointer may lag behind the
+// actual tail. Non-sentinal nodes are either all `Data` or all
+// `Blocked` (requests for data from blocked threads).
 pub struct MsQueue<T> {
     head: CachePadded<Atomic<Node<T>>>,
     tail: CachePadded<Atomic<Node<T>>>,
 }
 
 struct Node<T> {
-    data: T,
+    payload: Payload<T>,
     next: Atomic<Node<T>>,
 }
+
+enum Payload<T> {
+    /// A node with actual data that can be popped.
+    Data(T),
+    /// A node representing a blocked request for data.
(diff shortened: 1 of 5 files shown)


src/sync/ms_queue.rs
```diff
@@ -34,49 +68,209 @@ impl<T> MsQueue<T> {
         q
     }
 
-    /// Add `t` to the back of the queue.
+    #[inline(always)]
```
alexcrichton: :imp:

src/sync/ms_queue.rs
```diff
@@ -34,49 +68,209 @@ impl<T> MsQueue<T> {
...
+                        Payload::Blocked(signal) => Some((next, signal)),
+                        Payload::Data(_) => None,
+                    }
+                });
+                if let Some((blocked_node, signal)) = request {
+                    // race to dequeue the node
+                    if self.head.cas_shared(Some(head), Some(blocked_node), Release) {
+                        unsafe {
+                            // signal the thread
+                            ptr::write(&mut (*signal).data, Some(cache.into_data()));
```
alexcrichton: With this using `Option<T>` now, this `ptr::write` could just be a normal assignment?

src/sync/ms_queue.rs
```diff
@@ -34,49 +68,209 @@ impl<T> MsQueue<T> {
...
+        });
+
+        loop {
+            // try a normal pop
+            if let Ok(Some(r)) = self.pop_internal(&guard) {
+                return r;
+            }
+
+            let head = self.head.load(Relaxed, &guard).unwrap();
+            let tail = self.tail.load(Relaxed, &guard).unwrap();
```
alexcrichton: I remember when we were talking about this we concluded that it was important the load of tail happened before the load of the head. I can't quite recall why, but was the swap here intentional?

alexcrichton:
lgtm, I like the cleanup!

alexcrichton:
Oh wow, yeah with the #10 reference I just realized that this is an implementation of a more feature-ful channel than the standard library (mpmc, right?).

I'm curious what a benchmark would look like between the two? If it's fast enough we could just move it into the standard library wholesale!

src/sync/ms_queue.rs
aturon: The logic here works differently than when we were pairing, but I'll rewrite to be a bit more clear. In any case, the order of snapshots here *does not matter*.

aturon:
Pushed up a fix getting rid of `write` and clarifying the logic around head/tail snapshotting.

re: channels, the built-in benchmark compares them head-to-head already for an mpsc scenario. On my machine, the MSQ is very slightly faster.

But we'd have to think about the interaction with select, right? I thought that was a primary blocker for going to mpmc channels in Rust.

PR #28, opened 2016-01-18
Relax T: Sync for MsQueue<T>: Send/Sync, and add T: Send for SegQueue.

huonw:
The latter is particularly important, as previously SegQueue would allow
sending, e.g., Rc<_> across threads.


aturon:
Thanks!

PR #29, opened 2016-01-18
Fixing race condition in garbage collection

schets:
The collect function in ConcBag is unsound - it's possible for a thread inserting into the linked list at the same time a thread enters collect to make an insertion that isn't visible. This happens when the collecting thread loads head (previously line 127), then thread b successfully completes the CAS to update head (line 122), then the collecting thread stores nullptr to head (previously line 128) and continues using a stale value for head. Moving to an atomic swap means that the CAS will not race with the store - either the CAS will complete first and that value will be visible in the swap, or the swap will complete first and the CAS will fail.

A similar issue happens when two threads attempt to collect at the same time, and is fixed by this as well.

### DIFF
--- a/src/mem/epoch/garbage.rs
+++ b/src/mem/epoch/garbage.rs
@@ -9,7 +9,7 @@
 use std::ptr;
 use std::mem;
 use std::sync::atomic::AtomicPtr;
-use std::sync::atomic::Ordering::{Relaxed, Release};
+use std::sync::atomic::Ordering::{Relaxed, Release, Acquire};
 
 use mem::ZerosValid;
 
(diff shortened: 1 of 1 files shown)


aturon:
Thanks for the PR!

So, I think the real bug here is lack of documentation: there is an external invariant that the `collect` function is only called when the thread has *exclusive* access to the bag. This happens because of the epoch management scheme. There are no threads racing to insert or collect at the same time.

schets:
I may be misreading the global collection and garbage migration conditions - but if the global epoch is incremented during a call to migrate_garbage before the global epoch is loaded (which can happen if the migrating thread's epoch is up to date), then won't the calls to insert in migrate_garbage insert into each of the three global garbage bags - cur_epoc - 1, cur_epoch, and global::epoch (== cur_epoch + 1)? This would mean that one of the bags is equivalent to the bag being collected in try collect.

aturon:
@schets Ahha, you appear to be right. That garbage migration scheme was added later and I wasn't taking it into account with the concurrent bag. I wonder whether the migration scheme is worth it, vs just dumping the garbage into the current global epoch.

In any case, I'm going to take this PR for the time being. Thanks again!

schets:
> I wonder whether the migration scheme is worth it, vs just dumping the garbage into the current global epoch.

Maybe it's better to continue this discussion in an issue or another PR, but I think so - dumping everything into a big global epoch would make global collection more prone to big latency spikes. One could add support for performing partial global GC, but the current scheme naturally results in lower latency without making users worry about GC limits.

This shouldn't really affect performance anyways - the cost of the xchg will be dwarfed by the deallocations.

aturon:
@schets Yeah, that was definitely my original motivation. I just mean: I haven't rigorously measured any of this :-)

And yes, I agree about the cost of the xchg as well.

PR #31, opened 2016-01-25
Add Rust 1.2.0 compatiblity

tbu-:
This would allow Cargo to build on 1.2.0 again: https://github.com/rust-lang/cargo/pull/2297#issuecomment-174617125.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -44,6 +44,25 @@ impl<F: FnOnce()> FnBox for F {
     fn call_box(self: Box<Self>) { (*self)() }
 }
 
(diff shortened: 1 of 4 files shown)


alexcrichton:
On second though, @aturon and I briefly chatted and the conclusion was that [Cargo probably doesn't have a reason to support 1.2](https://github.com/rust-lang/cargo/pull/2319)

aturon:
Indeed. Thanks for the PR, @tbu-, but for the time being I'm going to let crossbeam march forward with new stable Rust releases.

PR #32, opened 2016-01-26
Relax the requirements for `Deque<T>` to be `Send` and `Sync`.

kinghajj:
The old `std::sync::deque` module, and the `deque` crate descended from it, require only that `T` implement `Send` in order for `Deque<T>` to implement `Send` and `Sync`. This crate's `chase_lev` module, however, currently requires `T` also to be `Sync`, making this example fail to compile.

    #![feature(optin_builtin_traits)]

    extern crate crossbeam;
    extern crate deque;

    use std::thread;

    struct Unsynced;

    impl !Sync for Unsynced { }

    #[test]
    fn deque_crate_unsynced() {
        use deque;

        let (worker, stealer) = deque::new();
        let handle = thread::spawn(move || {
            loop {
                match stealer.steal() {
                    deque::Stolen::Data(_) => {
                        break;
                    }
                    _ => {}
                }
            }
        });
        worker.push(Unsynced);
        handle.join().unwrap();
    }

    #[test]
    fn chase_lev_mod_unsynced() {
        use crossbeam::sync::chase_lev;

        let (mut worker, stealer) = chase_lev::deque();
        let handle = thread::spawn(move || {
            loop {
                match stealer.steal() {
                    chase_lev::Steal::Data(_) => {
                        break;
                    }
                    _ => {}
                }
            }
        });
        worker.push(Unsynced);
        handle.join().unwrap();
    }

Intuitively, this makes sense to me: each element has a clear ownership path from the originating thread, to the deque, and finally to the stealing thread.

### DIFF
--- a/src/sync/chase_lev.rs
+++ b/src/sync/chase_lev.rs
@@ -64,8 +64,8 @@ struct Deque<T> {
 }
 
 // FIXME: can these constraints be relaxed?
-unsafe impl<T: Send + Sync> Send for Deque<T> {}
-unsafe impl<T: Send + Sync> Sync for Deque<T> {}
+unsafe impl<T: Send> Send for Deque<T> {}
+unsafe impl<T: Send> Sync for Deque<T> {}
 
 /// Worker half of the work-stealing deque. This worker has exclusive access to
 /// one side of the deque, and uses `push` and `try_pop` method to manipulate it.


rphmeier:
This makes sense. Chase-Lev guarantees that every element pushed will only be either popped or stolen once total, so there is no aliasing of the value between threads. The `Sync` bound is unnecessary if you view it through the lens that the queue is a medium for performing a deferred send of the value from one thread to another.

aturon:
@alexcrichton Do you have thoughts on this? Was there a particular reason you ended up using the more conservative bounds when you landed this in crossbeam?

kinghajj:
Huh, didn't realize he wrote this module, too. (That explains why they're so similar!) This isn't a rigorous defense of the safety of the change, but I've been using crossbeam with this patch while working on a thread/work pool crate, and the tests for it all pass. I needed this change because the requirement that jobs be Sync is too onerous, and prevents moving channel Senders into worker threads so that they may send results back.

alexcrichton:
I mostly just did this out of fear of getting it wrong, I didn't necessarily have any particular error case in mind.

These types to seem very similar to `std::sync::mpsc`, however, so here's my thinking (framed in terms of the exported types):

* `Worker<T>` and `Stealer<T>` seem fine to both be `Send` if `T` is `Send`. That mirrors libstd's channels and "feels correct" to me as well.
* I believe there's no need for `Worker<T>` to be `Sync`, but even if it is you can't do anything with it unless you have exclusive access. I'm less sure what implies `Sync` for the bounds here, but I *think* that it's `T: Send` as well because the inner type itself is never being shared (ownership is always being transferred)
* I believe that `Stealer<T>` can be `Sync` so long as `T` is `Send`. Which is to say that if a `Stealer` is shared among a bunch of threads (which is basically intended) then it'll only ever transfer ownership of `T` out rather than attempt to share it among threads.

Basically, **tl;dr;** I believe removing `Sync` is fine.

alexcrichton:
Oh and although I was thinking about the exported Worker/Stealer types, I *think* that given the conclusion I reached it's fine to place the impls on the Deque type instead

aturon:
Great, thanks @kinghajj and @alexcrichton!

PR #33, opened 2016-01-26
Use more relaxed orderings in chase_lev

rphmeier:
I wasn't sure what to do with (most of) the orderings in resize-related methods, so I just left them as SeqCst.

It would be nice to have benchmarks here. I am also not sure that my merging of techniques from the two papers is completely accurate, although all tests pass on my machine.

I additionally changed the operations on top and bottom to be wrapping, but this shouldn't cause any issues.

### DIFF
--- a/src/sync/chase_lev.rs
+++ b/src/sync/chase_lev.rs
@@ -10,9 +10,10 @@
 
 //! A lock-free concurrent work-stealing deque
 //!
-//! This module contains an implementation of the Chase-Lev work stealing deque
-//! described in "Dynamic Circular Work-Stealing Deque". The implementation is
-//! heavily based on the pseudocode found in the paper.
+//! This module contains a hybrid implementation of the Chase-Lev work stealing deque
+//! described in ["Dynamic Circular Work-Stealing Deque"][chase_lev] and the improved version
+//! described in ["Correct and Efficient Work-Stealing for Weak Memory Models"][weak_chase_lev].
+//! The implementation is heavily based on the pseudocode found in the papers.
 //!
 //! # Example
 //!
@@ -33,15 +34,15 @@
 //! let stealer2 = stealer.clone();
 //! stealer2.steal();
 //! ```
-
-// FIXME: all atomic operations in this module use a SeqCst ordering. That is
-//      probably overkill
+//!
+//! [chase_lev]: http://neteril.org/~jeremie/Dynamic_Circular_Work_Queue.pdf
+//! [weak_chase_lev]: http://www.di.ens.fr/~zappa/readings/ppopp13.pdf
 
 use std::cell::UnsafeCell;
 use std::mem;
 use std::ptr;
-use std::sync::atomic::Ordering::SeqCst;
-use std::sync::atomic::AtomicIsize;
+use std::sync::atomic::Ordering::{Acquire, Relaxed, Release, SeqCst};
+use std::sync::atomic::{AtomicIsize, fence};
 use std::sync::Arc;
 
 use mem::epoch::{self, Atomic, Shared, Owned};
@@ -158,68 +159,82 @@ impl<T> Deque<T> {
     unsafe fn push(&self, data: T) {
         let guard = epoch::pin();
 
-        let mut b = self.bottom.load(SeqCst);
-        let t = self.top.load(SeqCst);
-        let mut a = self.array.load(SeqCst, &guard).unwrap();
+        let mut b = self.bottom.load(Relaxed);
+        let t = self.top.load(Acquire);
+        let mut a = self.array.load(Relaxed, &guard).unwrap();
+
         let size = b - t;
         if size >= (a.size() as isize) - 1 {
             // You won't find this code in the chase-lev deque paper. This is
             // alluded to in a small footnote, however. We always free a buffer
             // when growing in order to prevent leaks.
-            a = self.swap_buffer(b, a, a.resize(b, t, 1), &guard);
-            b = self.bottom.load(SeqCst);
+            a = self.swap_buffer(a, a.resize(b, t, 1), &guard);
+
+            // reload the bottom counter, since swap_buffer modifies it.
+            b = self.bottom.load(Relaxed);
         }
         a.put(b, data);
-        self.bottom.store(b + 1, SeqCst);
+        fence(Release);
+        self.bottom.store(b + 1, Relaxed);
     }
 
     unsafe fn try_pop(&self) -> Option<T> {
         let guard = epoch::pin();
 
-        let b = self.bottom.load(SeqCst);
-        let a = self.array.load(SeqCst, &guard).unwrap();
-        let b = b - 1;
-        self.bottom.store(b, SeqCst);
-        let t = self.top.load(SeqCst);
+        let b = self.bottom.load(Relaxed) - 1;
+        let a = self.array.load(Relaxed, &guard).unwrap();
+        self.bottom.store(b, Relaxed);
+        fence(SeqCst); // the store to bottom must occur before loading top.
+        let t = self.top.load(Relaxed);
+
         let size = b - t;
-        if size < 0 {
-            self.bottom.store(t, SeqCst);
-            return None
-        }
-        let data = a.get(b);
-        if size > 0 {
-            self.maybe_shrink(b, t, &guard);
-            return Some(data)
-        }
-        if self.top.compare_and_swap(t, t + 1, SeqCst) == t {
-            self.bottom.store(t + 1, SeqCst);
-            return Some(data)
+        if size >= 0 {
+            // non-empty case
+            let mut data = Some(a.get(b));
+            if size == 0 {
+                // last element in queue, check for races.
+                if self.top.compare_and_swap(t, t + 1, SeqCst) != t {
+                    // lost the race.
+                    mem::forget(data.take());
+                }
+
+                // set the queue to a canonically empty state.
+                self.bottom.store(b + 1, Relaxed);
+            } else {
+                self.maybe_shrink(b, t, &guard);
+            }
+            data
         } else {
-            self.bottom.store(t + 1, SeqCst);
-            mem::forget(data); // someone else stole this value
-            return None
+            // empty queue. revert the decrement of "b" and try to shrink.
+            //
+            // the original chase_lev paper uses "t" here, but the new one uses "b + 1".
+            // don't worry, they're the same thing: pop and steal operations will never leave
+            // the top counter greater than the bottom counter. After we decrement "b" at
+            // the beginning of this function, the lowest possible value it could hold here is "t - 1".
+            // That's also the only value that could cause this branch to be taken.
+            self.bottom.store(b + 1, Relaxed);
+            None
         }
     }
 
     fn steal(&self) -> Steal<T> {
         let guard = epoch::pin();
 
-        let t = self.top.load(SeqCst);
-        let old = self.array.load(SeqCst, &guard).unwrap();
-        let b = self.bottom.load(SeqCst);
-        let a = self.array.load(SeqCst, &guard).unwrap();
+        let t = self.top.load(Acquire);
+        fence(SeqCst); // top must be loaded before bottom.
+        let b = self.bottom.load(Acquire);
+
         let size = b - t;
         if size <= 0 {
             return Steal::Empty
         }
-        if (size as usize) % a.size() == 0 {
-            if *a as *const _ == *old as *const _ && t == self.top.load(SeqCst) {
-                return Steal::Empty
-            }
-            return Steal::Abort
-        }
+
         unsafe {
+            // while the paper uses a "consume" ordering here, the closest thing we have
+            // available is Acquire, which is strictly stronger.
+            let a = self.array.load(Acquire, &guard).unwrap();
             let data = a.get(t);
+            // we may be racing against other steals and a pop.
             if self.top.compare_and_swap(t, t + 1, SeqCst) == t {
                 Steal::Data(data)
             } else {
@@ -229,10 +244,12 @@ impl<T> Deque<T> {
         }
     }
 
+    // potentially shrink the array. This can be called only from the worker.
     unsafe fn maybe_shrink(&self, b: isize, t: isize, guard: &epoch::Guard) {
         let a = self.array.load(SeqCst, guard).unwrap();
-        if b - t < (a.size() as isize) / K && b - t > (1 << MIN_BITS) {
-            self.swap_buffer(b, a, a.resize(b, t, -1), guard);
+        let size = b - t;
+        if size < (a.size() as isize) / K && size > (1 << MIN_BITS) {
+            self.swap_buffer(a, a.resize(b, t, -1), guard);
         }
     }
 
@@ -245,22 +262,19 @@ impl<T> Deque<T> {
     // to be read after we flag this buffer for reclamation. All stealers,
     // however, have their own epoch pinned during this time so the buffer will
     // just naturally be free'd once all concurrent stealers have exited.
+    //
+    // This method may only be called safely from the workers due to the way it modifies
+    // the array pointer.
     unsafe fn swap_buffer<'a>(&self,
-                              b: isize,
                               old: Shared<'a, Buffer<T>>,
                               buf: Buffer<T>,
                               guard: &'a epoch::Guard)
                               -> Shared<'a, Buffer<T>> {
         let newbuf = Owned::new(buf);
-        let newbuf = self.array.store_and_ref(newbuf, SeqCst, &guard);
-        let ss = newbuf.size() as isize;
-        self.bottom.store(b + ss, SeqCst);
-        let t = self.top.load(SeqCst);
-        if self.top.compare_and_swap(t, t + ss, SeqCst) != t {
-            self.bottom.store(b, SeqCst);
-        }
+        let newbuf = self.array.store_and_ref(newbuf, Release, &guard);
         guard.unlinked(old);
-        return newbuf
+
+        newbuf
     }
 }
 
@@ -269,9 +283,11 @@ impl<T> Drop for Deque<T> {
     fn drop(&mut self) {
         let guard = epoch::pin();
 
-        let t = self.top.load(SeqCst);
-        let b = self.bottom.load(SeqCst);
-        let a = self.array.swap(None, SeqCst, &guard).unwrap();
+        // Arc enforces that we have truly exclusive access here.
+
+        let t = self.top.load(Relaxed);
+        let b = self.bottom.load(Relaxed);
+        let a = self.array.swap(None, Relaxed, &guard).unwrap();
         // Free whatever is leftover in the dequeue, then free the backing
         // memory itself
         unsafe {


rphmeier:
cc @alexcrichton

rphmeier:
you may want to hold off on merging this just yet if you have ARM targets in mind. I think that the implementation holds for a strong memory model but may not for weaker ones.

src/sync/chase_lev.rs
```diff
@@ -34,14 +34,11 @@
 //! stealer2.steal();
 //! ```
 
-// FIXME: all atomic operations in this module use a SeqCst ordering. That is
-//      probably overkill
```
alexcrichton: Could you replace this comment with a reference to the paper from which all atomic orderings are referencing? It's good to have something to double check against!

src/sync/chase_lev.rs
```diff
@@ -158,69 +155,93 @@ impl<T> Deque<T> {
...
 
-        let mut b = self.bottom.load(SeqCst);
-        let t = self.top.load(SeqCst);
-        let mut a = self.array.load(SeqCst, &guard).unwrap();
-        let size = b - t;
+        let mut b = self.bottom.load(Relaxed);
+        let t = self.top.load(Acquire);
+        let mut a = self.array.load(Relaxed, &guard).unwrap();
+        
+        let size = b.wrapping_sub(t);
```
alexcrichton: How come all the arithmetic operations were changed to wrapping? In general I would expect overflows here to indeed be fatal flaws that should be accounted for via different means elsewhere.

src/sync/chase_lev.rs
```diff
@@ -158,69 +155,93 @@ impl<T> Deque<T> {
...
     unsafe fn try_pop(&self) -> Option<T> {
+        // try to perform an early return.
+        {
+            let t = self.top.load(Relaxed);
+            let b = self.bottom.load(Relaxed);
+            let len = b.wrapping_sub(t);
+            if len <= 0 {
+                return None;
+            }
+        }
```
alexcrichton: Without performance benchmarks, I'd personally prefer to stick to keeping the code here as close to the papers as much as possible, so could this be elided?

src/sync/chase_lev.rs
```diff
@@ -158,69 +155,93 @@ impl<T> Deque<T> {
...
-            self.bottom.store(t + 1, SeqCst);
-            return Some(data)
+        let b = self.bottom.load(Relaxed).wrapping_sub(1);
+        let a = self.array.load(Relaxed, &guard).unwrap();
+        self.bottom.store(b, Relaxed);
+        fence(SeqCst); // the store to bottom must occur before loading top.
+        let t = self.top.load(Relaxed);
+        
+        let size = b.wrapping_sub(t);
+        if size >= 0 {
```
alexcrichton: This branch and some following logic seems to have changed from [the original paper](http://neteril.org/~jeremie/Dynamic_Circular_Work_Queue.pdf), which worries me a bit...

For example in the false branch here `b` is updated to `b + 1`, but it was originally updated to `t`

src/sync/chase_lev.rs
rphmeier: Generally speaking, I wanted to follow more closely to the papers which may have accounted for some integer overflow. The standard Chase-Lev paper has a footnote (4) specifying that they assume top will never overflow, implying that bottom might. In practice, this is probably not a big deal, but we should probably use i64 rather than isize. I don't mind reverting this change if you have a strong opinion.

src/sync/chase_lev.rs
rphmeier: Yeah, that's no issue.

src/sync/chase_lev.rs
rphmeier: I have copied that portion almost verbatim from http://www.di.ens.fr/~zappa/readings/ppopp13.pdf
(Link will be included in my next commit)

src/sync/chase_lev.rs
rphmeier: Also: the two are equivalent in this case, since we first decrement b and then compare it to t. If they are equal we can store b+1 or t+1 at the end, it makes no difference.

src/sync/chase_lev.rs
rphmeier: Also: the two are equivalent in this case, since we first decrement b and then compare it to t. If they are equal we can store b+1 or t+1 at the end, it makes no difference.

src/sync/chase_lev.rs
```diff
@@ -245,19 +262,25 @@ impl<T> Deque<T> {
...
         let newbuf = self.array.store_and_ref(newbuf, SeqCst, &guard);
+        // add the new size to the bottom and top counters as a way of signalling
+        // to current stealers that the buffer has been resized.
         let ss = newbuf.size() as isize;
-        self.bottom.store(b + ss, SeqCst);
+        self.bottom.store(b + ss, Relaxed);
         let t = self.top.load(SeqCst);
         if self.top.compare_and_swap(t, t + ss, SeqCst) != t {
-            self.bottom.store(b, SeqCst);
+            self.bottom.store(b, Relaxed);
```
rphmeier: Specifically, this is the kind of thing that might not fly on ARM. See also at line 174 where we reload the bottom counter after swapping. Does ARM guarantee that (on the same thread, with relaxed atomics) we will read back the value we just wrote? It seems to me that it's a safe guarantee using non-atomic instructions, but maybe the relaxed semantics bypass the memory model's usual guarantees for sequential code?

rphmeier: This should be fine, since it's one CPU.

aturon:
(Just to be clear, I'm going to treat Alex as the reviewer here, since he's got a deeper understanding of this particular algorithm.)

src/sync/chase_lev.rs
alexcrichton: Hm ok, let's stick with panicking operations for now and defer to a time where we can analyze a little more in detail about whether a wrapping operation is needed or not. I'm fine doing this in a follow-up PR though!

src/sync/chase_lev.rs
alexcrichton: Hm ok, I'll try to work through to make sure they're equivalent.

src/sync/chase_lev.rs
```diff
@@ -158,68 +159,82 @@ impl<T> Deque<T> {
...
 
-        let t = self.top.load(SeqCst);
-        let old = self.array.load(SeqCst, &guard).unwrap();
-        let b = self.bottom.load(SeqCst);
-        let a = self.array.load(SeqCst, &guard).unwrap();
+        let t = self.top.load(Acquire);
+        let old = self.array.load(Acquire, &guard).unwrap();
+        fence(SeqCst); // top must be loaded before bottom.
+        let b = self.bottom.load(Acquire);
+        let a = self.array.load(Acquire, &guard).unwrap();
```
alexcrichton: In the ordering-paper this has a consume ordering and happens after the check for `size <= 0`. I'm not entirely sure where consume ordering falls on the spectrum of orderings, but I wanted to confirm here. 

One plausible course of action could possibly be to leave this as `SeqCst` but move it after the `size <= 0` check?

src/sync/chase_lev.rs
```diff
@@ -158,68 +159,82 @@ impl<T> Deque<T> {
...
         if size <= 0 {
             return Steal::Empty
         }
+        
+        // check for a specific edge case where the buffer has been shrunk while
+        // this steal operation was in progress.
         if (size as usize) % a.size() == 0 {
-            if *a as *const _ == *old as *const _ && t == self.top.load(SeqCst) {
-                return Steal::Empty
+            if *a as *const _ == *old as *const _  && t == self.top.load(Acquire) {
```
alexcrichton: I don't think that this ordering was mentioned in the ordering paper (I don't think it dived much into freeing buffers), but that makes me a little hesitant to relax this a bit without a deeper understanding

src/sync/chase_lev.rs
alexcrichton: I'd be relatively confident that `SeqCst` is at least *a* correct ordering here, but I'd be less confident in relaxing them because these orderings aren't mentioned in the paper about orderings.

That being said, I believe this is a pretty cold path as resizing the internal buffer in theory shouldn't happen that often, so perhaps it wouldn't be too pad to leave these as `SeqCst` for now?

src/sync/chase_lev.rs
```diff
@@ -158,68 +159,82 @@ impl<T> Deque<T> {
...
+            } else {
+                self.maybe_shrink(b, t, &guard);
+            }
+            data
         } else {
-            self.bottom.store(t + 1, SeqCst);
-            mem::forget(data); // someone else stole this value
-            return None
+            // empty queue. revert the decrement of "b" and try to shrink.
+            self.bottom.store(b + 1, Relaxed);
```
alexcrichton: This seems to be an interesting difference between the two papers. The initial paper says `self.bottom = t`  and the ordering paper says `self.bottom = b + 1`. At this point We know that `b < t` so `b+1 <= t`, so I *think* that erring on the side of the initial paper to store `t` instead of `b+1` here should be fine. Does that make sense?

src/sync/chase_lev.rs
sorear: consume is not supported by LLVM but is strictly weaker than acquire.

alexcrichton:
Ok, with the exception of one store in `try_pop` it looks like the two papers are indeed the same. So I guess the summary of my comments are:

1. We may want to stick to `SeqCst` whenever we're managing the buffer. These orderings aren't mentioned in the orderings paper (as this was an addition to the original paper), and the resizing operations in theory happen rarely (and are accompanied with a memory allocation), so it may not be too bad to leave them all as such.
2. Can you add a comment to `try_pop` indicating that its structure matches the ordering-based paper, not the original one?
3. A few minor shufflings to resolve discrepancies between the papers, but otherwise looks good to me!

src/sync/chase_lev.rs
alexcrichton: Aha excellent! Can a comment be added to the effect to indicate this discrepancy?

src/sync/chase_lev.rs
rphmeier: The key change that we are checking for here occurs when we try to set "top" in `swap_buffer. For the `compare_and_swap`, we use a SeqCst ordering, so the Acquire load here should provide the same guarantees w.r.t sequential consistency as if we did a SeqCst load. In practice, this won't make any difference on x84, but it avoids a full memory barrier on ARM.

https://www.cl.cam.ac.uk/~pes20/cpp/cpp0xmappings.html

src/sync/chase_lev.rs
rphmeier: The second paper follows the pattern that "bottom" is always modified with "Relaxed" orderings, which is acceptable since those modifications only ever occur on a single thread. Bottom is only read in "steal" other threads after a full memory barrier, so Relaxed should be sufficient here.

It is a cold path, though, so maybe you are right that it's best not to tempt fate by relaxing the orderings.

rphmeier:
I don't know what the procedure here is with respect to the commit history, but once we've settled on the details let me know if you want me to rebase/squash.

src/sync/chase_lev.rs
rphmeier: My understanding is that in this case `b+1` == `t`. The paper doesn't allow for "bottom" to be less than "top" except during the execution of "try_pop" after we decrement "bottom" but before we re-increment it if necessary. So the only possible value of `b` in the else branch is `t-1`. Let me know if you see any flaws in my reasoning there.

rphmeier: (this can lead to an interesting bug when you use unsigned counters for top and bottom: if you try to pop from a queue with top = bottom = 0, decrementing bottom will cause it to wrap around and appear to be maximally full)

src/sync/chase_lev.rs
rphmeier: To clarify: I meant ARMv7. On ARMv8 the instruction seems to be the same.

src/sync/chase_lev.rs
alexcrichton: Ah ok, I think that makes sense to me yeah, thanks for explaining!

src/sync/chase_lev.rs
alexcrichton: Eh that seems like reasonable logic to me, and I suspect that using `Relaxed` here probably won't actually cause any problems. Can you add this in an explicit comment though?

src/sync/chase_lev.rs
alexcrichton: In terms of literal hardware instructions I think that's what we want yeah, but in terms of the abstract model I thought that `Acquire` can only pair with a `Release` somewhere? I didn't think that `Acquire` paired with `SeqCst`.

This part should be a pretty cold path as well, and I'd personally be more comfortable with a `SeqCst` to stick around here as it looks like it's not as straightforward as the `Relaxed` loads of `bottom` in `swap_buffer`. 

Does that sound ok?

sorear: SeqCst counts as both an Acquire and a Release, and is stronger than the combination; see http://llvm.org/docs/Atomics.html#sequentiallyconsistent .

alexcrichton: Ah right! One of these days I may understand this memory model... In that case this is probably fine

alexcrichton:
Ok, I think I'm happy with all the orderings here, I'd prefer a few more comments and @aturon would probably favor a squash as well, but otherwise seems good to go.

Thanks again @rphmeier!

rphmeier:
Okay, I think I've added the comments you wanted and squashed the commits. Glad to help when I can, @alexcrichton!

aturon:
Thanks @rphmeier and @alexcrichton!

I did have a question about this:

> you may want to hold off on merging this just yet if you have ARM targets in mind. I think that the implementation holds for a strong memory model but may not for weaker ones.

The C11 memory model (which Rust inherits) is intended to be hardware-independent, so this seems a bit surprising. Can you elaborate? ARM targets are definitely something I'd like to support, especially in this case, since Servo could eventually move over to using this deque.

rphmeier:
@aturon I'm not sure if that comment still applies, since that was when we were still going over orderings etc. I wasn't really sure what the semantics of a weak memory model were, and I'd heard some horror stories in the past about ARM's reorderings. As far as I can tell here, this should work on ARM.

rphmeier:
The travis failure here seems unrelated? All 7 chase_lev tests passed.

alexcrichton:
r+ from me for this, but those test failures definitely seem worrisome.

Basically looks like a segfault due to concurrency one way or another :(

src/sync/chase_lev.rs
```diff
@@ -158,68 +159,91 @@ impl<T> Deque<T> {
...
 
-        let mut b = self.bottom.load(SeqCst);
-        let t = self.top.load(SeqCst);
-        let mut a = self.array.load(SeqCst, &guard).unwrap();
+        let mut b = self.bottom.load(Relaxed);
+        let t = self.top.load(Acquire);
+        let mut a = self.array.load(Relaxed, &guard).unwrap();
+        
         let size = b - t;
         if size >= (a.size() as isize) - 1 {
```
rphmeier: Not sure what to make of this, but the original paper uses
`size >= a.size() - 1`
while the new paper uses
`size > a.size() - 1`

aturon:
Yep, something's busted, independently of this PR.

@rphmeier I'm going to hold off merging until I figure out what's going on with the library as a whole (there were some other recent changes that may have exposed this segfault).

rphmeier:
@aturon, that's fine, I just noticed a strange discrepancy between the papers.

aturon:
cc @msullivan -- I have not been able to reproduce the problem locally.

Amanieu:
The [deque](https://github.com/kinghajj/deque) crate already has an implementation of a Chase-Lev queue with relaxed atomic orderings. The API is the same since it is also based on the original deque implementation from the rust source tree.

aturon:
@Amanieu Oh cool, I didn't realize the orderings had been changed there.

aturon:
cc https://github.com/aturon/crossbeam/pull/42 regarding the segfaults. I think I've gotten to the bottom of it.

aturon:
@rphmeier This seems to rebase cleanly over my bugfixes. Want to push up a rebased version to kick travis?

rphmeier:
Done

aturon:
@rphmeier 

> I just noticed a strange discrepancy between the papers, so there may still be a hidden bug or two.

Have we resolved this discrepancy? Maybe looking at the [deque crate](https://github.com/kinghajj/deque) as @Amanieu mentioned would provide a good second opinion?

Amanieu:
When I added relaxed orderings to the deque crate I removed support for buffer shrinking because the paper describing the relaxed orderings did not support shrinking. In particular, adding buffer shrinking may affect the required memory orderings in `steal`.

alexcrichton:
Note that I think shrinking was something I added when I realized that buffers were leaked (way back when), but now that we actually gc buffers it may be possible to remove the shrinking aspect.

alexcrichton:
Oh wait nevermind that was a different algorithm I modified, not this one. Carry on!

src/sync/chase_lev.rs
rphmeier: Deque usese `size == a.size()`, which is equivalent to the new paper's check. I should probably change it to match.

rphmeier:
@Amanieu Specifically, which orderings? It seems to me that `steal` can load an old top, an old array, or both. If the loaded array is invalidated after the fact, the pinned epoch keeps it alive. 

Looking closer at the paper, section 4.2.2 defines the execution path for a generic resize. I've attempted to transcribe the resizing path into pseudocode (I personally found the symbols hard to read). I may have made errors, so please refer to the paper also.

```
if B-T >= size(x) -1:
  Read B from b
  Read T from t
  Read address of x from A
  call Resize:
  memory barrier
  Read address of x' from A
  Write Xi to x'[b]
  memory barrier
  Write B+1 to b

proc Resize:
  Copy all elems to new array
  memory barrier
  write address of x' to a
```

In Lemma 3 they show that these barriers do not have to be full fences.

The important information I notice here is that the resizing execution path never explicitly mentions what the new size of the array is, or even that it should necessarily be larger than the old. I think we can take this to mean that shrinking is supported in some form or another, as long as we follow the same protocol as we do while growing. Whether the current implementation adheres to the invariants described in the paper is probably still up for discussion.

@aturon As an aside, I think this section of the paper also clears up the inconsistency I've noticed before.

Amanieu:
@rphmeier The main issue with shrinking is that the proof relies on the fact that old buffers are not reused (Lemma 5).

That said after looking into some more, this is only a problem if you used a shared pool of buffer. From what I understand, crossbeam uses epoch-based garbage collection instead, which guarantees that a buffer is only freed once no more thieves are referencing it.

This means that you can use the shrinking algorithm from section 3 of the original Chase-Lev paper instead of the one from section 4. The code is section 3 is much simpler and is compatible with the relaxed memory orderings from the PPoPP paper.

aturon:
@rphmeier Any thoughts on @Amanieu's latest comment?

rphmeier:
I think that @Amanieu is right, we can probably switch to the code from section 3. I am pretty busy lately, but I can take a swing at porting it over in the next few days.

Amanieu:
In fact you could just copy the code from the deque crate and simply replace the linked list of old buffers with crossbeam's garbage collection.

rphmeier:
Looking at the code for `deque` now, it doesn't appear to have any code related to shrinking. I don't quite understand what you mean in that regard.

rphmeier:
I think I've addressed the points made here. I've left it as two separate commits, but I can squash once you're happy with the changes.

(I may have also borked the git history while rebasing. Let me know if it needs fixing)

src/sync/chase_lev.rs
```diff
@@ -269,9 +283,11 @@ impl<T> Drop for Deque<T> {
     fn drop(&mut self) {
         let guard = epoch::pin();
 
-        let t = self.top.load(SeqCst);
-        let b = self.bottom.load(SeqCst);
-        let a = self.array.swap(None, SeqCst, &guard).unwrap();
+        // exclusive access here means that we only need to enforce sequential consistency once.
+        fence(SeqCst);
```
Amanieu: A fence isn't required here. When drop is running you are already guaranteed to be the only thread still accessing the deque. This is guaranteed by fences in `Arc::drop`.

Amanieu:
You need to fix all the operations on top/bottom to use wrapping ops. Make sure you never compare them directly (`top < bottom`), but instead use `top.wrapping_sub(bottom) < 0'.

Amanieu:
Finally, the only thing I'm worried about here is that `epoch::pin` adds significant overhead to deque operations. This is because it adds a `fence(SeqCst)`, which adds quite a lot of overhead. In simple benchmarks which call `push` and `pop` many times, the largest source of overhead is the `mfence` instruction in `pop`. Adding a second one may significantly hurt performance.

src/sync/chase_lev.rs
rphmeier: Ah, I hadn't considered that `Arc` also enforced the sequential consistency. I had figured I should add it just to be safe, but it seems it can be removed now.

rphmeier:
My original draft of this PR had changed the ops to wrapping. @alexcrichton suggested we save that for a later change, which I was fine with. I'm not convinced that the papers do support wrapping operations.
https://github.com/aturon/crossbeam/pull/33#discussion_r50927368

I don't understand your point that we should never compare the two directly unless we do switch to wrapping ops. The PPoPP paper directly compares them, so I'd assume this is fine.

There isn't any real probability of overflow if we're using 64-bit integers (64 years of 4 billion operations per second). Unfortunately, we don't have Atomic[I | U]* types, so there is a possibility of overflow on 32-bit systems, where 4 billion ops per second would last about a second. What I have seen proposed is to periodically reset the counters when the queue is provably empty, but this is very tricky to get right. I don't know how to, at least.

Amanieu:
I would argue that using wrapping operations is a correctness issue for 32-bit, but it's ok to leave this for a later PR since it isn't really related to atomics.

Using wrapping ops is the correct approach because you only really care about the distance between top and bottom, not their actual values. It turns out that the paper is wrong and that it's fine to let both top and bottom overflow, as long as you are careful when comparing them (as I mentioned in my previous comment).

rphmeier:
@aturon @alexcrichton I can re-add wrapping ops for this PR if you'd like. Otherwise, I think it can be rebased and merged.

aturon:
Sorry for the delay -- I'm going to take this as-is, and we can continue to refine things like wrapping over time. Thanks again @rphmeier! (And @Amanieu and @alexcrichton for your help.)

PR #34, opened 2016-01-27
Add an is_empty() method for MsQueue

Ticki:


aturon:
Thank you!

aturon:
It looks like this actually broke some of the tests. Investigating now.

aturon:
I didn't look closely at the tests before merging, which was my mistake. Many of the assertions around emptiness for these tests were broken, due to not taking concurrency into account. I've pushed a fix.

Ticki:
Oh, sorry for that. It ran fine on my box...

PR #36, opened 2016-01-28
Fix bugs preventing memory reclamation

msullivan:
Fix two bugs that were preventing memory reclamation from occurring and another bug exposed by reclaiming the memory.

In my brief testing, performance impact is fairly small (and positive sometimes?). Might be worth regenerating the doc graphs anyways, though?

### DIFF
--- a/src/mem/epoch/garbage.rs
+++ b/src/mem/epoch/garbage.rs
@@ -94,7 +94,7 @@ impl Local {
     }
 
     pub fn size(&self) -> usize {
-        self.old.len() + self.cur.len()
+        self.old.len() + self.cur.len() + self.new.len()
     }
 }
 
--- a/src/mem/epoch/participant.rs
+++ b/src/mem/epoch/participant.rs
@@ -25,6 +25,10 @@ pub struct Participant {
     /// is ultimately used to free `Participant` records.
     pub active: AtomicBool,
 
+    /// Has the thread been passed to unlinked() yet?
+    /// Used to avoid a double free when reclaiming participants.
+    pub unlinked: AtomicBool,
+
     /// The participant list is coded intrusively; here's the `next` pointer.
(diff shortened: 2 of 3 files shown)


aturon:
Thanks so much!

Unfortunately, this appears to be failing some CI tests with a segfault, so there's still some trouble afoot.

msullivan:
Oh, oops, I only tested the benchmark program, not the test suite.

I think I've tracked down this crash, which also happens to be in lazy Participant reclamation code. The problem is that nodes in the participant list can get `unlinked` multiple times. If we have
`A --> B --> C --> null`, with `B` and `C` inactive, thread 1 looking at `A` and thread 2 looking at `B` (I think this requires `B` to become inactive while the list is being traversed), then, if thread 1 CASes `A's` next pointer to point at C and we wind up with
```
A --> C --> null
      ^
B-----|
```
Then threads 1 and 2 both successfully unlink `C` from the nodes they are looking at (`A` and `B`) and, having done that, call `unlinked` on `C`, which gets double freed.

msullivan:
I have a patch that I think works that addresses this issue by adding an `unlinked` boolean field to `Participant`. Before calling `unlinked`, we do a CAS on it to make sure it hasn't been unlinked before.

A variant of that is to generalize `active` to be an int instead of a bool and add another state for having been unlinked.

I don't love this solution; it seems hacky. But I think it should work and it has passed the test suite 300+ times in a row.

aturon:
Thanks again, @msullivan!

I agree both that this approach to participant reclamation seems plausible and also a bit hacky. I want think think a bit more deeply about an alternative representation for the list that allows for simpler cleanup. (Unfortunately I don't have a lot of time to work on crossbeam at the moment.) In the meantime, this PR is a clear improvement, and I'm going to merge.

PR #38, opened 2016-01-29
Added checks to garbage migration to avoid extra work

schets:
This will prevent the global bags from accepting an empty local bag, and will also stop garbage migration if there is no garbage to migrate (preventing the local collector from prematurely deallocating space in local bags)

More complex checks and paths could be added to migrate for doing checks and swaps on individual bags, but I don't know how much benefit that would bring.

### DIFF
--- a/src/mem/epoch/garbage.rs
+++ b/src/mem/epoch/garbage.rs
@@ -114,6 +114,9 @@ struct Node {
 
 impl ConcBag {
     pub fn insert(&self, t: Bag){
+        if t.len() == 0 {
+            return //don't insert an empty bag
+        }
         let n = Box::into_raw(Box::new(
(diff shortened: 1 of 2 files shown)


schets:
Not sure why this is failing on rust beta - I'll see if I can figure it out on my local machine. I don't see how my change would introduce a segfault. I had imagine any bugs would be memory leaks from not freeing memory.

aturon:
@schets I suspect the failure is related to a couple of bugs I'm trying to close with https://github.com/aturon/crossbeam/pull/42 -- after that lands, try rebasing.

schets:
This pull request only makes sense with local bags existing - I'll close it until there's a decidedly working version of local bags.

PR #42, opened 2016-02-02
Fix multiple segfault-causing bugs

aturon:
This commit closes two major holes, one related to the use of local
garbage bags, and the other to reclaiming participant data from exited
threads. After making these changes, I am no longer able to reproduce
segfaults or assertion failures. (I had previously build some aggressive
testing that was able to consistently segfault in a few seconds on every try.)

The use of local bags was creating a problem in which garbage could be
collected too early. It's crucial that when `unlinked` is called, the
data to be reclaimed is associated with the *current global
epoch* (and **not** the current local snapshot of the epoch). The reason
is that other threads may have outstanding pointers to the data, and may
be an epoch ahead; when the thread calling `unlinked` unpins, the epoch
could move forward again, causing the `unlinked` data to be collected
while pointers to it remain.

I was aware of this issue with the epoch algorithm, but Crossbeam
attempted to optimize things by using local bags of garbage nodes. These
local bags, however, did not manage to avoid the above problem.

This commit reverts the use of local bags (which has caused other
problems), returning to something closer to the original epoch algorithm
for the time being.

There was also a bug in freeing participant data from threads that have
exited after being involved in epoch management -- essentially, the code
removing a node from the middle of a linked-list was subject to race
conditions in which a previously-`unlinked` node could suddenly
reappear.

For the time being, I have simply removed all reclamation of participant
nodes. This is a memory leak, but a rather small one: it's proportional
to the number of threads that have ever used epochs and have since
terminated.

In a follow-up PR, I will try to roll out a working algorithm for
participant reclamation, likely based on marked pointers.

r? @alexcrichton

### DIFF
--- a/src/mem/epoch/participant.rs
+++ b/src/mem/epoch/participant.rs
@@ -1,12 +1,11 @@
 // Manages a single participant in the epoch scheme. This is where all
 // of the actual epoch management logic happens!
 
-use std::mem;
-use std::cell::UnsafeCell;
+use std::cell::Cell;
 use std::sync::atomic::{self, AtomicUsize, AtomicBool};
 use std::sync::atomic::Ordering::{Relaxed, Acquire, Release, SeqCst};
 
-use mem::epoch::{Atomic, Guard, garbage, global};
+use mem::epoch::{Atomic, Guard, global};
 use mem::epoch::participants::ParticipantNode;
 
 /// Thread-local data for epoch participation.
@@ -18,16 +17,12 @@ pub struct Participant {
     /// reentrant use of epoch management.
     in_critical: AtomicUsize,
 
-    /// Thread-local garbage tracking
-    garbage: UnsafeCell<garbage::Local>,
+    /// Number of operations since last GC
+    num_ops: Cell<u64>,
 
     /// Is the thread still active? Becomes `false` when the thread exits. This
     /// is ultimately used to free `Participant` records.
     pub active: AtomicBool,
-
-    /// Has the thread been passed to unlinked() yet?
-    /// Used to avoid a double free when reclaiming participants.
-    pub unlinked: AtomicBool,
 
     /// The participant list is coded intrusively; here's the `next` pointer.
     pub next: Atomic<ParticipantNode>,
@@ -35,14 +30,15 @@ pub struct Participant {
 
 unsafe impl Sync for Participant {}
 
+const GC_THRESH: u64 = 128;
+
 impl Participant {
     pub fn new() -> Participant {
         Participant {
             epoch: AtomicUsize::new(0),
             in_critical: AtomicUsize::new(0),
+            num_ops: Cell::new(0),
             active: AtomicBool::new(true),
-            unlinked: AtomicBool::new(false),
-            garbage: UnsafeCell::new(garbage::Local::new()),
             next: Atomic::null(),
         }
     }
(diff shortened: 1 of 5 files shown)


src/mem/epoch/participant.rs
```diff
@@ -74,7 +77,9 @@ impl Participant {
 
     /// Begin the reclamation process for a piece of data.
     pub unsafe fn reclaim<T>(&self, data: *mut T) {
-        (*self.garbage.get()).reclaim(data);
+        let cur_global = global::get().epoch.load(Acquire);
+        //let cur_global = self.epoch.load(Relaxed);
```
alexcrichton: commented out line

alexcrichton:
You'd know more than I but all the logic seems sound to me. Perhaps an issue could be opened about leaking participants?

otherwise lgtm

aturon:
Issue here: https://github.com/aturon/crossbeam/issues/37

Nit addressed, merging now.

msullivan:
I think I am missing something about the local bags bug. My reading of the situation is that after something gets added to thread A's local bag, thread A needs to witness three epoch advancements before the garbage gets removed, and I only see how two can happen.

So thread A `enter()`s, reads an epoch of 0. Thread B `enter()`s, advances the global epoch to 1, reads some pointer `p` out of a data structure. Thread A removes `p` from the data structure and calls `unlinked(p)`, `p` goes into A's `new` set. If A then `exit()`s and `enter()`s again, it will observe the global epoch is `1` and trigger a collection, moving `p` into `cur`. It can then immediately trigger a collection and advance the epoch to `2`, which will move `p` into `old`. But I don't see how the epoch could advance again before thread B `exit()`s, which would be required for `p` to get freed.

What am I missing?

schets:
Local is definitely free from this, but I had thought that migration to global could falsely move some garbage backwards and proposed a change to fix it.

Let's see...


1. Thread A enters, reads epoch of 0 (must happen before 2)
2. Thread B enters, advances epoch to 1, reads p (must happen before 3)
3. A unlinks p
4. A migrates garbage to global, putting new (and p with it) into global 1 ( while 0 is mayyyybe possible by memory orderings, I doubt the relaxed read of global [here](https://github.com/aturon/crossbeam/blob/d06ba1a26b471dfb8359ce2268b583e7ccf42dd3/src/mem/epoch/participant.rs#L114) was reordered that far back although who knows what the compiler did.
6. A exits.
7. Any non-B thread enters, updates the global epoch to 2 and GCs global 0. With some extreme reordering, this would cause issues for B. but I really doubt that's the case.

Another global GC can't happen until B exits, and then there's no problem. I also can't see some ordering of this where p somehow ends up in global 0 before the second global GC without crazy memory reordering.

Did the tests still segfault when only the fix to participant list was made?

msullivan:
@aturon could you check in or post the aggressive testing that you had causing consistent crashes?

aturon:
Hmm! I thought I'd found a possible execution that triggered this problem, but I'm unable to reconstruct it.

But yes, the segfaults happen even with just the "fix" to the participant list. They also continue to happen when I change *all* memory orderings to SeqCst.

As I think I've said elsewhere, I sadly have very limited time on crossbeam so (1) I really appreciate the help! and (2) I ended up just moving the whole library to a maximally-conservative position, with global garbage lists and participant leakage. I'm hoping we can start from this basis and make performance improvements over time.

I've just pushed up a file, `stress-msq.rs`, which is a variant of the spmc test. To quickly generate a segfault, used a little script to repeatedly launch copies of the test in parallel processes. It's a weird setup, but I get segfaults in < 10s on my 4 core machine.

Script:

```rust
use std::thread;
use std::process::{self, Command};
use std::io::prelude::*;

fn one_loop() {
    loop {
        let res = Command::new("target/release/stress-msq")
                      .status()
                      .unwrap();
        if !res.success() {
            println!("Bad exit: {:?}", res);
            std::io::stdout().flush();
            process::exit(1);
        }
    }
}

fn main() {
    println!("Slamming...");
    for _i in 0..3 {
        thread::spawn(one_loop);
    }
    one_loop();
}
```

msullivan:
Great, I'll take a look at that. I can dedicate a fair bit of time to crossbeam stuff since I'm planning to use it as the basis for a case study in my thesis.

msullivan:
Interesting. I haven't been able to reproduce a segfault yet but am able to get the `new > cur` assertion to fail in stress-msq after about a minute or so usually.

msullivan:
Are your tests using the 64-bit or 32-bit build?

msullivan:
Using nightly or something else?

aturon:
64-bit nightly.

msullivan:
Hm. I can not reproduce a segfault with participant reclamation disabled but with local bags still in use. I have observed an assertion failure on `new > cur`, in which a wrong value sometimes comes out of the queue, but infrequently. I think this might happen more often with local bags but I have observed it without them also.

aturon:
@msullivan I definitely am able to reproduce. Are you compiling in debug or release, btw? I was using release.

msullivan:
Release. Platform?

msullivan:
Weird. I have been *totally* unable to get https://github.com/msullivan/crossbeam/tree/crashes (crossbeam with the elimination of local garbage reverted) to segfault. Tested on two linux machines (a quad core and a quad core with HT) and one OS X machine (quad core).

schets:
I also can't reproduce the segfault, although I do get the assertion error after ~20 minutes. It seems like the queue needs some validating.

PR #45, opened 2016-02-07
Combined ready and data arrays in seg_queue

schets:
This combines the ready block and the data block in seq queue. Combining these has some good cache properties -

- Loads of the ready variable in pull will likely read the data into the cache, and same with loads of the data in push.
- Writes to the data and ready variable may well be combined.
- Elements are farther apart in the cache, so writes from pushers won't invalidate reads from pullers as much.

In my basic (and somewhat variable) tests, this saw performance gains of:

- mpmc: 5-15% increase
- spmc: 0-5% increase
- spsc: 15-35% increase

Also, this version is generally more consistent performance-wise probably due to less cache thrashing.

### DIFF
--- a/src/sync/seg_queue.rs
+++ b/src/sync/seg_queue.rs
@@ -19,8 +19,7 @@ pub struct SegQueue<T> {
 
 struct Segment<T> {
     low: AtomicUsize,
-    data: [UnsafeCell<T>; SEG_SIZE],
-    ready: [AtomicBool; SEG_SIZE],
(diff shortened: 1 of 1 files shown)


aturon:
Sorry for the slow review -- this looks good! I'm happy to merge after a rebase.

schets:
Just rebased!

aturon:
Lovely, and nice perf wins. Thanks again for your help with the library!

Issue #44, opened 2016-02-06
Program compiles but crashes


6D65:
Hi,

I know i'm most likely shooting myself into the foot. But I wanted to let you know. It's most likely that the created string inside the spawn goes out of scope or something. But this isn't caught by the compiler. I'm not sure wether it's the library or the compiler.

I guess there isn't a way out of this without using boxes.

Thanks

```Rust
extern crate crossbeam;

use crossbeam::sync::SegQueue;
use std::thread;
use std::time::Duration;

struct Task {
    text: String,
}

impl Task {
    pub fn new(text: String) -> Task {
        Task { text: text }
    }
}

fn main() {
    let seg_queue : SegQueue<Task> = SegQueue::new();

    seg_queue.push(Task::new("banana".to_string()));

    crossbeam::scope(|scope| {
        for i in 0..10 {
            let qr = &seg_queue;
            scope.spawn(move || {

                println!("Working 1");
                &qr.push(Task::new("Thread {}".to_string()));
                println!("Working 2");

            });
        }
    });

    loop {
        thread::sleep(Duration::new(0, 3));

        match seg_queue.try_pop() {
            Some(a) => println!("{}", a.text),
            None => break,
        }
    }

    println!("Exiting the program.");
}
```

schets:
Does this happen when using MsQueue? I think there's actually a bug in the current implementation of SegQueue in push, where [uninitialized memory is directly assigned to](https://github.com/aturon/crossbeam/blob/master/src/sync/seg_queue.rs#L65) instead of using ptr::write. I'm guessing that the program tries to drop whatever was at that location previously, but that was junk data.

schets:
I've confirmed that fixing the assignment also fixes your segfault - #45 or a pull request with just that fix should solve your problem.

aturon:
I've merged #47 and published a new version with the fix. @6D65, can you confirm this fixes the problem for you?

6D65:
Thank you @schets for the fix.
@aturon  the problem is indeed fixed.

PR #47, opened 2016-02-08
Fixed direct assignment to uninitialized memory

schets:
This removes a direct assignment to uninitialized memory. The assignment causes the program to try and drop the variable, which can cause issues - see #44 for an example.


aturon:
Good catch, thank you!

PR #48, opened 2016-02-08
Made it so that all producers in SegQueue will try to extend tail

schets:
This makes it so that any producer trying to push into the SegQueue can extend the tail if the current segment is full and the thread that filled the last element is stuck, or takes a long time to allocate, or for whatever reason takes a long time to extend the queue. Producers will try more and more frequently to update the tail to statistically lower the upper bound.

A producer can still block, but the window is far smaller -  it would have to get preempted between lines 61 and 62, which is a very small range and unlikely.

According to bench.rs, this doesn't really have a performance impact - if it did, it was within measuring variance.

This seems like a use case for a sort of mcas, since I think that would make it possible to perform this operation in a fashion where no thread could block producers - @aturon, do you think this would make sense, and could be implemented in a somewhat performant fashion in Rust? I'm not very familiar with the algorithm behind it.

### DIFF
--- a/src/sync/seg_queue.rs
+++ b/src/sync/seg_queue.rs
@@ -4,9 +4,11 @@ use std::{ptr, mem};
 use std::cmp;
 use std::cell::UnsafeCell;
 
-use mem::epoch::{self, Atomic, Owned};
+use mem::epoch::{self, Atomic, Owned, Shared, Guard};
 
 const SEG_SIZE: usize = 32;
+const MAX_TRIES: usize = 6;
+const MIN_TRIES: usize = 1;
 
 /// A Michael-Scott queue that allocates "segments" (arrays of nodes)
 /// for efficiency.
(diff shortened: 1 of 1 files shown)


aturon:
Thanks for your continuing work on crossbeam!

I have mixed feelings about this particular PR, though -- it's not a clear win to me. In particular, we end up trading some delay waiting for a writer for extra allocation, which still needs to be communicated back to other threads.

Since it also adds some complexity to the code, I'd prefer not to land this until we have more clear evidence of a performance win. Ideally we'd have a more robust benchmark suite for doing so.

(Sidenote: one thing I'd be interested in playing with is exponential backoff in the various retry loops, with benchmark measurements. I suspect there's some room for doing better on that front.)

re: MCAS, I'm not sure I'd want to apply it in this particular case, but I'd *love* to have an implementation in crossbeam!

schets:
That makes sense, this weekend I'll see if I can get an in-depth set of benchmarks written up so the effects of this can be better measured.

schets:
Some rough latency percentiles showed this didn't have much of an improvement in the tail, closing for now.

PR #49, opened 2016-02-08
Made SegQueue free memory

schets:
Unlink segments after the head advances past them


aturon:
Whelp, that's depressing.

PR #50, opened 2016-02-09
Don't advance the epoch in try_collect until after doing the collection.

### LINKED ISSUE #46: Check MsQueue for validity
Both @msullivan and I have gotten assertion errors in the stress-msq.rs test - for me, it takes a while, so this isn't an easy condition to hit, but this shouldn't ever happen. The assertion that we have both seen fail is the [new > cur == true assertion](https://github.com/aturon/crossbeam/blob/master/src/bin/stress-msq.rs#L28).

msullivan:
In try_collect(), it is important to reclaim memory from two epochs
ago *before* advancing the local epoch. As soon as the local epoch has
been advanced, another thread could advance the epoch and start
putting garbage in that same bag.

Closes #46.

### DIFF
--- a/src/mem/epoch/participant.rs
+++ b/src/mem/epoch/participant.rs
@@ -99,10 +99,10 @@ impl Participant {
             return false
         }
 
-        self.epoch.store(new_epoch, Relaxed);
         unsafe {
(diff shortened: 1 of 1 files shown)


aturon:
Thanks so much for tracking this down!

PR #51, opened 2016-02-09
Bring back local garbage bags.

msullivan:
This reverts much of 208bf821e70d432d18b2c1284da4aa2eb7bb46fc although
it also includes a few minor refactors.

### DIFF
--- a/src/mem/epoch/garbage.rs
+++ b/src/mem/epoch/garbage.rs
@@ -13,7 +13,93 @@ use std::sync::atomic::Ordering::{Relaxed, Release, Acquire};
 
 use mem::ZerosValid;
 
+/// One item of garbage.
+///
+/// Stores enough information to do a deallocation.
+struct Item {
+    ptr: *mut u8,
+    free: unsafe fn(*mut u8),
+}
+
+/// A single, thread-local bag of garbage.
+pub struct Bag(Vec<Item>);
+
+impl Bag {
+    fn new() -> Bag {
+        Bag(vec![])
+    }
+
+    fn insert<T>(&mut self, elem: *mut T) {
+        let size = mem::size_of::<T>();
+        if size > 0 {
+            self.0.push(Item {
+                ptr: elem as *mut u8,
+                free: free::<T>,
+            })
+        }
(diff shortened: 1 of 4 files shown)


src/mem/epoch/garbage.rs
```diff
@@ -13,7 +13,94 @@ use std::sync::atomic::Ordering::{Relaxed, Release, Acquire};
...
+        self.0 = data;
+    }
+}
+
+// needed because the bags store raw pointers.
+unsafe impl Send for Bag {}
+unsafe impl Sync for Bag {}
+
+/// A thread-local set of garbage bags.
+// FIXME: switch this to use modular arithmetic and accessors instead
```
aturon: Let's drop this FIXME, assuming you agree that the named bags are actually clearer here.

src/mem/epoch/garbage.rs
```diff
@@ -13,7 +13,94 @@ use std::sync::atomic::Ordering::{Relaxed, Release, Acquire};
...
+impl Local {
+    pub fn new() -> Local {
+        Local {
+            old: Bag::new(),
+            cur: Bag::new(),
+            new: Bag::new(),
+        }
+    }
+
+    pub fn reclaim<T>(&mut self, elem: *mut T) {
```
aturon: I'd like to rename this; probably just to `insert`.

aturon:
Looks good, modulo a couple of nits.

msullivan:
Nits addressed

msullivan:
(Am running some more intensive testing right now)

msullivan:
Tests are looking good

aturon:
Woo, merged! And thus (hopefully) ends quite the debugging saga.

gnzlbg:
> Woo, merged! And thus (hopefully) ends quite the debugging saga.

Would something like thread sanitizer have helped here?

aturon:
@gnzlbg It's hard to say. My experience with this kind of thing is that, usually, you just have to stare hard at the code and rethink the algorithm; it tends to be hard to find a meaningful smoking gun.

PR #52, opened 2016-02-10
Wait-free bounded spsc ringbuffer

schets:
Here's an spsc ringbuffer - it's fast, wait free, and doesn't require any allocations after the start. It doesn't use the epoch manager but it seems fitting.

The allocate/deallocate functions are kinda hacky, but I don't know of another stable way to get a raw vector.

### DIFF
--- a/src/bin/bench.rs
+++ b/src/bin/bench.rs
@@ -8,6 +8,7 @@ use std::time::Duration;
 use crossbeam::scope;
 use crossbeam::sync::MsQueue;
 use crossbeam::sync::SegQueue;
+use crossbeam::sync::{SpscBufferQueue, BufferConsumer, BufferProducer};
 
 use extra_impls::mpsc_queue::Queue as MpscQueue;
 
@@ -31,6 +32,23 @@ fn time<F: FnOnce()>(_f: F) -> Duration {
 fn nanos(d: Duration) -> f64 {
     d.as_secs() as f64 * 1000000000f64 + (d.subsec_nanos() as f64)
 }
+
+struct SpscDummy<T: Send> {
+    co: BufferConsumer<T>,
+    pr: BufferProducer<T>,
+}
+
+impl<T: Send> SpscDummy<T> {
+    pub fn new(s: usize) -> SpscDummy<T> {
+        let (pr, co) = SpscBufferQueue::new(s);
+        SpscDummy {
+            pr: pr,
+            co: co,
+        }
+    }
+}
+
+unsafe impl<T: Send> Sync for SpscDummy<T> {}
 
 trait Queue<T> {
     fn push(&self, T);
@@ -59,6 +77,21 @@ impl<T> Queue<T> for MpscQueue<T> {
                 Inconsistent => (),
             }
         }
+    }
+}
+
+impl<T: Send> Queue<T> for SpscDummy<T> {
+    fn push(&self, val: T) {
+        let mut temp = val;
+        loop {
+            match self.pr.try_push(temp) {
+                Ok(_) => break,
+                Err(rval) => temp=rval,
+            }
+        }
+    }
+    fn try_pop(&self) -> Option<T> {
+        self.co.try_pop()
     }
 }
 
@@ -150,6 +183,27 @@ fn bench_chan_mpsc() -> f64 {
     nanos(d) / ((COUNT * THREADS) as f64)
 }
 
+fn bench_queue_spsc<Q: Queue<u64> + Sync>(q: Q, scale: u64) -> f64 {
+    let d = time(||{
+        let qr = &q;
+        scope(|scope| {
+            scope.spawn(move ||{
+                for x in 0..COUNT*scale {
+                    qr.push(x)
+                }
+            });
+            for _ in 0..COUNT*scale {
+                loop {
+                    if let Some(_) = q.try_pop() {
+                        break;
+                    }
+                }
+            }
+        });
+    });
+    nanos(d) / ((COUNT*scale) as f64)
+}
+
 fn main() {
     println!("MSQ mpsc: {}", bench_queue_mpsc(MsQueue::new()));
     println!("chan mpsc: {}", bench_chan_mpsc());
@@ -158,6 +212,11 @@ fn main() {
 
     println!("MSQ mpmc: {}", bench_queue_mpmc(MsQueue::new()));
     println!("Seg mpmc: {}", bench_queue_mpmc(SegQueue::new()));
+
+    println!("BufferQueue spsc: {}", bench_queue_spsc(SpscDummy::new(512), 100));
+    println!("Seg spsc: {}", bench_queue_spsc(SegQueue::new(), 1));
+    println!("Ms spsc: {}", bench_queue_spsc(MsQueue::new(), 1));
+    println!("mpsc spsc: {}", bench_queue_spsc(MpscQueue::new(), 1));
 
 //    println!("queue_mpsc: {}", bench_queue_mpsc());
 //    println!("queue_mpmc: {}", bench_queue_mpmc());
--- a/src/sync/spsc_bufferqueue.rs
+++ b/src/sync/spsc_bufferqueue.rs
@@ -1,0 +1,398 @@
+//! SPSC ringbuffer
+
+use std::sync::atomic::Ordering::{Acquire, Release, Relaxed};
+use std::sync::atomic::{AtomicUsize, AtomicBool};
+use std::sync::Arc;
+use std::ptr;
+use std::mem;
+use std::cmp;
+use std::marker::PhantomData;
+use mem::CachePadded;
+
+#[inline(always)]
+unsafe fn deallocate<T>(ptr: *mut T, old_size: usize) {
+    Vec::from_raw_parts(ptr, 0, old_size);
+}
+
+#[inline(always)]
+unsafe fn allocate<T>(size: usize) -> *mut T {
+    let mut buf = Vec::with_capacity(size);
+    let ptr = buf.as_mut_ptr();
+    mem::forget(buf);
+    ptr
+}
+
+/// A single-producer, single consumer bounded wait-free ringbuffer queue
+///
+/// All operations on the buffer queue are wait-free,
+/// provided move/copy operations are waitfree.
+/// This queue does not allocate after constructions
+#[repr(C)]
+pub struct SpscBufferQueue<T: Send> {
+    // This is a pointer instead of a vector
+    // so that vector doesn't call constructors
+    data_block: *mut T,
+    size: usize,
+    _marker: PhantomData<T>,
+
+    // These dummies result in a tremendous performance improvement, ~300%+
+    _dummy_1: CachePadded<u64>,
+    // data for the consumer
+    head: AtomicUsize,
+    tail_cache: AtomicUsize,
+    prod_alive:AtomicBool, //seems weird, but consumer will read this
+
+    _dummy_2: CachePadded<u64>,
+    // data for the producer
+    tail: AtomicUsize,
+    head_cache: AtomicUsize,
+    cons_alive: AtomicBool, //seems weird, but producer will read this
+}
+
+unsafe impl<T: Send> Send for SpscBufferQueue<T> {}
+
+impl<T: Send> SpscBufferQueue<T> {
+    pub fn new(size: usize) -> (BufferProducer<T>, BufferConsumer<T>) {
+        let q = SpscBufferQueue {
+            data_block: unsafe{ allocate(size) },
+            size: cmp::min(size, (isize::max_value() - 1) as usize) + 1,
+            _marker: PhantomData,
+
+            _dummy_1: CachePadded::zeroed(),
+            head: AtomicUsize::new(0),
+            tail_cache: AtomicUsize::new(0),
+            prod_alive: AtomicBool::new(true),
+
+            _dummy_2: CachePadded::zeroed(),
+            tail: AtomicUsize::new(0),
+            head_cache: AtomicUsize::new(0),
+            cons_alive: AtomicBool::new(true),
+        };
+        let qarc = Arc::new(q);
+        let rtuple = (BufferProducer::new(qarc.clone()),
+                      BufferConsumer::new(qarc));
+        rtuple
+    }
+
+    /// Tries constructing the element and inserts into the queue
+    ///
+    /// Returns the closure if there isn't space
+    #[inline(always)]
+    fn try_construct<F>(&self, ctor: F) -> Result<(), F> where F: FnOnce() -> T {
+        let ctail = self.tail.load(Relaxed);
+        let mut next_tail = ctail + 1;
+        next_tail = if next_tail == self.size  { 0 } else { next_tail };
+        if next_tail == self.head_cache.load(Relaxed) {
+            let cur_head = self.head.load(Acquire);
+            self.head_cache.store(cur_head, Relaxed);
+            if next_tail == cur_head {
+                return Err(ctor);
+            }
+        }
+        unsafe {
+            let data_pos = self.data_block.offset(ctail as isize);
+            ptr::write(data_pos, ctor());
+        }
+        self.tail.store(next_tail, Release);
+        Ok(())
+    }
+
+    pub fn try_pop(&self) -> Option<T> {
+        let chead = self.head.load(Relaxed);
+        if chead == self.tail_cache.load(Relaxed) {
+            let cur_tail = self.tail.load(Acquire);
+            self.tail_cache.store(cur_tail, Relaxed);
+            if chead == cur_tail {
+                return None;
+            }
+        }
+
+        let mut next_head = chead + 1;
+        next_head = if next_head == self.size  { 0 } else { next_head };
+        unsafe {
+            let data_pos = self.data_block.offset(chead as isize);
+            let rval = Some(ptr::read(data_pos));
+            self.head.store(next_head, Release);
+            rval
+        }
+    }
+
+    pub fn capacity(&self) -> usize {
+        self.size - 1 //extra space added in ctor as buffer for head/tail
+    }
+}
+
+
+impl<T: Send> Drop for SpscBufferQueue<T> {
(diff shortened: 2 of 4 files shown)


schets:
The whole test suite occasionally freezes after the seq_queue tests (but they all pass) on one of my machines, so maybe hold off on this? Valgrind reports no errors, so I don't know if this commit is actually causing errors.

schets:
The above has been fixed, it was an off-by-one error in deallocating.

src/bin/bench.rs
```diff
@@ -160,6 +187,8 @@ fn main() {
     println!("MSQ mpmc: {}", bench_queue_mpmc(MsQueue::new()));
     println!("Seg mpmc: {}", bench_queue_mpmc(SegQueue::new()));
 
+    println!("SpscBufferQueue: {}", bench_spsc());
```
aturon: It'd be great to make this a generic bench like the others, so that the numbers can be compared across the data structures.

src/sync/spsc_bufferqueue.rs
```diff
@@ -0,0 +1,400 @@
...
+
+            _dummy_2: CachePadded::zeroed(),
+            tail: AtomicUsize::new(0),
+            head_cache: AtomicUsize::new(0),
+            cons_alive: AtomicBool::new(true),
+        };
+        let qarc = Arc::new(q);
+        let rtuple = (BufferProducer::new(qarc.clone()),
+                      BufferConsumer::new(qarc));
+        fence(SeqCst);
```
aturon: Why the fence?

src/sync/spsc_bufferqueue.rs
```diff
@@ -0,0 +1,400 @@
...
+    }
+
+    /// Tries constructing the element and inserts into the queue
+    ///
+    /// Returns the closure if there isn't space
+    #[inline(always)]
+    fn try_construct<F>(&self, ctor: F) -> Result<(), F> where F: FnOnce() -> T {
+        let ctail = self.tail.load(Relaxed);
+        let mut next_tail = ctail + 1;
+        next_tail = if next_tail == self.size  { 0 } else { next_tail };
```
aturon: Probably nicer to just use `%` (modular arithmetic) here.

src/sync/spsc_bufferqueue.rs
```diff
@@ -0,0 +1,400 @@
...
+#[repr(C)]
+pub struct SpscBufferQueue<T: Send> {
+    // This is a pointer instead of a vector
+    // so that vector doesn't call constructors
+    data_block: *mut T,
+    size: usize,
+    _marker: PhantomData<T>,
+
+    // These dummies result in a tremendous performance improvement, ~300%+
+    _dummy_1: CachePadded<u64>,
```
aturon: I wonder if it's worth splitting the consumer/producer field groups into their own structs, then wrapping those directly with CachePadded.

src/sync/spsc_bufferqueue.rs
```diff
@@ -0,0 +1,400 @@
...
+        let chead = self.head.load(Relaxed);
+        if chead == self.tail_cache.load(Relaxed) {
+            let cur_tail = self.tail.load(Acquire);
+            self.tail_cache.store(cur_tail, Relaxed);
+            if chead == cur_tail {
+                return None;
+            }
+        }
+
+        let mut next_head = chead + 1;
```
aturon: Again, `%` is preferred.

src/sync/spsc_bufferqueue.rs
```diff
@@ -0,0 +1,400 @@
...
+
+    pub fn capacity(&self) -> usize {
+        self.size - 1 //extra space added in ctor as buffer for head/tail
+    }
+}
+
+
+impl<T: Send> Drop for SpscBufferQueue<T> {
+    fn drop(&mut self) {
+        fence(SeqCst);
```
aturon: Why the fence?

src/sync/spsc_bufferqueue.rs
```diff
@@ -0,0 +1,400 @@
...
+
+impl<T: Send> BufferConsumer<T> {
+    pub fn new(queue: Arc<SpscBufferQueue<T>>) -> BufferConsumer<T> {
+        BufferConsumer {
+            spsc: queue,
+        }
+    }
+
+    /// Creates a new producer if the current one is dead
+    pub fn create_producer(&self) -> Option<BufferProducer<T>> {
```
aturon: Hmm -- I don't quite understand the motivation for this API design. Why not follow the design of Rust's standard channels, and simply provide a free function in the module that generates a producer/consumer pair?

aturon:
Thanks again for this PR! Left an initial round of comments.

Still to do: naming bikeshedding, reviewing the tests.

src/sync/spsc_bufferqueue.rs
schets: The mod is way more expensive than the rest of the push/pop - in my tests in increases the time for a push/pop pair from ~3-5 ns to 16 ns.

src/sync/spsc_bufferqueue.rs
schets: It was originally intended to ensure that the destructor only runs once all previous memory accesses have happened. However, since that's already guarded by the Arc, it's kinda pointless now.

src/sync/spsc_bufferqueue.rs
schets: It originally was so that users wouldn't have to think about ordering when sharing the queue across threads, but I realized now thinking about it either an external mechanism will be used or someone would have to enforce ordering on both the construction and queue load, not just construction. It's an edge case anyways and is much harder to trigger in Rust anyways without realizing it.

src/sync/spsc_bufferqueue.rs
schets: I thought about that, but am not sure if it's worth the more complicated datastructure definition/member functions. It would get rid of the drop warning, but that could also be accomplished with a single other datastructure wrapped by the dropping one.

I can see this getting harder to scale to a larger or more complicated object though.

src/sync/spsc_bufferqueue.rs
```diff
@@ -0,0 +1,398 @@
...
+        let rtuple = (BufferProducer::new(qarc.clone()),
+                      BufferConsumer::new(qarc));
+        rtuple
+    }
+
+    /// Tries constructing the element and inserts into the queue
+    ///
+    /// Returns the closure if there isn't space
+    #[inline(always)]
+    fn try_construct<F>(&self, ctor: F) -> Result<(), F> where F: FnOnce() -> T {
```
huonw: I suspect the use of `&self` here is problematic, as it allows calling `try_construct` *inside* the call to `ctor`, which will probably not behave like one expects. Specifically,

```rust
let prod: BufferProducer<i32> = ...;
prod.try_construct(|| {
     prod.try_construct(|| 1);
     2
})
```

I suspect, since this is spsc, all these methods should take `&mut self`, to express the "single" requirement statically and also defend against this problem.

schets: Good find! fixed in the next push

schets:
I was able to consistently reproduce the segfault that occurs in the benchmark, and I'm not sure if it's the fault of the spsc_queue or not.

What triggers the error is causing a 'large' allocation/deallocation after doing a decent amount of MSQ+Seg work.

The guilty code in bench is along the lines of 

```rust
    println!("MSQ mpsc: {}", bench_queue_mpsc(MsQueue::new()));
    println!("Seg mpsc: {}", bench_queue_mpsc(SegQueue::new()));

    println!("MSQ mpmc: {}", bench_queue_mpmc(MsQueue::new()));
    println!("Seg mpmc: {}", bench_queue_mpmc(SegQueue::new()));

    println!("BufferQueue spsc: {}", bench_queue_spsc(SpscDummy::new(512), 1));
...
```

And the segfault reliably occurs on deallocation of the spsc queue or during allocation/deallocating in the following test.

The ways to prevent this bug from happening are:

- Move the spsc_queue test to before the MSQ and Seg tests
- Shrink the size of the spsc buffer to be comparable to that of the Seg buffers
- Comment out both the MSQ and Seg tests
- Comment out the spsc_queue test


and while not really a preventions, shrinking the amount of push/pop pairs reduces the of seeing the bug as does adding dummy large allocations between the queue test and crossbeam tests.

Given that shrinking the amount of operations (and the amount of crossbeam allocations) reduces the bug frequency along with shrinking the spsc_queue to not be 'large', I'm inclined to think this may not be an issue of the spsc_queue but that some behavior in crossbeam is corrupting portions of the heap that deal with larger allocations. That moving the test and removing the crossbeam tests is interesting, but I don't feel that's as strong evidence.

On the other hand, I'm only able to replicate this using the spsc_queue. If I create large vectors, or even allocate vectors in one thread and deallocate them in another, I can't reproduce the error. I'm pretty sure that the block allocation method I used is correct, but maybe not? Also, no other tests or programs have exhibited such behavior. It's possible there's an error in the spsc_queue allocation/deallocation that only causes visible issues when it happens after the allocator has done a decent amount of work and the allocation is large.

arthurprs:
@schets I think I stumbled on a similar problem, probably the same. I'll try to construct a TOELF contained est case that reproduces the problem.

arthurprs:
So, I couldn't make a self contained test but I'll provide the code I was using when I found the issue (warning: code is awful).

Just clone `https://github.com/arthurprs/burst-trie/tree/thread_safe` and run `cargo test correct -- --nocapture`

Running it through gdb I can see where it gets stuck

```
#0  __lll_lock_wait () at ../nptl/sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
#1  0x00007ffff79c268d in _L_lock_1082 () from /lib/x86_64-linux-gnu/libpthread.so.0
#2  0x00007ffff79c2607 in __GI___pthread_mutex_lock (mutex=0x7ffff6c13d58) at ../nptl/pthread_mutex_lock.c:134
#3  0x0000555555608b5e in je_tcache_bin_flush_small ()
#4  0x000055555560942d in je_tcache_event_hard ()
#5  0x00005555555655a8 in burst_trie::heap::deallocate (ptr=0x7ffff5a99780 "\001\335\377\365\377\177", old_size=8, align=8) at ../src/liballoc/heap.rs:113
#6  0x0000555555573ef3 in burst_trie::raw_vec::RawVec<T>.Drop::drop (self=0x7ffff5ffe248) at ../src/liballoc/raw_vec.rs:567
#7  0x0000555555573dfe in alloc..raw_vec..RawVec$LT$crossbeam..mem..epoch..Shared$LT$map..ContainerNode$LT$collections..string..String$C$$u20$usize$GT$$GT$$GT$::drop.10450::h1e6f2d70add59bc3 ()
#8  0x0000555555573de6 in collections..vec..Vec$LT$crossbeam..mem..epoch..Shared$LT$map..ContainerNode$LT$collections..string..String$C$$u20$usize$GT$$GT$$GT$::drop.10447::hfc536de1ee2df8ce ()
#9  0x00005555555737ba in collections..vec..Vec$LT$crossbeam..mem..epoch..Shared$LT$map..ContainerNode$LT$collections..string..String$C$$u20$usize$GT$$GT$$GT$::drop.10442::hfc536de1ee2df8ce ()
#10 0x000055555557378b in burst_trie::mem::drop<collections::vec::Vec<crossbeam::mem::epoch::Shared<burst_trie::map::ContainerNode<collections::string::String, usize>>>> (_x=Vec<crossbeam::mem::epoch::Shared<burst_trie::map::ContainerNode<collections::string::String, usize>>>(len: 0, cap: 1)) at ../src/libcore/mem.rs:515
#11 0x0000555555573756 in burst_trie::mem::epoch::garbage::ConcBag::insert::free<crossbeam::mem::epoch::Shared<burst_trie::map::ContainerNode<collections::string::String, usize>>> (t=0x7ffff5a99780 "\001\335\377\365\377\177") at /home/arthurprs/.multirust/toolchains/nightly/cargo/registry/src/github.com-88ac128001ac3a9a/crossbeam-0.2.5/src/mem/epoch/garbage.rs:32
#12 0x0000555555591e09 in crossbeam::mem::epoch::garbage::ConcBag::collect (self=0x7ffff5830100) at /home/arthurprs/.multirust/toolchains/nightly/cargo/registry/src/github.com-88ac128001ac3a9a/crossbeam-0.2.5/src/mem/epoch/garbage.rs:59
#13 0x000055555558fbba in crossbeam::mem::epoch::participant::Participant::try_collect (self=0x7ffff5828000, guard=0x7ffff5ffe4c8) at /home/arthurprs/.multirust/toolchains/nightly/cargo/registry/src/github.com-88ac128001ac3a9a/crossbeam-0.2.5/src/mem/epoch/participant.rs:104
#14 0x000055555558f139 in fnfn (p=0x7ffff5828000) at /home/arthurprs/.multirust/toolchains/nightly/cargo/registry/src/github.com-88ac128001ac3a9a/crossbeam-0.2.5/src/mem/epoch/mod.rs:422
#15 0x000055555558ec21 in fnfn (e=0x7ffff5fff600) at /home/arthurprs/.multirust/toolchains/nightly/cargo/registry/src/github.com-88ac128001ac3a9a/crossbeam-0.2.5/src/mem/epoch/local.rs:32
#16 0x000055555558eb66 in crossbeam::thread::local::LocalKey<T>::with<closure,crossbeam::mem::epoch::Guard> (self=0x555555849fa0 <mem::epoch::local::LOCAL_EPOCH::h00a9358e1d604887Yma>, f=closure = {...}) at ../src/libstd/thread/local.rs:211
#17 0x000055555558eab8 in crossbeam::mem::epoch::local::with_participant<closure,crossbeam::mem::epoch::Guard> (f=closure) at /home/arthurprs/.multirust/toolchains/nightly/cargo/registry/src/github.com-88ac128001ac3a9a/crossbeam-0.2.5/src/mem/epoch/local.rs:32
#18 0x000055555558ea8d in crossbeam::mem::epoch::pin () at /home/arthurprs/.multirust/toolchains/nightly/cargo/registry/src/github.com-88ac128001ac3a9a/crossbeam-0.2.5/src/mem/epoch/mod.rs:413
#19 0x0000555555578007 in burst_trie::map::BurstTrieMap<K, V>::insert (self=0x7ffff5ffea80, key="h6UURGeD0b42OuFODiBO2tPY4Ebl11Yt3VBb5H54IMYQ32Q0nz5Eq0g8RPkpIELjz6hfJbh0r5tAmsYntq1gS8gfVN3D8dVseZ73DyGm8wDWxCew0vHrqc3n5oM16mvcG1S3bI6N3uYp2j4eJralgkQraCN602YNrdjskABWW0ZctzuYW2p3QpOSrN4pmvY5dLhrymDrxaqTI4J3ip8CGfrs3J6Ua76VmE0QT1mcULXklBYiwtrFdRrghfnDZEPK2GCj4yOU1NNibHgEe4NTpecqnSzCMdCScGaIeNzkkGCGgP2al6pxyGieIzKUGGNQZLF5G54sp9y20HVk9fcrrDqiS4islwzrX1z6SaxuNNjGScn4jZCB5bGI8YKdcfR0Ew5vwjnPs3x3PmnVrNxt6iaBiW6hgP7If7q8I3QQJdxfuzk5zBBIi6CNHA0J3HgIHPZ0Sg44IxhI6ga9YxFR2c", value=14392861986062705416) at src/map.rs:90
#20 0x0000555555570587 in burst_trie::map::tests::test_correctness () at src/map.rs:595
#21 0x00005555555afb07 in boxed::F.FnBox$LT$A$GT$::call_box::h16101388979507783664 ()
#22 0x00005555555b21fc in sys_common::unwind::try::try_fn::h444404887678121012 ()
#23 0x00005555555da26c in __rust_try ()
#24 0x00005555555d7a2e in sys_common::unwind::inner_try::h6de88a54d8fb341cz2t ()
#25 0x00005555555b257b in boxed::F.FnBox$LT$A$GT$::call_box::h5798114155829425016 ()
#26 0x00005555555dc2da in sys::thread::Thread::new::thread_start::ha2764ce591c12364WTy ()
#27 0x00007ffff79c0182 in start_thread (arg=0x7ffff5fff700) at pthread_create.c:312
#28 0x00007ffff74d747d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:111
```

aturon:
@arthurprs I'm confused -- it looks like you're hitting some sort of deadlock, rather than a segfault? At least, that's what I get when I try the branch.

@schets Thanks for the investigation. It could certainly be a bug somewhere in existing crossbeam code. I'll try to dig into it myself soon.

arthurprs:
@aturon yeah, I think it's due to some sort of corruption since there's one only thread running.

schets:
@arthurprs That looks similar to some of the backtraces that I found, and also backtraces that I found in previous memory corruption bugs

@aturon I saw segfaults of the same sort in different memory corruption bugs - I would suspect that some data on the boundary of an arena memory blob is overwriting adjacent bookkeeping data and this failure eventually manifests itself when jemalloc tries to take a lock. It is a bit suspicious that is consistently occurs in deallocation (and flushing arenas), but maybe that's just a coincidence (unfortunate placement of mutex/pointer to mutex?)

I've had trouble getting valgrind to catch the bug - it's super slow for the spsc queue, and the bug may only get triggered in a truly multithreaded environment

arthurprs:
helgrind spills some stuff but memcheck is not able to detect the corruption

https://gist.github.com/arthurprs/c26718285493161322b7

arthurprs:
@schets did you ever found out is this is actually a bug in crossbeam?

schets:
I haven't yet, but hopefully once I get my computer set up this week I'll have a reasonable dev environment to work on it

arthurprs:
I did another round of debugging today, hoping it could be related to drop flags or something, but it still happens. Funny enough it works perfectly if I force the usage of system allocator.

code: https://github.com/arthurprs/burst-trie/tree/thread_safe
for reproducing: `RUST_BACKTRACE=1 cargo test correct -- --nocapture`

https://paste2.org/WvdgOsJB

schets:
I'm going to close this until some of the outstanding bugs in the queues and participant lists get fixed, this memory issue is outright bizzare.

PR #61, opened 2016-03-19
Add an AtomicReference type

sfackler:
See https://github.com/sfackler/log4rs/commit/5ed2895bfe2ac92e30a4a5506bde98081bc85cce for a usage example.

I'm not sure `AtomicReference` is the right name - I stole it from Java, but we might want this to be called something else to indicate that it's tied up with `Arc`.

It's also a bit sketchy that we assume that `Arc<T>` is the same size as `usize`, but that seems preferable to defining a new smart pointer just for this.

### DIFF
--- a/src/sync/arc_cell.rs
+++ b/src/sync/arc_cell.rs
@@ -4,18 +4,18 @@ use std::sync::Arc;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
 /// A type providing atomic storage and retrieval of an `Arc<T>`.
-pub struct AtomicReference<T>(AtomicUsize, PhantomData<Arc<T>>);
+pub struct ArcCell<T>(AtomicUsize, PhantomData<Arc<T>>);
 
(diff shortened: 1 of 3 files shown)


aturon:
I'm fine with the bit of sketchiness here. This lib is overall pretty experimental, anyway.

`AtomicReference` isn't great -- it doesn't strongly suggest what's going on here. But I'm struggling to do better. Some ideas:

- `ArcCell` -- conveys the idea that this provides interior mutability over an `Arc`. Usually `Cell` indicates non-concurrent, though.
- `MutArc` -- a changeable `Arc`?
- ... I guess that's it :)

I think I lean toward `ArcCell`. It's a shame we never did manage to get a more coherent naming scheme between `Mutex`, `RwLock`, `Cell` and `RefCell`.

sfackler:
ArcCell seems fine - updated.

aturon:
Thanks @sfackler!

PR #65, opened 2016-04-11
Implement the `Debug` trait for all public types

### LINKED ISSUE #62: Implement Debug trait on all structs
`Debug` not being implemented means that types that use crossbeam internally can't derive Debug.

fitzgen:
This commit implements `std::fmt::Debug` for all public types. In most cases,
this was simply adding `#[derive(Debug)]` to the type and the types of its
members. However, there are two exceptions, where `Debug` is implemented
manually because it cannot be derived: (1) `Scope<'a>`, which uses `FnBox` which
is not `Debug`, and (2) `ScopedJoinHandle<T>`, which uses `JoinState` which uses
`std::thread::JoinHandle<T>` which is not `Debug`.

Fixes #62.

### DIFF
--- a/src/bin/extra_impls/mpsc_queue.rs
+++ b/src/bin/extra_impls/mpsc_queue.rs
@@ -40,12 +40,14 @@
 
 pub use self::PopResult::*;
 
+use std::fmt;
 use std::ptr;
 use std::cell::UnsafeCell;
 
 use std::sync::atomic::{AtomicPtr, Ordering};
 
 /// A result of the `pop` function.
+#[derive(Debug)]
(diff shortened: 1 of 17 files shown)


fitzgen:
Although, this compiled fine locally, it does not seem to on Travis CI...

fitzgen:
Just realized I was using Nightly, not Stable. Force pushed some updates that should make everything groovy.

aturon:
Away on vacation right now, will look next week. Thanks for the PR!

aturon:
Thanks for the PR!

PR #66, opened 2016-04-18
Added CompareAndSwap

Popog:
Added a compare-and-swap operation to Atomic. Replaced load+cas with a single compare_and_swap in MsQueue::push_internal
Also fixed a typo.


Vtec234:
Hm, `Atomic` has already got CAS support. What does this change?

ticki:
I'm closing as this is covered in the next major release.

PR #69, opened 2016-04-19
rename pop to try_pop with deprecation warning

chills42:
I'd love to just mark `pop` as deprecated, but that's not stable until 1.9, so I added some documentation and feature flagging so that we'd provide a warning where possible.


aturon:
Thanks much!

PR #72, opened 2016-05-04
WIP Add a steal_half method to sync::chase_lev

fitzgen:
Questions:
- We could change the signature to make callers supply a vec, rather than
  allocate our own in `steal_half`. This could potentially be more efficient if
  the caller is repeatedly retrying.
- Is this safe? I think so, but I'm not 100% sure. The paper mentions that it
  would be interesting to apply a steal-half operation to this deque, but they
  didn't do it. I couldn't find any other papers or libraries that implemented
  this operation either, although my google-fu may be weak in this regard. The
  most important part, as far as I can tell, is populating the results array
  before performing the CAS to ensure that we get the correct set of resulting
  items (after the CAS, the circular buffer could circle back over the slots
  we're stealing and overwrite them with new values). Everything seems to be
  pretty much the same as stealing one item, which leads to my next question.
- Should we factor out the common parts between `steal` and `steal_half`?
  There's a ton of duplication here, as I've authored it.

If you think this is a valuable operation to have, then I can write some tests
and fix this up.

### DIFF
--- a/src/sync/chase_lev.rs
+++ b/src/sync/chase_lev.rs
@@ -102,6 +102,19 @@ pub enum Steal<T> {
     Data(T),
 }
 
+/// When stealing half of the data, this is an enumeration of the possible
+/// outcomes.
+#[derive(PartialEq, Eq, Debug)]
+pub enum StealHalf<T> {
+    /// The deque was empty at the time of stealing.
+    Empty,
+    /// The stealer lost the race for stealing data, and a retry may return more
+    /// data.
+    Abort,
+    /// The stealer has successfully stolen some data.
+    Data(Vec<T>),
+}
+
 // An internal buffer used by the chase-lev deque. This structure is actually
 // implemented as a circular buffer, and is used as the intermediate storage of
 // the data in the deque.
@@ -136,6 +149,12 @@ impl<T> Stealer<T> {
     /// Steals work off the end of the queue (opposite of the worker's end)
     pub fn steal(&self) -> Steal<T> {
         self.deque.steal()
+    }
+
+    /// Steals half of the work off the end of the queue (opposite of the
+    /// worker's end)
+    pub fn steal_half(&self) -> StealHalf<T> {
+        self.deque.steal_half()
     }
 }
 
@@ -252,6 +271,40 @@ impl<T> Deque<T> {
                 Steal::Abort
             }
         }
+    }
+
+    fn steal_half(&self) -> StealHalf<T> {
+        let guard = epoch::pin();
+
+        let t = self.top.load(Acquire);
+        fence(SeqCst);
+        let b = self.bottom.load(Acquire);
+
+        let size = b - t;
+        if size <= 0 {
+            return StealHalf::Empty;
+        }
+
+        let half = size + 1 / 2;
+
+        unsafe {
+            let a = self.array.load(Acquire, &guard).unwrap();
+
+            let mut data = Vec::with_capacity(half as usize);
+            for i in t..t + half {
+                data.push(a.get(i));
+            }
+
+            if self.top.compare_and_swap(t, t + half, SeqCst) == t {
+                StealHalf::Data(data)
+            } else {
+                while data.len() > 0 {
+                    mem::forget(data.pop());
+                }
+                StealHalf::Abort
+            }
+        }
+
     }
 
     // potentially shrink the array. This can be called only from the worker.


fitzgen:
Maybe folks have tried this and it doesn't really give any better load balancing than stealing one item at a time does?

schets:
> Maybe folks have tried this and it doesn't really give any better load balancing than stealing one item at a time does?

I can't find the paper, but steal_half performs far better in cases where jobs don't generate many children.

src/sync/chase_lev.rs
```diff
@@ -254,6 +273,40 @@ impl<T> Deque<T> {
...
+        let t = self.top.load(Acquire);
+        fence(SeqCst);
+        let b = self.bottom.load(Acquire);
+
+        let size = b - t;
+        if size <= 0 {
+            return StealHalf::Empty;
+        }
+
+        let half = size + 1 / 2;
```
jeehoonkang: maybe `(size + 1) / 2`?

jeehoonkang:
> The most important part, as far as I can tell, is populating the results array
before performing the CAS to ensure that we get the correct set of resulting
items (after the CAS, the circular buffer could circle back over the slots
we're stealing and overwrite them with new values).

I "guess" this would kill the performance for some degree.  Maybe we can relax it by using two locations (just say `bottom1` and `bottom2`) for the purposes of the present `bottom` variable: (1) signalling that the circular buffer is reclaimed and reusable, and (2) signalling that elements before `bottom` are already stolen by some threads.

The implementation seems correct to my eyes, for the same reason that that of `steal` is correct.  Though I am not 100% confident as I don't have a formal proof yet...

src/sync/chase_lev.rs
fitzgen: Maybe ;P

fitzgen:
> The implementation seems correct to my eyes, for the same reason that that of steal is correct. Though I am not 100% confident as I don't have a formal proof yet...

Have you seen [Correct and Efficient Work-Stealing for Weak Memory Models](https://pdfs.semanticscholar.org/9c0e/c4b2cda00b6d5ea52e6b01a068fdc45b8dff.pdf)? For someone who is familiar with this area of research (which I am definitely not, but I think you are ;)) I assume it probably shouldn't be too hard to extend the proofs for this case.

I really enjoyed the *Promising* paper, btw 👍

jeehoonkang:
> Have you seen Correct and Efficient Work-Stealing for Weak Memory Models?

Yes I've read it, but the proof is just beyond my level and I couldn't grasp it..  Yet I understand the key idea, which is what the authors called "cumulativity", that implies it is impossible that concurrent invocations of `steal` and `pop` see old values of `bottom` and `top` at the same time.  The exact same logic applies to `steal_half` AFAICT.

These days I am working on reproving (weak) Chase-Lev deque in a (hypothetical) program logic, and I will try to come up with a proof of `steal_half`.

> I really enjoyed the Promising paper, btw 👍

I really thank you for reading the paper ;-)  As soon as the slide for POPL is done, I will share it.

src/sync/chase_lev.rs
jeehoonkang: I think this is indeed a bug.  The stealer effectively tries to steal all the remaining elements, breaking the invariant for the popper that the last element will not be stolen.  I think it should be `let half = (size + 1) / 2;`

stjepang:
Do we have any updates on this? I think Rayon might benefit from a `steal_half` method.

I did some googling, but couldn't find any papers about steal-half operation in Chase-Lev deques. Papers that do talk about steal-half usually have the following structure: they mention Chase-Lev as previous work, state that steal-half is a difficult operation to implement, and then present a totally new deque design that supports steal-half.

While papers never say that steal-half cannot be supported in Chase-Lev deques, the fact that apparently noone did it makes me think it's definitely non-trivial.

That said, we could still add a steal-half method to our Chase-Lev implementation that simply repeatedly calls `steal()`. Note that every call to `steal()` pins the current thread, so we could actually do steal-half by enclosing all those steals with a single call to `pin()`. That could be much faster than naively calling `steal()` multiple times in a row.

fitzgen:
Please steal this PR from my work queue :)

jeehoonkang:
@stjepang Would you please give me the references to the papers that says it's hard to support steal_half() in Chase-Lev?  Because still, @fitzgen's implementation looks correct to my eyes (albeit the comment I left).

stjepang:
@jeehoonkang 

The Chase-Lev [paper](https://pdfs.semanticscholar.org/3771/77bb82105c35e6e26ebad1698a20688473bd.pdf) simply says at the end:

"It may be interesting to see how our techniques are applied to other schemes that improve on ABP-work stealing such as the locality-guided work-stealing of Acar, Blelloch and Blumofe [1] or the steal-half algorithm of Hendler and Shavit [6]."

There is also a [paper](https://www.chargueraud.org/research/2013/ppopp/full.pdf) from 2013 that mentions Chase-Lev:

"Several years later the nonblocking algorithm was extended by Chase and Lev to support dynamic resizing  [10]. Their first and, to our knowledge, only proof of correctness of a nonblocking work-stealing algorithm is not trivial: it spans over thirty pages [11]. Moreover, there is, in the literature, no nonblocking algorithm which combines resizeability with other extensions, such as steal half, possibly owing to the complexity involved in extending the proof of correctness."

jeehoonkang:
@stjepang The [paper](https://pdfs.semanticscholar.org/9c0e/c4b2cda00b6d5ea52e6b01a068fdc45b8dff.pdf) @fitzgen mentioned is also published in PPoPP 2013, and this contains a proof of an implementation of Chase-Lev in C/C++11.  I am quite confident that the proof scales to `steal_half`.  Though I agree that this issue can only be answered decisively with a rigorous mathematical proof.  (The decades of research prove that "confident" means nothing in concurrency.)

AFAICT, Chase-Lev deque is difficult to formally reason about due to two reasons: (i) use of relaxed atomics, and (ii) use of SC fences.  I am interested in both difficulties, and these issues make Chase-Lev very interesting to me.  I will dig further into it..

For the time being, I prefer merging it (after fixing the bug I mentioned in the comment), but I agree not to do so if anyone objects to do that.

stjepang:
That's great then! :) No objections to merging from my side either.

Two things we should probably do that'd increase our confidence of correctness:
1. Reaching out to the authors of "Correct and Efficient Work-Stealing for Weak Memory Models" and asking about the steal-half operation.
2. Stress testing on a weakly ordered architecture.

jeehoonkang:
Closing it for the same reason with #148

Issue #41, opened 2016-02-01
Junction HashMaps: Lessons to be learned?


lilianmoraru:
There is no gitter set-up so I decided to open an issue.

Jeff Preshing(people might know him from his blog [Preshing on Programming](http://preshing.com/) or his [CppCon presentation](https://www.youtube.com/watch?v=X1T3IQ4N-3g)) has [announced](http://preshing.com/20160201/new-concurrent-hash-maps-for-cpp/) a new lock-free concurrent HashMap library called [Junction](https://github.com/preshing/junction).
His benchmarks indicate that it outperforms the other C++ concurrent HashMaps.

Are there lesson to be learned from it?
Crossbeam doesn't yet have a HashMap implementation, so I wanted to post to provide some inspiration.

oblitum:
I also had learned about that and also got curious about this :-)

aturon:
FWIW, my current plan is to start with a simple skiplist-based map, to focus first on the end-user API (which, in Rust, is not as straightforward as you might think). But in the long run, Rust needs a very high-perf hashmap, and I'd love to draw from the state of the art here!

schets:
I'm interested in how this does with many concurrent modifying threads. I'll run some of the benchmarks and look at the source later, but typically concurrent hashmaps suffer badly during the resize phase. Some more detailed benchmarking will reveal how the read/write performance scales individually - if the read performance is good, which I expect, leapfrog or grandpa would be well suited to a read-focused map. If write performance scales well I would be pleasantly surprised, but looking at the posted benchmark it's less than linear for a mix of read/writes so either something isn't scaling or he's using hyperthreads. Also, these use QBSR for a performance boost, something which doesn't really belong in a general purpose api.

The leapfrog table also appears exceptional for a single writer scenario. I think having a single-writer map api as well as a general one would be really great since the performance is so much higher for both for reads and writes and it's a common scenario.

As a side note on possible map datastructures, hash tries scale very well with multiple writers (almost linearly with real cores!), and I believe have wait-free reads, but suffer from lower read throughput and worse write throughput with few writers.

All in all, I don't think that Rust should just have one high-performance map. There should be a standard ConcurrentMap that has good read/write scalability and no performance surprises, but also a few different options like read-focused, single writer, or consistent latency for example.

schets:
[...]

PR #80, opened 2016-08-01
Add a MarkableArcCell to allow building reference-counted concurrent collections

Vtec234:
## What?

This PR adds a new type to `crossbeam::sync` - the `MarkableArcCell`. This type provides all the functionality of `ArcCell` and more. In particular, it allows one to mark the cell with a boolean value and then perform atomic `compare_exchange` operations on the `(Arc<T>, bool)` pair. It is based on a type from the `java.util.concurrent.atomic` module called the `AtomicMarkableReference` ([docs](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/atomic/AtomicMarkableReference.html)) and is supposed to provide the same kind of functionality.

I also made a tiny addition to `ArcCell` - `ArcCell::with_val(v)` is equivalent to `ArcCell::new(Arc::new(v))`. I think it's nicer not to have to expliciitly initialize the `Arc` each time, but if the change is unwanted I don't mind removing it as it's not too important.
## Why?

This type allows us to build concurrent lock-free collections as described in "_The Art of Multiprocessor Programming_" by M. Herlihy and N. Shavit. The book provides code in Java, which is a garbage-collected language. The big problem in writing such structures in non-GC languages is memory cleanup. In my opinion, the easiest to use equivalent of a GC is atomic reference counting, provided by `Arc<T>`.

I have already written a lock-free linked-list-based map and will publish it after this PR or its equivalent is merged, as the map depends on the `MarkableArcCell`. This will allow us to then create more performant maps, such as skiplist-based ones and stabilize the interface for a concurrent map in Rust, as discussed in #41, after which work on a lock-free hashmap could begin in `crossbeam`.

I have made some decisions regarding the interface of `MarkableArcCell` which might raise questions, so please ask and I will try to justify them.

### DIFF
--- a/src/sync/markable_arc_cell.rs
+++ b/src/sync/markable_arc_cell.rs
@@ -3,7 +3,7 @@ use std::mem;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
-/// A `MarkableArcCell` maintains an `Arc<T>` and a marker bit, providing atomic storage,
+/// A type, which maintains an `Arc<T>` and a marker bit, providing atomic storage,
 /// retrieval and CAS operations on the pair.
 #[derive(Debug)]
 pub struct MarkableArcCell<T> {
@@ -53,6 +53,8 @@ impl<T> MarkableArcCell<T> {
     }
 
     /// Returns the current values of both the `Arc` and the marker bit.
+    ///
+    /// This method is wait-free.
     pub fn get(&self) -> (Arc<T>, bool) {
         self.sem.fetch_add(1, Ordering::Relaxed);
         let t = untag_val(self.ptr.load(Ordering::SeqCst));
@@ -64,6 +66,8 @@ impl<T> MarkableArcCell<T> {
     }
 
     /// Returns the current value of the `Arc`.
+    ///
+    /// This method is wait-free.
     pub fn get_arc(&self) -> Arc<T> {
         self.sem.fetch_add(1, Ordering::Relaxed);
         let t = untag_val(self.ptr.load(Ordering::SeqCst)).0;
@@ -75,6 +79,8 @@ impl<T> MarkableArcCell<T> {
     }
 
     /// Returns the current value of the marker bit.
+    ///
+    /// This method is wait-free.
     pub fn is_marked(&self) -> bool {
         self.ptr.load(Ordering::Relaxed) & 1 == 1
     }
@@ -88,6 +94,7 @@ impl<T> MarkableArcCell<T> {
     /// `is_same(Arc::new(2), Arc::new(2)) == false`
     ///
     /// `let v = Arc::new(2); is_same(v.clone(), v) == true`
+    // TODO is this wait-free?
     pub fn compare_arc_exchange_mark(&self, current_t: Arc<T>, new_m: bool) -> bool {
         let current: usize = unsafe { mem::transmute(current_t) };
         drop::<Arc<T>>(unsafe { mem::transmute(current) });
@@ -147,24 +154,25 @@ impl<T> MarkableArcCell<T> {
 
 #[cfg(test)]
 mod test {
-    use std::sync::Arc;
+    use std::sync::{Arc, Barrier};
     use std::sync::atomic::{ATOMIC_USIZE_INIT, AtomicUsize, Ordering};
 
     use super::*;
+    use scope;
 
     #[test]
     fn basic() {
-        let r = MarkableArcCell::new(Arc::new(3), false);
+        let r = MarkableArcCell::new(Arc::new(1), false);
 
         let vals = r.get();
-        assert_eq!(*vals.0, 3);
+        assert_eq!(*vals.0, 1);
         assert_eq!(vals.1, false);
 
-        assert_eq!(*r.get_arc(), 3);
+        assert_eq!(*r.get_arc(), 1);
         assert_eq!(r.is_marked(), false);
 
         let prev = r.set(Arc::new(2), true);
-        assert_eq!(*prev.0, 3);
+        assert_eq!(*prev.0, 1);
         assert_eq!(prev.1, false);
 
         assert_eq!(*r.get_arc(), 2);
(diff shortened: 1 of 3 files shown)


src/sync/markable_arc_cell.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// Creates a new `MarkableArcCell` with the specified initial values of the `Arc` interior
+    /// and the marker bit.
+    pub fn with_val(v: T, m: bool) -> MarkableArcCell<T> {
+        MarkableArcCell(AtomicUsize::new(unsafe { mem::transmute(Arc::new(v)) }), AtomicBool::new(m), PhantomData)
+    }
+
+    // Locks the internal spinlock.
+    fn take(&self) -> Arc<T> {
+        loop {
+            match self.0.swap(0, Ordering::Acquire) {
```
schets: This isn't lock free, a suspended thread between a take/put operation can block arbitrarily.

schets:
What's the use case for this? The only literature that I found used this as some sort of atomic Optional. Are there uses for this where the nullness of the pointer and the marker status differ?

Is there a reason this must be reference counted instead of using the epoch memory management?

If there are any guarantees on heap pointer alignment in rust, this could also be done by using unused bits and would be implementable in a lock-free manner. If/when double word atomics are added they could make this current implementation lock free as well.

Vtec234:
> If there are any guarantees on heap pointer alignment in rust, this could also be done by using unused bits and would be implementable in a lock-free manner.

Yes, a tagged pointer would be an optimization over two values, but the reason I use a spinlock is not that there are two values. When you look at `ArcCell`, it also spinlocks. The problem is in the `Arc`->`AtomicUsize` conversion. 

Consider the situation:
Thread 1 reads the `usize` and transmutes it to an `Arc`. Assuming no other `Arc` copies are present, the refcount is `1`. Then it suspends. Thread 2 comes in, and uses `set_arc` (or `set` in `ArcCell`) to set a new value. `set_arc` returns the old value, which Thread 2 then drops. The refcount is `0`, and Thread 1 holds an invalid value.
The spinlock protects from this, since between a `take` and `put` the value is guaranteed to be valid.

The ability to make this lock-free therefore depends not on squashing the boolean into the unused bits of the `Arc` pointer, but on making `ArcCell` lock-free. It would be great if somebody could do that, unfortunately I have no idea how at the moment.

> If/when double word atomics are added they could make this current implementation lock free as well.

I wouldn't count on hardware support for this, unfortunately, but as explained above a DCAS is not necessary.

> What's the use case for this?

The problem I am trying to solve is that we need a stronger guarantee than just knowing that the memory we have a pointer to can be accessed, because while the memory may be valid, it might've been _logically_ deleted from the list.

When holding a `mem::epoch::Shared` pointer, we know that the node has not been deallocated, but we are not sure whether it is still in the list. Inserting a new node after a node that has been logically deleted will have no effect, and would therefore be invalid. A marker bit solves this, as we can use atomically make sure that we are setting the `next` pointer of the previous node to the new one only if the previous node's marker is `false`, which indicates still being in the list. A `remove` procedure would switch that bit to `true`.

My explanation is probably a bit wonky, so please see 4.3.3.1, p. 57 in [this](https://www.cl.cam.ac.uk/techreports/UCAM-CL-TR-579.pdf).

> Is there a reason this must be reference counted instead of using the epoch memory management?

Probably not, but to my knowledge epoch memory management would require a similar addition of a pointer-boolean pair for this to work, unless it already contains something like this. Due to the `MarkableArcCell` not actually being lock-free as you remarked, we might consider adding a markable pointer to `mem::epoch`.

src/sync/markable_arc_cell.rs
Vtec234: You're right, and we should consider other ways of creating a markable pointer.

schets:
I'm at work right now so this will be short, but I think that this commit is really doing two things at once - it's adding the idea of a markable pointer, and also the idea of an Atomic Arc. Would it make sense to separate them? If only a markable pointer was needed it would be far cheaper than the AtomicMarkableArc for loads/stores, and the AtomicArc also wouldn't have to deal with the overhead of being markable.

Vtec234:
> it's adding the idea of a markable pointer, and also the idea of an Atomic Arc.

If by AtomicArc you mean a type that provides atomic storage and retrieval of an Arc, then, well, it's already in crossbeam and it's called `ArcCell`. It's true that it doesn't provide CAS, but it might be extended to do that as well. Anyhow, this PR doesn't add that idea, as it's already there.

A markable pointer type is something that is indeed not yet present, and it's the aim of this PR to add one.

> If only a markable pointer was needed it would be far cheaper than the AtomicMarkableArc for loads/stores, and the AtomicArc also wouldn't have to deal with the overhead of being markable.

The aim of this PR is to build a markable pointer that is already integrated with a cleanup scheme (not simply a ptr-boolean pair that has to be freed like a normal ptr) and the current choice is ref-counting or the epoch GC.  It's built on top of `ArcCell`, because I want to try to build some ref-counted structures and see how they hold up.

An alternative to ref-counting, as you mentioned, is a GC (e.g. the epoch GC), and a similar marked pointer type might be added to `mem::epoch`, but it's a matter of benchmarking to see whether it is really cheaper - no refcounts, but GC sweeps, etc.
You also rightly said that it's not lock-free, so I made an `ArcCell` with a wait-free `get()` and I'll try to rewrite the markable variant on top of that.

Vtec234:
Finally got around to writing a more concurrency-friendly version of `MarkableArcCell`. In this one, all read methods are wait-free and all methods which write the `Arc` part are wait-free with respect to each other (other writes), but will lock until all readers are done. `compare_arc_exchange_mark`, which writes only the marker bit, may loop a little due to a CAS operation, so i'm not sure whether it can be classified as wait-free, but it's definitely lock-free.

With these guarantees, the guarantees that a skiplist as described in "_The Art of Multiprocessor Programming_" by M. Herlihy and N. Shavit implemented using this type would be:
`find`/`contains` - always lock-free
concurrent `remove`s and `insert`s - won't block
concurrent writes with `find`s - writes may block until reads are done

What do you think?

arthurprs:
I think this would be more appealing if you could link the code of the use cases you said to have. Code can be enlightening.

Vtec234:
@arthurprs You're right. I have uploaded a repository containing a concurrent linked list based on the `MarkableArcCell`: [`SortedLazyList`](https://github.com/Vtec234/lists-rs/blob/master/src/lib.rs). It's just a PoC, so I might've gotten my lock-freedom guarantees wrong, there might be undetected race conditions (even though the tests seem to pass), etc., but it demonstrates how data structures can be built based on this primitive. It's also terribly slow, but it could be extended relatively easily into the skiplist I mentioned earlier, and skiplists are not bad at all.
I'm sorry for taking so long, but I've been busy with uni and other silly things.

arthurprs:
That's great. But I'm confused, you say the linked list is terribly slow so how the skip list would fare better?

arthurprs:
Oh I see it's not actually a linked list, just similar name.

Vtec234:
Yes, skiplists are essentialy made by using several linked lists internally, each one with a different number of elements. Thanks to this and some other properties, skiplists scale much better than linked lists.

arthurprs:
I see the use case more clearly now, thanks. Does this scheme perform well in practice though?

Vtec234:
The linked list itself is not that useful, but good as a starting point. I used it in that way to write [a map implemented as a skiplist](https://github.com/Vtec234/lists-rs/blob/master/src/sorted_lazy_skiplist.rs) using pretty much the same techniques as the linked list does. The problem with that map is that it is only lock-free if the underlying primitive (`MarkableArcCell`) is lock-free, and at the moment it is not (Please disregard the documentation on my repo, it assumes that `MarkableArcCell` is lock-free. Also, I was wrong when stating here earlier that it is). It nevertheless seems to perform reasonably well, maintaining the `O(log n`) complexity of a skiplist, although I hadn't yet done any significant optimizations or benchmarked it properly, since I focused on correctness when writing it. Like the linked list, it's based on "_The Art of Multiprocessor Programming_".

Now, even though the skiplist is more or less fast, we want it to perform well under contention, which can only be guaranteed by lock-freedom. I can see two ways of achieving this. One is to fix `MarkableArcCell` to be lock-free. I am unsure whether that's even possible, but feel free to try.

The other way is to replace the memory recollection scheme with `crossbeam`'s epoch GC. That seems like the way to go for now, since plain pointers (`epoch::Atomic`) can trivially be paired with a `bool` (we cannot avoid the `Markable` part for skiplists) in a lock-free atomic primitive. I will probably PR something like `MarkableAtomic` soon and then try to rewrite the skiplist on top of that. This will have the additional advantage of being [most likely faster](http://csng.cs.toronto.edu/publication_files/0000/0159/jpdc07.pdf). (Link taken from aturon's excellent [blog post](https://aturon.github.io/blog/2015/08/27/epoch/)), but some disadvantages with respect to interface design (`&ValueType` cannot be returned, only `ValueType`).

__TL;DR__ concurrency is hard

arthurprs:
@Vtec234 should this be closed (same as #102)?

Vtec234:
@arthurprs This primitive is integrated with a different memory-recollection scheme - reference counting. If the authors of `crossbeam` do not want to look into this scheme, since it's quite a bit slower, this should be closed, but i'll keep it open for now as reference counting has some advantages over garbage collection, primarily having more predictable recollection times.

Vtec234:
More general types have been proposed in crossbeam-rs/rfcs#28.

PR #81, opened 2016-08-02
ArcCell: wait-free get()

Vtec234:
As an idea I got from @schets in #80, I wrote a slightly less locking implementation of `ArcCell`. In this version:
`get()` is always wait-free.
`set()` is wait-free with respect to other `set()`s. Whenever any `get()` arrives, the `set()` will block and have to wait until no concurrent `get()`s are active, behaving like the original implementation would in the degenerate case of the spinlock always going to `get()`s.
Since this will scale better for many readers, but most likely worse for many writers, I may also PR this as a new type, even though the change is backwards-compatible in terms of behaviour.

### DIFF
--- a/src/sync/arc_cell.rs
+++ b/src/sync/arc_cell.rs
@@ -41,7 +41,7 @@ impl<T> ArcCell<T> {
     pub fn set(&self, t: Arc<T>) -> Arc<T> {
         unsafe {
             let t: usize = mem::transmute(t);
-            let old: Arc<T> = mem::transmute(self.ptr.swap(t, Ordering::Relaxed));
+            let old: Arc<T> = mem::transmute(self.ptr.swap(t, Ordering::Acquire));
             while self.sem.load(Ordering::Relaxed) > 0 {}
             old
         }
--- a/src/sync/arc_cell.rs
+++ b/src/sync/arc_cell.rs
@@ -5,48 +5,56 @@ use std::sync::atomic::{AtomicUsize, Ordering};
 
 /// A type providing atomic storage and retrieval of an `Arc<T>`.
 #[derive(Debug)]
-pub struct ArcCell<T>(AtomicUsize, PhantomData<Arc<T>>);
+pub struct ArcCell<T> {
+    ptr: AtomicUsize,
+    sem: AtomicUsize,
+    _marker: PhantomData<Arc<T>>,
+}
 
 impl<T> Drop for ArcCell<T> {
     fn drop(&mut self) {
-        self.take();
+        unsafe { mem::transmute::<_, Arc<T>>(self.ptr.load(Ordering::Relaxed)); }
     }
 }
 
 impl<T> ArcCell<T> {
     /// Creates a new `ArcCell`.
     pub fn new(t: Arc<T>) -> ArcCell<T> {
-        ArcCell(AtomicUsize::new(unsafe { mem::transmute(t) }), PhantomData)
-    }
-
-    fn take(&self) -> Arc<T> {
-        loop {
-            match self.0.swap(0, Ordering::Acquire) {
-                0 => {}
-                n => return unsafe { mem::transmute(n) }
-            }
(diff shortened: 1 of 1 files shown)


arthurprs:
I was looking at this again today and was wondering if it makes sense to set the high bit if the writer can't acquire the lock after a while. A reader seeing that should backoff.

sstewartgallus:
If you already have an Arc then isn't what you want to implement https://en.wikipedia.org/wiki/Read-copy-update ?

schets:
@sstewartgallus RCU requires a memory management scheme to go along with the read-copy-update part. Imagine that you load the Arc ptr, stall, somebody drops the arc, and then resume trying to acquire the pointer. The Linux kernel scheme for doing so is often referred to RCU as a whole. Crossbeam EBR could also be used to manage the memory. This is managing it independently of the EBR with what's essentially a second refcounting scheme.

The orderings here are subtly incorrect. In get, the SeqCst load doesn't prevent the refcount increment from reordering past it, [see generated code here](https://godbolt.org/g/QOaWI1). SeqCst on a load is essentially an acquire load that's also ordered with respect to all other SeqCst operations and observe a global order of all SeqCst operations.

In an ideal world an Acquire on the fetch_add should work, but on aarch64 the load could still reorder behind the store part of the ll/sc pair (it's basically load-linked-acquire->store-conditional). aarch64 rmw orderings are extremely screwy and a major stretch of the c++11 atomic memory model, but oh well. I guess acquire only applies to the load ordering technically. I'm not sure of a performance-friendly way to do this without platform specific code.

src/sync/arc_cell.rs
```diff
@@ -5,48 +5,56 @@ use std::sync::atomic::{AtomicUsize, Ordering};
...
 
     /// Stores a new value in the `ArcCell`, returning the previous
     /// value.
     pub fn set(&self, t: Arc<T>) -> Arc<T> {
-        let old = self.take();
-        self.put(t);
-        old
+        unsafe {
+            let t: usize = mem::transmute(t);
+            let old: Arc<T> = mem::transmute(self.ptr.swap(t, Ordering::Acquire));
```
schets: This ordering is subtly incorrect, see my comment in the main thread for why. There's a crate atomic_utilities that has a fence/ordering combo that will create an optimal rmw setup that is ordered-before future loads and stores. I'm adding docs to it now but there's an example in the code.

arthurprs:
@schets I was checking this thread, what's wrong with the linked code? I get

```
load(std::atomic<int>&, std::atomic<int>&):
        nop nop nop

.L2:
        ldxr    w2, [x1]
        add     w2, w2, 1
        stxr    w3, w2, [x1]
        cbnz    w3, .L2
        nop nop nop

        ldar    w0, [x0]
        nop nop nop

.L3:
        ldxr    w2, [x1]
        sub     w2, w2, #1
        stxr    w3, w2, [x1]
        cbnz    w3, .L3
        nop nop nop

        ret

```

I know little about arm64 assembly but it doesn't look unordered to me.

schets:
@arthurprs It's possible for the ldar instruction to be ordered-before the stxr and I'm pretty sure the ldxr instruction, so it's possible for the SeqCst load to happen partially or completely before the fetch_add/swap. That's definitely a problem in get(), and I believe a problem in set as well (less clear on that one).

arthurprs:
Ahh I misunderstood the "reorder", you meant inside the execution units. Just like I say, my arm64 skills such.

You suggest using this, right? https://docs.rs/crate/atomic_utilities/0.5.0/source/src/fence_rmw.rs

schets:
That's what I've used for similar cases. It basically does:
  
  * acquire rmw on x86, since that's just an rmw + full fence all in one
  * relaxed rmw + seqcst fence, since every other architecture needs that sort of fencing or does it for acquire

Vtec234:
More general types have been proposed in crossbeam-rs/rfcs#28.

PR #84, opened 2016-08-27
Change mutability in `push` and `try_pop`

pmarcelll:
This makes it possible to use `chase_lev` as a drop-in replacement for
the deque in the `deque` crate, especially in `rayon`

### DIFF
--- a/src/sync/chase_lev.rs
+++ b/src/sync/chase_lev.rs
@@ -19,7 +19,7 @@
 //!
 //! ```
 //! use crossbeam::sync::chase_lev;
-//! let (mut worker, stealer) = chase_lev::deque();
+//! let (worker, stealer) = chase_lev::deque();
 //!
 //! // Only the worker may push/try_pop
 //! worker.push(1);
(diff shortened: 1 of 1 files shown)


rphmeier:
`Worker<T: Send>` is `Sync`, so your changes will allow safe code to violate the invariants of the queue (only one thread may push or pop).

pmarcelll:
I looked at what `deque˙ does, and they use a `PhanthomData`. https://github.com/kinghajj/deque/blob/master/src/lib.rs#L70
Is it a viable solution here?

rphmeier:
Yeah, that should be enough. Make sure to leave a comment like in `Deque`, so it isn't accidentally removed.

ticki:
This makes it non-concurrent, so I'm going to close it as WONTFIX.

pmarcelll:
@ticki can you elaborate? I'm fine with closing this PR but I don't see the concurrency problem here.

ticki:
> @ticki can you elaborate? I'm fine with closing this PR but I don't see the concurrency problem here.

Rust cannot have multiple `&mut` references, which is why every concurrent structure must do everything through internal mutability.

BurntSushi:
@ticki Could you elaborate more? I don't understand what you're talking about. The existing `chase_lev::Worker` takes `&mut self` and the existing `chase_lev::Stealer` takes `&self`. In contrast, the `deque::Worker` takes `&self` instead of `&mut`.

> which is why every concurrent structure must do everything through internal mutability.

That isn't under contention here.

ticki:
Oh, it's not shared between threads. I'll reopen then.

ticki:
I'm sorry.

PR #85, opened 2016-09-02
Relax an acquire load in Chase-Lev deque

jeehoonkang:
A relaxed load followed by an acquire fence is effectively an acquire
load.  This commit relaxes the ordering of an acquire load followed
by an SC fence, into a relaxed load.


jeehoonkang:
I misunderstood the Chase-Lev deque; this PR breaks the correctness of the Chase-Lev deque.  Sorry.

PR #96, opened 2016-12-27
Mark `src/bin/bench.rs` non-executable

### LINKED ISSUE #94: Please remove the executable bit from src/bin/bench.rs
src/bin/bench.rs has the executable bit set, which makes it look like a script with no `#!` line at the top. This confuses the Debian packaging lint tool "lintian", which complains about a file with the executable bit set but no ELF header or `#!` line.  Can you please remove the executable bit from that file?

apasel422:
Closes #94


alexcrichton:
Thanks!

PR #99, opened 2016-12-29
Standard library does not provide semaphores

dimbleby:
They were deprecated at 1.7.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,8 @@ freed. The `mem` module provides generic, easy to use, and high-performance APIs
 for managing memory in these cases.
 
 - **Synchronization**. The standard library provides a few synchronization
-primitives (locks, semaphores, barriers, etc) but this crate seeks to expand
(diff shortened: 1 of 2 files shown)


dimbleby:
This just fixes inaccurate docs.  

Since semaphores aren't in the standard library any more, would it be desirable for `crossbeam` to offer an implementation?

arthurprs:
@dimbleby I'd :+1: that.

sstewartgallus:
@dimbleby Event Semaphores or normal semaphores? I'm currently working on an event semaphore implementation at https://sstewartgallus.com/git?p=uevents-rs.git;a=blob;f=README;h=72f45cec4eaf19978cdad6259ab6a2bc6d894d82;hb=HEAD which has the potential to scale better for very high core machines but is currently still not well optimized enough.

Also, of consideration is whether or not the semaphores should respect thread priorities.

PR #101, opened 2017-01-02
Improve unlinked doc

Vtec234:
The epoch GC does not drop structures it reclaims the memory from, it only frees their memory. This behaviour seems to me non-obvious and since it's not documented I had to look for the associated code (https://github.com/aturon/crossbeam/blob/master/src/mem/epoch/garbage.rs#L43), so I propose specifying it in the documentation of `guard.unlinked(..)`.


arthurprs:
Shall we close this as #57 is on it's way?

PR #102, opened 2017-01-06
Add MarkableAtomic

Vtec234:
This PR adds a markable pointer `MarkableAtomic` similar to the one in #80, except that one is reference-counted and this one is meant to be used with the epoch GC. This type is strictly necessary to build concurrent lists and its behaviour cannot unfortunately be reproduced with just `Atomic`, since having an invalid, but not null (last bit marked) pointer in there would be unsafe. It has all the same methods as `Atomic`, plus an additional `unsafe from_ptr(ptr: *mut T, mark: bool) -> MarkableAtomic` for convenience so that we don't have to use a long sequence of loads and stores.

This type allows us to build lock-free concurrent skiplists, such as [this one](https://github.com/Vtec234/lists-rs/blob/master/src/epoch_skiplist.rs), which we could hopefully merge into crossbeam once it has a reasonable interface.

### DIFF
--- a/src/mem/epoch/markable_atomic.rs
+++ b/src/mem/epoch/markable_atomic.rs
@@ -51,7 +51,7 @@ fn opt_owned_into_usize<T>(val: Option<Owned<T>>) -> usize {
 impl<T> MarkableAtomic<T> {
     /// Create a new, null, markable atomic pointer with the given marker value.
     #[cfg(feature = "nightly")]
-    pub const fn null(mark: bool) -> MarkableAtomic<T> {
+    pub /*const*/ fn null(mark: bool) -> MarkableAtomic<T> {
         debug_assert!(mem::align_of::<T>() >= 2);
 
         MarkableAtomic {
--- a/src/mem/epoch/markable_atomic.rs
+++ b/src/mem/epoch/markable_atomic.rs
@@ -1,0 +1,220 @@
+use std::marker::PhantomData;
+use std::mem;
+use std::ptr;
+use std::sync::atomic::{self, Ordering};
+
+use super::{Owned, Shared, Guard};
+
+/// Like a `(std::sync::atomic::AtomicPtr, bool)` pair.
+///
+/// Provides atomic access to a pair consisting of:
+///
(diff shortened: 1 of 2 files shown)


Vtec234:
The fix for `--features="nightly"` is a bit artificial, but having the check for alignment is IMO more important than having it be `const`. Both can probably be included after `const fn`s are stabilized.

arthurprs:
This is much nicer and probably scales well in practice!

arthurprs:
I realized that there is an equivalent Wrapper in the PRhttps://github.com/aturon/crossbeam/pull/70

Vtec234:
Huh, thanks, I did not notice that. It also makes it quite clear that we absolutely need this type. I commented some more on that PR.

src/mem/epoch/markable_atomic.rs
```diff
@@ -0,0 +1,220 @@
...
+        debug_assert!(mem::align_of::<T>() >= 2);
+
+        MarkableAtomic {
+            val: atomic::AtomicUsize::new(tag_val(Box::into_raw(Box::new(data)) as usize, mark)),
+            _marker: PhantomData,
+        }
+    }
+
+    /// Unsafely create a new markable atomic pointer from the given value.
+    pub unsafe fn from_ptr(ptr: *mut T, mark: bool) -> MarkableAtomic<T> {
```
gereeter: `from_raw` seems like a slightly more appropriate name here.

src/mem/epoch/markable_atomic.rs
```diff
@@ -0,0 +1,220 @@
+use std::marker::PhantomData;
+use std::mem;
+use std::ptr;
+use std::sync::atomic::{self, Ordering};
+
+use super::{Owned, Shared, Guard};
+
+/// Like a `(std::sync::atomic::AtomicPtr, bool)` pair.
```
gereeter: This description is a bit inconsistent: the pointer is atomic and mutable, while the boolean in not. I think this line should either be cut (leaving the more precise description below) or be consistent, saying either `(AtomicPtr, AtomicBool)` or `(*const T, bool)`.

Vtec234:
All of this functionality is now contained in #70, closing.

PR #108, opened 2017-01-27
Don't try to gc except when we have just entered an epoch.

### LINKED ISSUE #105: Unsoundness bug: Thread-epoch can be advanced even if a pin is held
As this example demonstrates, the thread-local epoch can be advanced, even though we have the current epoch pinned:

```rust
extern crate crossbeam;

use crossbeam::mem::epoch::{self, Atomic};
use std::sync::{Arc, Barrier};
use std::sync::atomic::Ordering::Relaxed;
use std::thread;

const VALUE1: u32 = 42;
const VALUE2: u32 = 1337;

fn reset_ptr(ptr: Arc<Atomic<u32>>) {
    let g = epoch::pin();
    if let Some(shared) = ptr.swap(None, Relaxed, &g) {
        unsafe { g.unlinked(shared) };
    }
}

fn advance_epoch() {
    // This function uses unsafe, however in this case it is safe to do so,
    // or rather it should be if not for the unsoundness bug. We are are simply
    // allocating a new atomic pointer, setting it to null and asserting that
    // the old value is unreachable (which it is).
    let g = epoch::pin();

    for _ in 0..32 {
        let ptr = Atomic::new(VALUE2);
        let shared = ptr.swap(None, Relaxed, &g).unwrap();
        unsafe { g.unlinked(shared) };
    }
}

fn main() {
    let barrier = Arc::new(Barrier::new(2));
    let ptr = Arc::new(Atomic::new(VALUE1));

    {
        let barrier = barrier.clone();
        let ptr = ptr.clone();
        // The barriers are here to sync the threads so the use-after-free
        // happens deterministically. In most cases it would happen as a result
        // of a race condition.
        thread::spawn(move || {
            barrier.wait();
            barrier.wait();
            reset_ptr(ptr);
            barrier.wait();
        });
    }

    barrier.wait();
    let g = epoch::pin();
    let shared = ptr.load(Relaxed, &g).unwrap();
    barrier.wait();
    barrier.wait();

    // The thread has now finished and our shared pointer still contains the
    // original atomic value.
    assert!(**shared == VALUE1);

    // Run some unrelated code after loading the shared pointer. This code
    // should not be able to advance the epoch, since we have it pinned, however
    // it is still possible to do so.
    advance_epoch();
    advance_epoch();
    advance_epoch();

    // However as soon, our shared pointer has been freed and it now contains a
    // new value instead.
    assert!(**shared != VALUE1);
}
```

msullivan:
This fixes a bug where entering nested epochs could cause the
epoch to advance invalidly. Closes #105.

### DIFF
--- a/src/mem/epoch/guard.rs
+++ b/src/mem/epoch/guard.rs
@@ -22,13 +22,13 @@ pub struct Guard {
 /// pinning around the entire set of operations.
 pub fn pin() -> Guard {
     local::with_participant(|p| {
-        p.enter();
+        let entered = p.enter();
 
         let g = Guard {
             _marker: marker::PhantomData,
         };
 
-        if p.should_gc() {
+        if entered && p.should_gc() {
             p.try_collect(&g);
         }
 


aturon:
Thanks!

arthurprs:
@aturon This probably deserves a new release in crates.io

PR #114, opened 2017-03-07
Implement `Default` for `Atomic<T>` (returning null pointer).

ticki:
Following the style of `AtomicPtr<T>`, `*const T`, and `Option<T>`.


jonhoo:
r+

PR #115, opened 2017-03-07
Introduce `map` for `Shared<T>`.

ticki:
This is somewhat similar to owning_ref, but integrated into the pointer type, and much simpler as it is just a wrapper type.

It is similar to other methods named `map`, outside the crate.


jonhoo:
r+

PR #118, opened 2017-03-08
Add compare-and-swap and rename all the `cas_` prefixed methods.

ticki:
This effectively implements what is described in https://github.com/aturon/crossbeam/issues/116

**Warning**: This change is non-trivial in that it changes API (although without any breakage).


jonhoo:
r+

PR #119, opened 2017-03-08
Introduce owned epoch guards (`Pinned<T>`).

ticki:
Currently every epoch guard is usually used through a reference, making
it somewhat inconvinient for certain usecases, especially when you need
to expose an API, where you don't want the downstream user to have to
interact with the crossbeam API.

This patch solves this issue by introducing the `Pinned<T>` type, which
wraps `T` in a manner that only allow access through immutable
reference. `Pinned<T>` also holds an `epoch::Guard` meaning that `T` is
essentially "bound" to this guard.

Initially, I tried my hands at a `pin_then` method taking a closure, but
it proved inconvinient and possibly unsound, so instead I let
`Pinned<T>` be created through the methods of `Owned<T>` and
`Shared<T>`.

TLDR. This gets rid of the lifetime hell by providing a wrapper type,
which guarantees that a `Guard` exists while the inner value is held.

cc. @aturon @carllerche @alexcrichton

### DIFF
--- a/src/mem/epoch/guard.rs
+++ b/src/mem/epoch/guard.rs
@@ -1,4 +1,4 @@
-use std::marker;
+use std::{marker, ops};
 
 use super::{local, Shared};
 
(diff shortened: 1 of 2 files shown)


src/mem/epoch/guard.rs
```diff
@@ -12,14 +12,45 @@ pub struct Guard {
     _marker: marker::PhantomData<*mut ()>, // !Send and !Sync
 }
 
+/// A value pinned to an epoch.
+///
+/// This wraps an arbitrary type, such that it can only be indirectly access, while also holding a
+/// epoch guard, such that the value must span an epoch.
```
jonhoo: I think a documentation code example might come in really handy here.

PR #122, opened 2017-03-10
Major-release candidate

ticki:
**Changes**

- [x] Remove scoped threads.
- [x] Remove `chase_lev` as it is too complex to be in the crossbeam core crate (I plan making a side crate with this).
- [x] Get rid of `Owned<T>` in favor of `Box<T>`.
- [x] Give every single item detailed documentation.
- [x] Write nontrivial code in semi-literate style.
- [x] Refactor all the code to go from decent code to well above average code.
- [x] Fix several bugs.
- [x] Allow creation of `Shared<T>` from `Guard`.
- [x] Add owned guards through `Pinned`.
- [x] Remove the `mem` module, moving its items to the crate root.
- [x] Remove `ArcCell` (eventually moved to the secondary crate).
- [x] Quad-double the GC threshold to improve performance.
- [x] Audit all the code.
- [x] Rename CAS and CAS prefixed methods to compare-and-set.
- [x] Get rid of `cas()`.
- [x] Add real compare-and-swap.
- [x] Remove `AtomicOption`, which basically mirrored `epoch::Atomic`.
- [x] Remove all the deprecated methods.
- [x] Follow the naming conventions of libstd, renaming `try_pop` to `pop`.
- [x] Rename pop/push on queues to queue/dequeue to mark that queues are queues, not stacks.
- [x] Remove 3 use-after-frees in `Atomic<T>`.
- [x] Remove several memory leaks all around the code.
- [x] Remove the Scala benchmark, which seem to have no relavancy other than comparison.
- [x] Import modules over direct items.
- [x] Add `Default` to most of the types.
- [x] Add `map` to `Shared<T>`, allowing internal referencing with `owning_ref`.
- [x] Use `lazy_static` rather than manual implementation of it.
- [ ] Get rid of `CachePadded`

(was that all?)

Right now, I'm going to get some sleep. I don't have much time tomorrow, so I request someone to help me with this: The tests are broken, from what I suspect to be a simple bug in `CachePadded`, but I'm too tired to figure it out.


arthurprs:
Great work on all this.

aturon:
Nice! Thanks @ticki!

One note though: I don't think we should jump directly to 1.0. We can put out a new major release, but I feel like the library needs much more time to bake and grow before we give it 1.0 status. In particular, several people have talked about pulling out some of what's currently in the library into separate crates. And there are still segfaults and other problems in the issue tracker.

ticki:
@aturon How else will we justify the breakage?

aturon:
@ticki I'm not sure I understand? The current published version is 0.2. If we make this 0.3, that's considered a new major semver release (as per the rules about pre-1.0 versions), so it won't break anyone.

ticki:
> (as per the rules about pre-1.0 versions)

Ah, didn't know that convention were common in the Rust community.

aturon:
On another note, I also don't think we should rush to publish a new version to crates.io. It'd be good for people to get some experience with the new APIs, and to put some time into tracking down some of the pending issues. But I'll let others weigh in too.

ticki:
What about releasing `0.3.0-rc` to crates.io?

ticki:
I got to bed now, and won't be active before tomorrow afternoon. I encourage people to review changes and give feedback while.

djzin:
"Remove scoped threads" why is this? I thought this was a major feature of this crate

Vtec234:
@djzin: AFAIK the idea is to split them into a separate crate in rust-lang-nursery, since they are a feature used quite often and make crossbeam focus on data structures and lock-free algorithms.

src/sync/arc_cell.rs
```diff
@@ -1,90 +0,0 @@
...
-#[derive(Debug)]
-pub struct ArcCell<T>(AtomicUsize, PhantomData<Arc<T>>);
-
-impl<T> Drop for ArcCell<T> {
-    fn drop(&mut self) {
-        self.take();
-    }
-}
-
-impl<T> ArcCell<T> {
```
Vtec234: Why do you think the `ArcCell` is broken? It's certainly not lock-free, but it seems to work fine. I think we should keep it as it's in crossbeam's scope of concurrency primitives.

src/sync/chase_lev.rs
```diff
@@ -1,602 +0,0 @@
...
-unsafe impl<T: Send> Send for Deque<T> {}
-unsafe impl<T: Send> Sync for Deque<T> {}
-
-/// Worker half of the work-stealing deque. This worker has exclusive access to
-/// one side of the deque, and uses `push` and `try_pop` method to manipulate it.
-///
-/// There may only be one worker per deque, and operations on the worker
-/// require mutable access to the worker itself.
-#[derive(Debug)]
-pub struct Worker<T> {
```
Vtec234: I don't think moving this into a separate crate is a good idea. Either we restrict crossbeam to be purely a collection of primitives and move _all_ structures (queues, stack) out or we keep all structures. I'm leaning towards the latter, because the end result would be a batteries-included low-level concurrency crate containing both primitives and data structures rather than just something people have to build upon to end up with anything useful.

src/epoch/guard.rs
```diff
@@ -0,0 +1,140 @@
...
+impl Drop for Guard {
+    fn drop(&mut self) {
+        // Exit the epoch guard.
+        local::with_participant(|p| p.exit());
+    }
+}
+
+/// A value pinned to an epoch.
+///
+/// This wraps an arbitrary type, such that it can only be indirectly access, while also holding a
```
Vtec234: Small typo here, "access" should be "accessed".

src/epoch/mod.rs
```diff
@@ -135,99 +130,81 @@ mod participant;
...
-}
 
-impl<T> DerefMut for Owned<T> {
-    fn deref_mut(&mut self) -> &mut T {
-        &mut self.data
+    /// Convert it to a owned guard/pinned reference.
+    ///
+    /// This converts `self` into a value holding both `guard` and the reference, getting rid of
+    /// the bound epoch guard by instead storing it (owned) in the return value.
+    pub fn into_pinned(self, guard: Guard) -> Pinned<&'a T> {
```
Vtec234: How is this supposed to work? Here's how I thought this would be used, but it's not possible:
```rust
let grd: Guard = epoch::pin();
let shared: Shared<..> = some_atomic.load(Ordering::SeqCst, &grd);
let pinned: Pinned<..> = shared.into_pinned(grd); // Oops, grd is borrowed, cannot move out.
// Using another Guard here is unsound, since the Shared we obtained is only bound to `grd`.
```

src/epoch/garbage.rs
```diff
@@ -0,0 +1,210 @@
...
+impl Local {
+    /// Insert a new element to be deallocated eventually.
+    pub fn insert<T>(&mut self, elem: *mut T) {
+        self.new.insert(elem)
+    }
+
+    /// Collect one epoch of garbage and rotate the local garbage bags.
+    pub unsafe fn collect(&mut self) {
+        // All the old garbage will be destroyed.
+        let ret = self.old.collect();
```
sstewartgallus: Why do you need to set `ret` here? Doesn't `collect` return a `()`?

src/epoch/garbage.rs
```diff
@@ -0,0 +1,210 @@
...
+            data: t,
+            next: AtomicPtr::new(ptr::null_mut()),
+        }));
+
+        // Spin until the ABA problem is solved.
+        // TODO: Use striping to avoid this from spin multiple times (not as unlikely as you can
+        //       think, especially if several threads constantly creates new garbage).
+        loop {
+            // Load the head.
+            let head = self.head.load(atomic::Ordering::Acquire);
```
sstewartgallus: Why load the head twice here? `compare_and_swap` returns the newly updated version. See https://sstewartgallus.com/git?p=uevents.git;a=blob;f=src/node.h;h=0f84fde0094e179f1094789bd07c8754dfdfa198;hb=HEAD#l231 for an example.

src/epoch/garbage.rs
```diff
@@ -0,0 +1,210 @@
...
+            // Store the loaded head in the node which might become the new head node.
+            unsafe { (*n).next.store(head, atomic::Ordering::Relaxed); }
+            // To avoid leaking nodes, we use CAS to ensure that our snapshot (`head`) is still
+            // valid. If it is, we can replace the node with node `n` linking it and break the
+            // loop. If not, we will have to spin again until the ABA condition is over.
+            if self.head.compare_and_swap(head, n, atomic::Ordering::Release) == head {
+                break;
+            }
```
sstewartgallus: x86 should have a `pause` instruction here to reduce contention.

src/epoch/participants.rs
```diff
@@ -0,0 +1,146 @@
...
+
+        // We ultimately use epoch tracking to free `Participant` nodes, but we can't actually
+        // enter an epoch here, so fake it; we know the node can't be removed until marked inactive
+        // anyway.
+        let fake_guard = ();
+        let guard: &'static Guard = unsafe { mem::transmute(&fake_guard) };
+
+        loop {
+            // Load the head of the participant list we're traversing.
+            let head = self.head.load(atomic::Ordering::Relaxed, guard);
```
sstewartgallus: Like with ConcBag we load head twice in this loop. The compare and swap already returns the new value so we don't need to load head again. We should also use a pause instruction on x86.

src/sync/treiber_stack.rs
```diff
@@ -1,82 +1,107 @@
...
+            data: elem,
             next: Atomic::null(),
         });
+
+        // Pin the epoch.
         let guard = epoch::pin();
+        // In order to solve the ABA condition, we spin until the CAS succeeds.
         loop {
+            // Load the head node.
             let head = self.head.load(Relaxed, &guard);
```
sstewartgallus: Again we load the head twice when we already got it from compare and swap. Also we should use pause on x86.

src/cache_padded.rs
```diff
@@ -58,63 +42,102 @@ zeros_valid!(i8 i16 i32 i64 isize);
...
+    ///
+    /// The wrapped data is never dropped; move out using `ptr:read` if you need to run dtors.
     ///
-    /// **Warning**: the wrapped data is never dropped; move out using
-    /// `ptr:read` if you need to run dtors.
+    /// # Panics
+    ///
+    /// If `T` is incompatible with `CachePadded<T>` (e.g. too big), this will hit an assertion.
     pub fn new(t: T) -> CachePadded<T> {
+        // Assert the validity.
```
sstewartgallus: Useless comment.

src/cache_padded.rs
```diff
@@ -58,63 +42,102 @@ zeros_valid!(i8 i16 i32 i64 isize);
...
         assert_valid::<T>();
+
+        // Construct the (zeroed) type.
```
sstewartgallus: Useless.

src/cache_padded.rs
```diff
@@ -58,63 +42,102 @@ zeros_valid!(i8 i16 i32 i64 isize);
...
         let ret = CachePadded {
-            data: UnsafeCell::new(([0; CACHE_LINE])),
-            _marker: ([], marker::PhantomData),
+            data: UnsafeCell::new([0; CACHE_LINE]),
+            _pad: [],
+            _marker: marker::PhantomData,
         };
+
+        // Copy the data into the untyped buffer.
```
sstewartgallus: Useless.

src/epoch/atomic.rs
```diff
@@ -0,0 +1,233 @@
...
+use std::{marker, ptr, mem};
+
+use super::{Shared, Guard};
+
+/// Convert `Option<Shared<T>>` into `*mut T`.
+///
+/// `None` maps to the null pointer.
+#[inline]
+fn opt_shared_into_raw<T>(val: Option<Shared<T>>) -> *mut T {
+    // If `None`, return the null pointer.
```
sstewartgallus: Redundant.

src/epoch/atomic.rs
```diff
@@ -0,0 +1,233 @@
...
+    val.map(|p| p.as_raw()).unwrap_or(ptr::null_mut())
+}
+
+/// Convert `Option<Box<T>>` into `*mut T`.
+///
+/// `None` maps to the null pointer.
+#[inline]
+fn opt_box_into_raw<T>(val: Option<Box<T>>) -> *mut T {
+    // If `None`, return the null pointer.
```
sstewartgallus: Redundant.

src/epoch/atomic.rs
```diff
@@ -0,0 +1,233 @@
...
+    val.map(Box::into_raw).unwrap_or(ptr::null_mut())
+}
+
+/// Convert `&Option<Box<T>>` into `*mut T`.
+///
+/// `&None` maps to the null pointer.
+#[inline]
+fn opt_box_as_raw<T>(val: &mut Option<Box<T>>) -> *mut T {
+    // If `None`, return the null pointer.
```
sstewartgallus: Redundant.

src/epoch/atomic.rs
```diff
@@ -0,0 +1,233 @@
...
+    /// In order to perform the load, we must pass in a borrow of a `Guard`. This is a way of
+    /// guaranteeing that the thread has pinned the epoch for the entire lifetime `'a`. In return,
+    /// you get an optional `Shared` pointer back (`None` if the `Atomic` is currently null), with
+    /// lifetime tied to the guard.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `ord` is `Release` or `AcqRel`.
+    pub fn load<'a>(&self, ord: Ordering, guard: &'a Guard) -> Option<Shared<'a, T>> {
+        // Load the pointer.
```
sstewartgallus: What else would the code do? Useless comment.

src/epoch/atomic.rs
```diff
@@ -0,0 +1,233 @@
...
+        unsafe { self.ptr.load(ord).as_ref() }
+            // Construct the `Shared` pointer.
```
sstewartgallus: Useless comment.

src/epoch/atomic.rs
```diff
@@ -0,0 +1,233 @@
...
+    /// Atomically compare the value against `old` and swap it with `new` if matching.
+    ///
+    /// This is commonly refered to as 'CAS'. `self` is compared to `old`. If equal, `self` is set
+    /// to `new` and `Ok(())` returned. Otherwise, `Err(actual_value)` is returned (with
+    /// `actual_value` being the read value of `self`).
+    ///
+    /// All this is done atomically according to the atomic ordering specified in `ord`.
+    pub fn compare_and_swap<'a>(&self, old: Option<Shared<T>>, new: Option<Shared<T>>, ord: Ordering, guard: &'a Guard)
+        -> Result<(), Option<Shared<'a, T>>> {
+        // Convert `old` into a raw pointer.
```
sstewartgallus: Useless.

src/epoch/atomic.rs
```diff
@@ -0,0 +1,233 @@
...
+        let old = opt_shared_into_raw(old);
+        // Compare it against `new`.
```
sstewartgallus: Useless.

src/epoch/atomic.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    /// Compare-and-set from a `Shared` to an `Box` pointer with a given memory ordering and get
+    /// a shared reference to it.
+    ///
+    /// This operation is analogous to `store_and_ref`.
+    pub fn compare_and_set_ref<'a>(&self, old: Option<Shared<T>>, new: Box<T>, ord: Ordering, guard: &'a Guard)
+        -> Result<Shared<'a, T>, Box<T>> {
+        // FIXME: Ugly code.
+
+        // Cast the box into a raw pointer.
```
sstewartgallus: Useless.

src/epoch/atomic.rs
```diff
@@ -0,0 +1,233 @@
...
+        let new = Box::into_raw(new);
+
+        if self.ptr.compare_and_swap(opt_shared_into_raw(old), new, ord)
+            == opt_shared_into_raw(old) {
+            // Create a `Shared` pointer.
```
sstewartgallus: Useless.

src/epoch/atomic.rs
```diff
@@ -0,0 +1,233 @@
...
+            == opt_shared_into_raw(old)
+    }
+
+    /// Do an atomic swap with an `Box` pointer with the given memory ordering.
+    pub fn swap<'a>(&self, new: Option<Box<T>>, ord: Ordering, guard: &'a Guard)
+        -> Option<Shared<'a, T>> {
+        unsafe {
+            // Swap the inner.
+            self.ptr.swap(opt_box_into_raw(new), ord).as_ref()
+                // Construct the `Shared` pointer.
```
sstewartgallus: Useless.

src/epoch/atomic.rs
```diff
@@ -0,0 +1,233 @@
...
+        }
+    }
+
+    /// Do an atomic swap with a `Shared` pointer with a given memory ordering.
+    pub fn swap_shared<'a>(&self, new: Option<Shared<T>>, ord: Ordering, guard: &'a Guard)
+        -> Option<Shared<'a, T>> {
+        unsafe {
+            // Swap the inner.
+            self.ptr.swap(opt_shared_into_raw(new), ord).as_ref()
+                // Construct the `Shared` pointer.
```
sstewartgallus: Useless.

src/epoch/garbage.rs
```diff
@@ -0,0 +1,210 @@
...
+pub struct Bag(Vec<Item>);
+
+impl Bag {
+    /// Insert a pointer to be deallocated when the garbage is collected.
+    ///
+    /// This inserts `elem` into the bag of garbage with the destructor deallocating it.
+    fn insert<T>(&mut self, elem: *mut T) {
+        /// The freeing destructor.
+        unsafe fn dtor<T>(t: *mut u8) {
+            // Construct a vector and drop it.
```
sstewartgallus: Useless comment.

src/epoch/garbage.rs
```diff
@@ -0,0 +1,210 @@
...
+            drop(Vec::from_raw_parts(t as *mut T, 0, 1));
+        }
+
+        // Obtain the size of the element.
```
sstewartgallus: Useless comment.

src/epoch/garbage.rs
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    /// Get the number of items in this garbage bag.
+    fn len(&self) -> usize {
+        self.0.len()
+    }
+
+    /// Run all destructors of the garbage in the bag.
+    pub unsafe fn collect(&mut self) {
+        // Go over all the garbage, remove it and run the destructors.
```
sstewartgallus: Redundant with the public documentation.

src/epoch/garbage.rs
```diff
@@ -0,0 +1,210 @@
...
+        let n = Box::into_raw(Box::new(Node {
+            data: t,
+            next: AtomicPtr::new(ptr::null_mut()),
+        }));
+
+        // Spin until the ABA problem is solved.
+        // TODO: Use striping to avoid this from spin multiple times (not as unlikely as you can
+        //       think, especially if several threads constantly creates new garbage).
+        loop {
+            // Load the head.
```
sstewartgallus: Useless comment.

src/epoch/garbage.rs
```diff
@@ -0,0 +1,210 @@
...
+            // loop. If not, we will have to spin again until the ABA condition is over.
+            if self.head.compare_and_swap(head, n, atomic::Ordering::Release) == head {
+                break;
+            }
+        }
+    }
+
+    /// Run the destructor of all the garbage.
+    pub unsafe fn collect(&self) {
+        // Fetch the head node.
```
sstewartgallus: Useless comment.

src/epoch/garbage.rs
```diff
@@ -0,0 +1,210 @@
...
+        let mut head = self.head.load(atomic::Ordering::Relaxed);
+        // Make sure it isn't null (if it is, the stack is empty).
```
sstewartgallus: Just say "Make sure the stack isn't empty." The code already says that it's checking that the pointer isn't null.

src/epoch/garbage.rs
```diff
@@ -0,0 +1,210 @@
...
+    /// Collect the garbage of some epoch.
+    pub fn collect(&self, epoch: usize) {
+        // Find the associated array entry and collect it. Due to the invariants of this type, this
+        // is safe.
+        unsafe { self.bags[epoch % 3].collect() }
+    }
+
+    /// Insert a bag of garbage into some epoch..
+    pub fn insert(&self, epoch: usize, bag: Bag) {
+        // Find the associated array entry and insert the bag.
```
sstewartgallus: Instead of stating "find the associated array entry" why not just create a private method called `find_entry`?

src/epoch/guard.rs
```diff
@@ -0,0 +1,140 @@
...
+    // Set the current thread as participant while handling the epoch.
+    local::with_participant(|p| {
+        // Start the epoch.
+        let entered = p.enter();
+
+        // Construct the guard. This is safe as we have already called `enter`, ensuring that there
+        // is an active epoch for this thread.
+        let g = unsafe { Guard::fake() };
+
+        // Collect the garbage if needed.
```
sstewartgallus: Useless comment.

src/epoch/guard.rs
```diff
@@ -0,0 +1,140 @@
...
+///
+/// Pinning requires a full memory barrier, so is somewhat expensive. It is however rentrant -- you can
+/// safely acquire nested guards, and only the first guard requires a barrier.
+///
+/// In cases where you expect to perform several lock-free operations in quick succession, you may
+/// consider pinning around the entire set of operations.
+pub fn pin() -> Guard {
+    // Set the current thread as participant while handling the epoch.
+    local::with_participant(|p| {
+        // Start the epoch.
```
sstewartgallus: Useless comment.

src/epoch/local.rs
```diff
@@ -0,0 +1,56 @@
...
+    fn drop(&mut self) {
+        // Get the participant.
+        let p = unsafe { &*self.participant };
+
+        // Enter the active state.
+        p.enter();
+        // Propagate the garbage to the global state.
+        p.migrate_garbage();
+
+        // Exit the active state.
```
sstewartgallus: The code already _says_ exit the active state. You don't need to repeat it twice.

src/epoch/participant.rs
```diff
@@ -0,0 +1,176 @@
...
+
+impl Participant {
+    /// Enter a critical section.
+    ///
+    /// This method is reentrant, allowing for nested critical sections.
+    ///
+    /// It returns `true` is this is the first entry on the stack (as opposed to a re-entrant
+    /// call).
+    pub fn enter(&self) -> bool {
+        // Increment the counter.
```
sstewartgallus: Useless comment.

src/epoch/participant.rs
```diff
@@ -0,0 +1,176 @@
...
+        if self.in_critical.fetch_add(1, atomic::Ordering::Relaxed) > 0 {
+            // Since it was above zero, we return false.
```
sstewartgallus: This comment explains what but not why. Instead state something like "Since the counter was above zero indicate that we did not enter the critical section."

src/epoch/participant.rs
```diff
@@ -0,0 +1,176 @@
...
+                }
+            }
+
+            true
+        }
+    }
+
+    /// Exit the current (nested) critical section.
+    pub fn exit(&self) {
+        // Decrement the counter.
```
sstewartgallus: Useless comment.

src/epoch/participant.rs
```diff
@@ -0,0 +1,176 @@
...
+        self.in_critical.fetch_sub(1, atomic::Ordering::SeqCst);
+    }
+
+    /// Begin the reclamation process for a piece of data.
+    pub unsafe fn reclaim<T>(&self, data: *mut T) {
+        // Insert it into the garbage list.
```
sstewartgallus: The code already _says_ insert into the garbage list.

src/epoch/participant.rs
```diff
@@ -0,0 +1,176 @@
...
+        (*self.garbage.get()).insert(data);
+    }
+
+    /// Attempt to collect garbage by moving the global epoch forward.
+    ///
+    /// If it failed, it returns `Err(())`.
+    pub fn try_collect(&self, guard: &Guard) -> Result<(), ()> {
+        // Load the global epoch counter.
```
sstewartgallus: Useless comment.

src/epoch/participant.rs
```diff
@@ -0,0 +1,176 @@
...
+        // issue here.
+        // TODO: Is this really correct?
+        self.epoch.store(new_epoch, atomic::Ordering::Release);
+
+        Ok(())
+    }
+
+    /// Move the current thread-local garbage into the global garbage bags.
+    pub fn migrate_garbage(&self) {
+        // Load the current epoch counter.
```
sstewartgallus: Useless comment.

src/epoch/participant.rs
```diff
@@ -0,0 +1,176 @@
...
+        let cur_epoch = self.epoch.load(atomic::Ordering::Relaxed);
+        // Replace the local garbage set with an empty set.
```
sstewartgallus: You could have this comment or instead do something like.

~~~
let empty = garbage::Local::default();
let local = unsafe { mem::replace(&must *self.garbage.get(), empty) };
~~~

src/epoch/participant.rs
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    /// How many garbage items is this participant currently storing?
+    pub fn garbage_len(&self) -> usize {
+        unsafe { (*self.garbage.get()).len() }
+    }
+
+    /// Is this participant past its local GC threshold?
+    pub fn should_gc(&self) -> bool {
+        // We only GC when the garbage length is above some threshold.
```
sstewartgallus: Useless comment.

src/epoch/participants.rs
```diff
@@ -0,0 +1,146 @@
...
+        let mut participant = Box::new(ParticipantNode::default());
+
+        // We ultimately use epoch tracking to free `Participant` nodes, but we can't actually
+        // enter an epoch here, so fake it; we know the node can't be removed until marked inactive
+        // anyway.
+        let fake_guard = ();
+        let guard: &'static Guard = unsafe { mem::transmute(&fake_guard) };
+
+        loop {
+            // Load the head of the participant list we're traversing.
```
sstewartgallus: Useless comment.

src/epoch/participants.rs
```diff
@@ -0,0 +1,146 @@
...
+    next: &'a Atomic<ParticipantNode>,
+    // Has `self.next()` **not** been called before?
+    first: bool,
+}
+
+impl<'a> Iterator for Iter<'a> {
+    type Item = &'a Participant;
+    fn next(&mut self) -> Option<&'a Participant> {
+        // Load the next node.
+        let mut cur = self.next.load(if self.first {
```
sstewartgallus: At least GCC pushes all orderings up to sequential consistency if they're dynamically determined. It's still better style anyway to separate out the if anyway.

src/epoch/participants.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// The next node.
+    next: &'a Atomic<ParticipantNode>,
+    // Has `self.next()` **not** been called before?
+    first: bool,
+}
+
+impl<'a> Iterator for Iter<'a> {
+    type Item = &'a Participant;
+    fn next(&mut self) -> Option<&'a Participant> {
+        // Load the next node.
```
sstewartgallus: Useless comment.

src/sync/treiber_stack.rs
```diff
@@ -1,82 +1,107 @@
...
+        }
     }
 
-    /// Push `t` on top of the stack.
-    pub fn push(&self, t: T) {
-        let mut n = Owned::new(Node {
-            data: t,
+    /// Push an element on top of the stack.
+    pub fn push(&self, elem: T) {
+        // Construct the node.
```
sstewartgallus: Really, "Construct the node."?

src/sync/treiber_stack.rs
```diff
@@ -1,82 +1,107 @@
...
-
-    /// Attempt to pop the top element of the stack.
-    ///
-    /// Returns `None` if the stack is observed to be empty.
-    pub fn try_pop(&self) -> Option<T> {
+        // Pin the epoch.
         let guard = epoch::pin();
+        // Spin until the ABA condition is resolved.
         loop {
+            // Load the node.
```
sstewartgallus: Useless comment.

src/sync/treiber_stack.rs
```diff
@@ -1,82 +1,107 @@
...
+    /// This returns `None` if the stack is observed to be empty.
     pub fn pop(&self) -> Option<T> {
-        self.try_pop()
-    }
-
-    /// Attempt to pop the top element of the stack.
-    ///
-    /// Returns `None` if the stack is observed to be empty.
-    pub fn try_pop(&self) -> Option<T> {
+        // Pin the epoch.
```
sstewartgallus: Useless comment.

src/sync/treiber_stack.rs
```diff
@@ -1,82 +1,107 @@
...
         let guard = epoch::pin();
+        // Spin until the ABA condition is resolved.
         loop {
+            // Load the node.
             match self.head.load(Acquire, &guard) {
                 Some(head) => {
+                    // Load the tail node.
```
sstewartgallus: Referring this to as the tail node is confusing when in other contexts such as the queue the tail is not the next pointer.

src/sync/treiber_stack.rs
```diff
@@ -1,82 +1,107 @@
...
                     let next = head.next.load(Relaxed, &guard);
-                    if self.head.cas_shared(Some(head), next, Release) {
+                    // Set the node to the tail node, unless it changed (ABA condition).
+                    if self.head.compare_and_set_shared(Some(head), next, Release) {
                         unsafe {
+                            // Unlink the head node from the epoch.
                             guard.unlinked(head);
+                            // Read the data.
```
sstewartgallus: Useless comment.

src/sync/treiber_stack.rs
```diff
@@ -1,82 +1,107 @@
...
+            // Load the node.
             match self.head.load(Acquire, &guard) {
                 Some(head) => {
+                    // Load the tail node.
                     let next = head.next.load(Relaxed, &guard);
-                    if self.head.cas_shared(Some(head), next, Release) {
+                    // Set the node to the tail node, unless it changed (ABA condition).
+                    if self.head.compare_and_set_shared(Some(head), next, Release) {
                         unsafe {
+                            // Unlink the head node from the epoch.
```
sstewartgallus: Useless comment.

src/sync/treiber_stack.rs
```diff
@@ -1,82 +1,107 @@
...
         let guard = epoch::pin();
+        // Spin until the ABA condition is resolved.
         loop {
+            // Load the node.
             match self.head.load(Acquire, &guard) {
                 Some(head) => {
+                    // Load the tail node.
                     let next = head.next.load(Relaxed, &guard);
-                    if self.head.cas_shared(Some(head), next, Release) {
+                    // Set the node to the tail node, unless it changed (ABA condition).
```
sstewartgallus: Useless comment.

src/sync/treiber_stack.rs
```diff
@@ -1,82 +1,107 @@
...
                 }
+                // No nodes to pop.
                 None => return None,
             }
         }
     }
 
     /// Check if this queue is empty.
     pub fn is_empty(&self) -> bool {
+        // Pin the epoch.
```
sstewartgallus: Useless comment.

src/sync/treiber_stack.rs
```diff
@@ -1,82 +1,107 @@
...
         let guard = epoch::pin();
+        // Test if the head is a null pointer.
```
sstewartgallus: Useless comment.

src/epoch/participant.rs
```diff
@@ -0,0 +1,176 @@
...
+impl Participant {
+    /// Enter a critical section.
+    ///
+    /// This method is reentrant, allowing for nested critical sections.
+    ///
+    /// It returns `true` is this is the first entry on the stack (as opposed to a re-entrant
+    /// call).
+    pub fn enter(&self) -> bool {
+        // Increment the counter.
+        if self.in_critical.fetch_add(1, atomic::Ordering::Relaxed) > 0 {
```
schets: I expect this would make pinning more expensive. On Intel, you've essentially replaced a load->store->mfence sequence with a lock xadd -> mfence sequence, which isn't too different than a mfence->mfence sequence. Maybe the second fence is fast and can be ordered around since the previous fence already blocked everything, but it seems inefficient at first glance. The story is better but not great on ll/sc architectures. Since it's not a big change in clarity over load->add->store I think it should get changed back.

src/epoch/mod.rs
```diff
@@ -1,45 +1,40 @@
...
-//! pointers it extracts act like references with lifetime tied to the
-//! `Guard`. This allows threads to safely read from snapshotted data, being
-//! guaranteed that the data will remain allocated until they exit the epoch.
+//! - When the thread subsequently reads from a lock-free data structure, the pointers it extracts
+//!   act like references with lifetime tied to the `Guard`. This allows threads to safely read
+//!   from snapshotted data, being guaranteed that the data will remain allocated until they exit
+//!   the epoch.
 //!
 //! To put the `Guard` to use, Crossbeam provides a set of three pointer types meant to work together:
 //!
```
pitdicker: Now there are only two pointer types?

src/sync/ms_queue.rs
```diff
@@ -185,66 +212,83 @@ impl<T> MsQueue<T> {
...
             true
         }
     }
 
     /// Attempt to dequeue from the front.
     ///
-    /// Returns `None` if the queue is observed to be empty.
-    pub fn try_pop(&self) -> Option<T> {
+    /// This returns `None` if the queue is observed to be empty.
+    pub fn dequeue(&self) -> Option<T> {
```
ticki: Rust uses `push_front` and `push_back`, which is only appropriate when you can push/pop from both ends.

src/cache_padded.rs
```diff
@@ -58,63 +42,102 @@ zeros_valid!(i8 i16 i32 i64 isize);
...
+    _pad: Pad,
+    /// The typed marker.
+    _marker: marker::PhantomData<T>,
+}
 
-    /// A const fn equivalent to mem::zeroed().
+impl<T: ZerosValid> CachePadded<T> {
+    /// Create a zeroed cache padded value.
     #[cfg(feature = "nightly")]
     pub const fn zeroed() -> CachePadded<T> {
```
ticki: `Default`.

src/epoch/mod.rs
ticki: > // Using another Guard here is unsound, since the Shared we obtained is only bound to `grd`.

Not true. The `Shared` type only asserts that a guard is active.

src/epoch/garbage.rs
arthurprs: We punt on that for this version though.

src/sync/ms_queue.rs
arthurprs: I see no reason not to stick with the standard {try_}{push/pop} here.

src/sync/treiber_stack.rs
arthurprs: Overall I'd refrain from these comments, unless the comment is badly worded or incorrect. The code review surface is already huge to start nitpicking these.

src/sync/chase_lev.rs
arthurprs: We need to discuss some sort of guidelines.

src/epoch/garbage.rs
arthurprs: I see your point now, the first load can be moved to outside of the loop and you can updated it with the result of the cas.

Amanieu:
I have two general comments to make about the `Atomic` type:
- Consider basing the CAS function on the newer [`compare_exchange`](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicPtr.html#method.compare_exchange) API instead of the deprecated `compare_and_swap` one. The newer API allows you to specify a memory ordering on CAS failure and allows for weak CAS (`compare_exchange_weak`) which can be more efficient on certain architectures.
- Could we maybe rename `Atomic` to something more descriptive like `AtomicSharedPtr`? The name `Atomic` somewhat conflicts with my [`atomic`](https://github.com/Amanieu/atomic-rs) crate which provides a generic `Atomic` type.

src/cache_padded.rs
Vtec234: In that case how about adding `assert_valid` to `default` and `const fn zeroed` to address the `ZerosValid` size assertion issue?

src/sync/chase_lev.rs
Vtec234: I base my assumptions about the crate purpose on its README: 
> Ultimately the goal is to include stacks, queues, deques, bags, sets and maps.

src/mem/epoch/participants.rs
```diff
@@ -39,82 +46,99 @@ impl DerefMut for ParticipantNode {
...
+    // This ensures that we can free inactive nodes.
     guard: &'a Guard,
+    /// The next node.
     next: &'a Atomic<ParticipantNode>,
-
-    // an Acquire read is needed only for the first read, due to release
-    // sequences
-    needs_acq: bool,
+    // Has `self.next()` **not** been called before?
+    first: bool,
```
Vtec234: From a Reddit comment by /u/dbaupp: `first` is never set back to `false`.

src/epoch/participants.rs
```diff
@@ -0,0 +1,146 @@
...
+#[derive(Debug)]
+pub struct Iter<'a> {
+    // Pin to an epoch.
+    //
+    // This ensures that we can free inactive nodes.
+    guard: &'a Guard,
+    /// The next node.
+    next: &'a Atomic<ParticipantNode>,
+    // Has `self.next()` **not** been called before?
+    first: bool,
```
Vtec234: From a Reddit comment by /u/dbaupp: `first` is never set back to `false`.

Vtec234:
Given the size and complexity of this whole thing, I've come to think that it might be better if @ticki could split this into several smaller PRs. It's barely been reviewed and we're already crawling through an 80-comment thread. A potential organisation:

- Changes to EBR. Especially this need explanation, did you just refactor it and add documentation or did you make changes to algorithms? This PR should _not_ include moving the module and merging/splitting files, since the lack of diffs makes this especially difficult to review.
- Move `epoch` module, after above
- Add compare-and-swap, rename `cas`
- Add ability to create `Shared` from `Guard`, add `Pinned`
- Remove `Owned` and `AtomicOption`
- Remove `ArcCell`, Chase-Lev Deque and Scala benchmark
- Other fixes and addition I missed

What do people think?

ticki:
@Vtec234 I like that. It's definitely a mistake of mine to make such a big PR with a lot of unrelated changes.

src/sync/treiber_stack.rs
sstewartgallus: Fair enough. On later thought I think I sort of sounded kind of mean and I didn't mean that.

ticki:
btw, I've enabled such that all with write access can add comments

> If checked, users with write access to crossbeam-rs/crossbeam can add new commits to your master branch. You can always change this setting later.

Please feel free to use.

src/cache_padded.rs
konstin: It's not all "useless comments", it's @ticki's coding style. It's explained in the internals thread corresponding to this PR: https://internals.rust-lang.org/t/crossbeam-request-for-help/4933/16. 

After the experience of working with a huge undocumented code base, I do appreciate this style very much. And even if those comments are directly deducible from the code, having a human readable comment still improves the readability.

sstewartgallus: @konstin "it's somebodies coding style" is not an argument. It directly impedes readability to labal things redundantly. Any piece of the source code, even if it is a comment, should pay its weight or go to the chopping block. Most software source code metrics such as cyclomatic complexity and number of bugs directly correlate with the numbers of line of code. More code means more room for bugs and out of date documentation.

gnzlbg:
Are the test running with the thread sanitizer already on CI by default already?

jeehoonkang:
I strongly agree with @Vtec234 's suggestion to split this PR (https://github.com/crossbeam-rs/crossbeam/pull/122#issuecomment-285875721).  I would like someone to close this PR, and make a tracking issue for each change item mentioned above.

Having said about "someone", I feel this project currently lacks a proper decision making procedure, even for closing issues and merging PRs :\  For now I would like to ask if @ticki could provide leadership for this PR (which happens to be currently the most important one).  If you are in short of time, please kindly tell me what you want me to do :)  Of course, for a longer term, we eventually need to establish some institution.

src/cache_padded.rs
drewhk: I agree with @konstin. Concurrency code can benefit from extensive documentation, higher granularity than usual if necessary.

> Most software source code metrics such as cyclomatic complexity and number of bugs directly correlate with the numbers of line of code.

I don't see how comments would contribute to cyclomatic complexity and code complexity in general.

> More code means more room for bugs and out of date documentation.

So conversely, zero documentation means no room for being out-of-date? There is obviously a balance here. With core, concurrency related code, I prefer heavy commenting. Even tiny things can have significance. Also, out-of-date comments should be less of an issue as refactoring core concurrency code should be infrequent and heavily reviewed anyway.

Just my 2c.

jeehoonkang: I generally agree with @drewhk that extensive documentation helps a lot, especially for concurrency code.  But for this specific comment, I agree with @sstewartgallus : `Assert the validity.` doesn't give us additional information than the code itself (`assert_valid::<T>();`).  Surely it would be very helpful to write down the reason why we need to assert the validity.

drewhk: Sure, there is a discussion to be had about the _right_ balance here ;) I just wanted to voice my concern about blanket statements about comment granularity.

ticki:
@jeehoonkang

> If you are in short of time, please kindly tell me what you want me to do :) Of course, for a longer term, we eventually need to establish some institution.

I really want some help with one thing in particular: The bug in `CachePadded<T>`, which causes most tests to fail. I suspect it's just some stupid bug, but having looked at the code too long makes it really hard to get new insides. I'd like someone to help me figure out why it is that `assert_valid` panics. The `CachePadded` is pretty hacky, so maybe it's just some change in memory layout.

Either way it's hard for me to do alone.

aturon:
@jeehoonkang 

> Having said about "someone", I feel this project currently lacks a proper decision making procedure, even for closing issues and merging PRs

See https://github.com/crossbeam-rs/crossbeam/issues/124 for the time being. We'll have to play it a bit by ear for a while, I think, and gradually formalize a procedure.

src/mem/cache_padded.rs
```diff
@@ -1,48 +1,32 @@
...
-/// **Warning**: the wrapped data is never dropped; move out using `ptr::read`
-/// if you need to run dtors.
-pub struct CachePadded<T> {
-    data: UnsafeCell<[usize; CACHE_LINE]>,
-    _marker: ([Padding; 0], marker::PhantomData<T>),
-}
+/// Do not let your code's safety or correctness depend on this being the real size. The actual
+/// value may not be equal to this.
+// TODO: For now, treat this as an arch-independent constant.
+const CACHE_LINE: usize = 64;
```
jeehoonkang: In x86-64, the size of `CachePadded<T>` changed from 256 (`sizeof(usize) * CACHE_LINE` = 8 * 32 = 256) to 64 (`sizeof(u8) * CACHE_LINE = 1 * 64 = 64).  But 64 bytes are not sufficient: e.g. `Participant`, which is fed to `CachePadded`, is 104 bytes big.  That causes all the test failures.

The tests still fail even after fixing this by setting `const CACHE_LINE: usize = 256;`, though..

ticki: Holy fuck! So simple.

> The tests still fail for other reasons even after fixing this by setting const CACHE_LINE: usize = 256;, though..

All the tests run for me...

ticki: Oh, no. I might be wrong.

ticki: Yep. I was one the wrong `HEAD`.

src/epoch/garbage.rs
sstewartgallus: Just for future notes: I'm trying to add `pause` to the standard libs at this issue if anyone has any comments https://github.com/rust-lang/rust/pull/40537

arthurprs:
I spend a few hours on this with no success, most commits left the build broken and then a latter commit fixes it. So git bisect was useless. I tried getting each commit compiling but after a few hours I gave up, but that's probably the only way to find it without eyeballing.

whataloadofwhat:
>  Use `lazy_static` rather than manual implementation of it.

Why? `lazy_static` blocks in its implementation. Using a locking algorithm to implement lock-free data structures seems counter-intuitive, even if it is unlikely to cause an issue. The manual implementation was optimistic and lock-free, which matches the ideology of the crate a lot better I think.

> Get rid of `Owned<T>` in favour of `Box<T>`

Not sure I agree with this. `Owned<T>` is currently implemented as a `Box<T>` but that's an implementation detail that I find worrying. Realistically, `Owned<T>` is akin to `Box<ManualDrop<T>>`. I'd much rather the code work *with* this rather than just assume that `Vec<T>` with len 0 and cap 1 is the same as just freeing the memory of `Box<T>`. It feels hacky and confusing in a codebase that is already difficult to follow (not anybody's fault, concurrency is hard).

> Add owned guards through `Pinned`.

I feel like this is a mistake. It's not unsafe or anything, I just think people will use it incorrectly. 

> Add map to `Shared<T>`, allowing internal referencing with `owning_ref`.

Not sure about this either. `guard::unlinked` will be just plain wrong with a mapped `Shared<T>`. Seems like a potential footgun. `&T` can be mapped just fine and `Shared<T>` will deref into that anyway.

ticki:
@arthurprs 

> I spend a few hours on this with no success, most commits left the build broken and then a latter commit fixes it. So git bisect was useless. I tried getting each commit compiling but after a few hours I gave up, but that's probably the only way to find it without eyeballing or some magical debug powers.

It's my horrible habit: Just write and then fix when you're done.

@whataloadofwhat 

> Why? lazy_static blocks in its implementation. Using a locking algorithm to implement lock-free data structures seems counter-intuitive, even if it is unlikely to cause an issue. The manual implementation was optimistic and lock-free, which matches the ideology of the crate a lot better I think.

The previous implementation blocked as well. AFAICS there is no non-blocking way of doing this, and the blocking part is only during initialization and even then is rare.

> Not sure I agree with this. Owned<T> is currently implemented as a Box<T> but that's an implementation detail that I find worrying. Realistically, Owned<T> is akin to Box<ManualDrop<T>>. I'd much rather the code work with this rather than just assume that Vec<T> with len 0 and cap 1 is the same as just freeing the memory of Box<T>. It feels hacky and confusing in a codebase that is already difficult to follow (not anybody's fault, concurrency is hard).

That is, well, very weird thing. Like, of course it calls the inner destructor, and even the free is more sensical than the vector hack. If you want a changed destructor, you can just wrap your `T` in a newtype.

> I feel like this is a mistake. It's not unsafe or anything, I just think people will use it incorrectly.

I see literally no way of 'misusing' this?

> Not sure about this either. guard::unlinked will be just plain wrong with a mapped Shared<T>. Seems like a potential footgun. &T can be mapped just fine and Shared<T> will deref into that anyway.

This one is interesting. I don't think it is an issue in reality, as `unlinked` is already `unsafe`.

whataloadofwhat:
> The previous implementation blocked as well. AFAICS there is no non-blocking way of doing this, and the blocking part is only during initialization and even then is rare.

Maybe blocking was the wrong word? I'm sorry. The manual implementation is lock-free. `lazy_static` is not.

> I see literally no way of 'misusing' this?

`Pinned<T>` is lifetime free which makes it very tempting to just pass around and store, but doing so is going to inhibit garbage collection. Like I say, it's not unsafe, it's just ... encouraging bad patterns, I think. You shouldn't really be storing a `Guard` for a long period of time and `Pinned<T>` kind of encourages you to do so.

>This one is interesting. I don't think it is an issue in reality, as `unlinked` is already unsafe.

This is true, but it's expanding the scope in which it can be used incorrectly. I don't think it should be done idly, but I don't see much of a benefit to it either. The old invariant is "only used on `Shared<T>` where no new readers are going to be created". You've now created a new invariant, "ensure the `Shared<T>` has not been mapped".

> That is, well, very weird thing. Like, of course it calls the inner destructor, and even the free is more sensical than the vector hack. If you want a changed destructor, you can just wrap your T in a newtype.

I think I may have caused some confusion here so I'll drop this point. Sorry.

ticki:
>Pinned<T> is lifetime free which makes it very tempting to just pass around and store, but doing so is going to inhibit garbage collection. Like I say, it's not unsafe, it's just ... encouraging bad patterns, I think. You shouldn't really be storing a Guard for a long period of time and Pinned<T> kind of encourages you to do so.

That's a good point. It can be tempting to store, say, in a struct.

jeehoonkang:
I am closing this, as (1) we decided to split this PR into smaller ones, and (2) there is not so much traffic on this PR these days.

As a follow-up, I will make the smaller PRs of the changes made in this PR.  As noted above, the commits are quite mixed so that it will be hard to maintain the commits as-is.  I will make new commits based on the diffs.

arthurprs:
@jeehoonkang thanks for picking this up. I'll do my best to review in a timely manner.

PR #125, opened 2017-03-24
add is_empty() interface for SeqQueue

Xudong-Huang:
we need a way to quickly test if the queue is empty without taking the action to pull the queue

### DIFF
--- a/src/sync/seg_queue.rs
+++ b/src/sync/seg_queue.rs
@@ -89,6 +89,21 @@ impl<T> SegQueue<T> {
         }
     }
 
+    /// Judge if the queue is empty.
+    ///
+    /// Returns `true` if the queue is empty.
+    pub fn is_empty(&self) -> bool {
+        let guard = epoch::pin();
+        let head = self.head.load(Acquire, &guard).unwrap();
+        let tail = self.tail.load(Acquire, &guard).unwrap();
+        if head.as_raw() != tail.as_raw() {
+            return false;
+        }
+
+        let low = head.low.load(Relaxed);
+        low >= cmp::min(head.high.load(Relaxed), SEG_SIZE)
+    }
+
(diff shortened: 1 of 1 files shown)


Xudong-Huang:
* you are right. every container type may need that interface. But I only use seq_queue for it's high performance.  Sorry that I don't have time to touch other queues implementation right now.

* I have test the new interface with private project. it works as expected. I'm confidence that it works but I can't prove to say it's correct. I will update a new version with test code.

src/sync/seg_queue.rs
```diff
@@ -89,6 +89,21 @@ impl<T> SegQueue<T> {
...
+    /// Returns `true` if the queue is empty.
+    pub fn is_empty(&self) -> bool {
+        let guard = epoch::pin();
+        let head = self.head.load(Acquire, &guard).unwrap();
+        let tail = self.tail.load(Acquire, &guard).unwrap();
+        if head.as_raw() != tail.as_raw() {
+            return false;
+        }
+
+        let low = head.low.load(Relaxed);
```
jeehoonkang: Could you explain why this (and the one below) can be relaxed?

jeehoonkang:
Hi!  I just reconfigured Travis CI for the new repo (crossbeam-rs/crossbeam).  In order to test this PR, would you please send a fresh commit?  `git commit --amend` would work, I think.

src/sync/seg_queue.rs
Xudong-Huang: actually the logic just come from `try_pop()`, I didn't change the `Relaxed` logic here. But I think in line 97 we already has an **Acquire** sync which is enough.

arthurprs: Relaxed should be fine as synchronization doesn't really matter here.

PR #127, opened 2017-04-04
Move and remove files

jeehoonkang:
This is the first of the series of PRs I will make (cf. https://github.com/crossbeam-rs/crossbeam/pull/122#issuecomment-285875721).  I will start with a very mild one, which just move and remove files.


arthurprs:
For the sake of gathering opinions. Do we really want to remove the `mem::` module?

From my POV it's fine. If we ever add another memory reclamation algorithm it'll be a major version bump anyway and we can revisit this.

aturon:
@arthurprs I'm fine with that.

jeehoonkang:
I am fine with removing `mem::`, for the same reason with @arthurprs , but am a little bit concerned with the backward compatibility: the `epoch` API is broken by the renaming.  But I am still OK with that, as Crossbeam has not matured yet to commit to stability, in my humble opinion.  I would like to make other breaking changes before reaching version 0.3.0.

jeehoonkang:
Hi team,

I have pending PRs waiting for this.  Since this commit changes the directory structure, I think other PRs cannot be posted in parallel.  If this PR looks OK, could anyone please merge it?  Thanks.

PR #128, opened 2017-04-06
Clean up `atomic.rs` in `epoch`

jeehoonkang:
This PR is a subset of #122 , which is on `atomic.rs`.  Most notably:
- Rename `cas` into `compare_and_set`, and introduce `compare_and_swap`.
- Pass `&'a Guard` to `Shared::from_{raw,ref,owned}`, and mark them as safe.

Those are the changes in #122 I **didn't** make:
- Add `into_pinned`.  Currently I don't quite understand the use case of this.  I will make another PR when it's necessary.
- Remove `Owned`.  As @whataloadofwhat pointed out, `Owned<T>` is rather `Box<ManualDrop<T>>` (https://github.com/crossbeam-rs/crossbeam/pull/122#issuecomment-288436809).  Furthermore it would be nice to hide the implementation detail that `Owned` is implemented as `Box`, IMO.

I made the following breaking changes on `epoch`, but I think it's okay: we didn't commit to stability for `epoch` yet:
- Rename `store_and_ref` into `store_ref`.  This is consistent with the naming of `compare_and_set_ref`.
- Rename `cas*`.

### DIFF
--- a/src/epoch/atomic.rs
+++ b/src/epoch/atomic.rs
@@ -85,8 +85,8 @@ impl<T> Atomic<T> {
     /// # Panics
     ///
     /// Panics if `ord` is `Release` or `AcqRel`.
-    pub fn load<'a>(&self, ord: Ordering, _: &'a Guard) -> Option<Shared<'a, T>> {
-        unsafe { Shared::from_raw(self.ptr.load(ord)) }
+    pub fn load<'a>(&self, ord: Ordering, guard: &'a Guard) -> Option<Shared<'a, T>> {
+        Shared::from_raw(self.ptr.load(ord), guard)
     }
 
     /// Do an atomic store with the given memory ordering.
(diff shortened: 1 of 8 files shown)


whataloadofwhat:
Minor thing but I prefer `compare_and_store` because I think it matches better

`swap` -> `compare_and_swap`
`store` -> `compare_and_store`

Not a big deal though

jeehoonkang:
@whataloadofwhat I followed the nomenclature of the Rust standard library: https://doc.rust-lang.org/std/sync/atomic/struct.AtomicPtr.html

The only exception is `compare_and_set*`.  The purpose of these functions is getting back the `Owned` resources when the update fails.  So they does not satisfy the specification of compare-and-swap, which **should** return the old value.  I think the name "compare-and-set" matches well here: it usually returns a boolean (https://en.wikipedia.org/wiki/Compare-and-swap), but in addition we will return the given resource back when update fails.

whataloadofwhat:
Ah, I'm sorry, I mean that I feel that `compare_and_store` fits more than `compare_and_set`, because there is no `set` function on `Atomic`, but there is `store`.

jeehoonkang:
Closing it, as we are eventually planning to move to `crossbeam-epoch`.

PR #130, opened 2017-04-06
Revise CachePadded

jeehoonkang:
Here are the changes I made:

- Add documentation (originally written by @ticki in #122)

- Make the padding trick more readable.  Though I am not sure of its correctness (https://github.com/jeehoonkang/crossbeam/blob/cachepadded/src/cache_padded.rs#L15).  Would anyone please verify it?  Also I am wondering if there is a documentation on this trick.

- Split pad from the marker in `CachePadded`.

- Add an assertion in `assert_valid`.

- Remove validity check when dereferencing and dropping `CachePadded`.

Remaining questions:

- Is it better to describe the cache line size as `CONST * sizeof(usize)`?  Why not just write down the size in bytes?

- I think eventually `CACHE_LINE` should depend on the (micro-)architecture.  But I cannot find any existing C/C++ library on it.  Do anyone have an idea/pointer on this?

  Alternatively, we can get the cache line size from the environmental variable, with 64 bytes as default.

- How to support `T` whose size is bigger than `CACHE_LINE`?  It is needed when we want the cache line size to be 64 bytes, as `sizeof(Participant)` > 64.

### DIFF
--- a/src/cache_padded.rs
+++ b/src/cache_padded.rs
@@ -1,17 +1,18 @@
-use std::marker;
 use std::cell::UnsafeCell;
-use std::fmt;
-use std::mem;
-use std::ptr;
 use std::ops::{Deref, DerefMut};
+use std::{ptr, mem, fmt, marker};
 
-// For now, treat this as an arch-independent constant.
+/// An over-approximation of ((cache-line-size) / sizeof(usize))
+// FIXME: arch-dependent value for CACHE_LINE
 const CACHE_LINE: usize = 32;
 
-#[cfg_attr(feature = "nightly",
-           repr(simd))]
+// FIXME: this is a spooky hack that aligns `CachePadded` to the cache line.
+// It would be better to have a language feature on the alignment, e.g.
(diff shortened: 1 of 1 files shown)


src/cache_padded.rs
```diff
@@ -1,17 +1,18 @@
...
-#[cfg_attr(feature = "nightly",
-           repr(simd))]
+// FIXME: this is a spooky hack that aligns `CachePadded` to the cache line.
+// It would be better to have a language feature on the alignment, e.g.
+// `#[align(64)]`.  See https://github.com/rust-lang/rfcs/issues/325 for more
+// details.
+#[cfg_attr(feature = "nightly", repr(simd))]
 #[derive(Debug)]
-struct Padding(u64, u64, u64, u64);
+struct Padding([usize; CACHE_LINE]);
```
arthurprs: We need to make it u/i/XX in order to compile on nightly.

Vtec234:
A heads up, the alignment feature (rust-lang/rust#39999) just landed and should be available on nightly soon :)

jeehoonkang:
Closing it, as `crossbeam-utils` implements a new version of `CachePadded`.

PR #135, opened 2017-04-09
Add a slurp method to Treiber stack

sstewartgallus:
This is useful because it is wait-free instead of lock-free.

### DIFF
--- a/src/sync/treiber_stack.rs
+++ b/src/sync/treiber_stack.rs
@@ -58,7 +58,7 @@ impl<T> TreiberStack<T> {
         // new is exclusively owned and so can be written and read
         // with relaxed atomics
         let newhead = new.head.swap(None, Relaxed, &guard);
-        let head = self.head.swap_shared(newhead, AcqRel, &guard);
+2        let head = self.head.swap_shared(newhead, Relaxed, &guard);
         new.head.store_shared(head, Relaxed);
         return new;
     }
--- a/src/sync/treiber_stack.rs
+++ b/src/sync/treiber_stack.rs
@@ -49,6 +49,17 @@ impl<T> TreiberStack<T> {
         self.try_pop()
     }
 
+    /// Takes all the elements off the stack at once.
+    ///
+    /// Wait-free
+    pub fn slurp(&self) -> TreiberStack<T> {
+        let newstack = TreiberStack { head: Atomic::null() };
+        let guard = epoch::pin();
+        let head = self.head.swap(None, Acquire, &guard);
+        newstack.head.store_shared(head, Release);
+        return newstack;
+    }
+
     /// Attempt to pop the top element of the stack.
     ///
     /// Returns `None` if the stack is observed to be empty.
@@ -94,5 +105,9 @@ mod test {
         assert!(q.is_empty());
         q.push(25);
         assert!(!q.is_empty());
+
+        let r = q.slurp();
+        assert!(!r.is_empty());
+        assert!(q.is_empty());
     }
 }
--- a/src/sync/treiber_stack.rs
+++ b/src/sync/treiber_stack.rs
@@ -1,4 +1,4 @@
-use std::sync::atomic::Ordering::{AcqRel, Acquire, Release, Relaxed};
+use std::sync::atomic::Ordering::{Acquire, Release, Relaxed};
 use std::ptr;
 
 use mem::epoch::{self, Atomic, Owned};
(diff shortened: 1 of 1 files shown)


Vtec234:
This looks quite useful! What about implementing a more general `swap` instead? The `swap` would just swap the stack with the one given in its argument and `slurp` would be a `swap` with a new empty stack passed into it.

sstewartgallus:
@Vtec234 Good question. That seems reasonable. Annoyingly you can't go from an Atomic to an Owned but only a Shared so I'll have to think about how to implement it properly.

src/sync/treiber_stack.rs
```diff
@@ -49,6 +49,20 @@ impl<T> TreiberStack<T> {
...
+    /// Swaps a stack with a new one.  Can be used to take all the
+    /// elements off the stack at once.
+    ///
+    /// Wait-free
+    pub fn swap(&self, new: TreiberStack<T>) -> TreiberStack<T> {
+        let guard = epoch::pin();
+        // new is exclusively owned and so can be written and read
+        // with relaxed atomics
+        let newhead = new.head.swap(None, Relaxed, &guard);
+        let head = self.head.swap_shared(newhead, AcqRel, &guard);
```
Vtec234: It may be possible to make this `Relaxed` as well, since there's a data dependency flow between the three calls and `new` is only locally accessible. It's also fine for two adjacent calls to `swap` to get their `self.head.swap_shared`s reordered.

sstewartgallus: I think it needs to be at least Release so that pushes to new are guaranteed to happen before it is swapped in.

Vtec234: Do you mean the `new.head.swap`? I think it is guaranteed to happen on the current thread before `self.head.swap_shared` anyway due to the data dependency of the `swap_shared` on `newhead` and since `new` is only accessible on the current thread, that's enough.

src/sync/treiber_stack.rs
```diff
@@ -49,6 +49,20 @@ impl<T> TreiberStack<T> {
...
 
+    /// Swaps a stack with a new one.  Can be used to take all the
+    /// elements off the stack at once.
+    ///
+    /// Wait-free
+    pub fn swap(&self, new: TreiberStack<T>) -> TreiberStack<T> {
+        let guard = epoch::pin();
+        // new is exclusively owned and so can be written and read
+        // with relaxed atomics
+        let newhead = new.head.swap(None, Relaxed, &guard);
```
schets: I think this atomic swap is needlessly expensive. Since new is exclusively owned and new.head is written to later, could this just be rewritten to a load?

sstewartgallus: @schets It'll most likely be optimized away because new is passed by value but I'll change it.

schets: LLVM doesn't optimize it away in a minimal case, and I don't think LLVM will reliably trying to do such optimizations soon

src/sync/treiber_stack.rs
jeehoonkang: Even though there is a data dependency, C/C++ memory model allows an execution where the instructions are executed as if they are reordered.  Note that data dependency plays a key role for only consume reads.  (Though consume's semantics is not yet clearly sorted out..)

I think the ordering should be indeed `Release`, in order to "publish" the data in the `new` stack in the `self.head` location.

src/sync/treiber_stack.rs
```diff
@@ -49,6 +49,20 @@ impl<T> TreiberStack<T> {
         self.try_pop()
     }
 
+    /// Swaps a stack with a new one.  Can be used to take all the
+    /// elements off the stack at once.
+    ///
+    /// Wait-free
+    pub fn swap(&self, new: TreiberStack<T>) -> TreiberStack<T> {
```
jeehoonkang: How about `new: &mut TreiberStack<T>`?

sstewartgallus: I chose to do it similar to the swap method for AtomicBool and such. IMO a signature like `fn function(&x: A, &mut y: A);` is a bit ugly and confusing. I don't have a strong opinion either way though.

stjepang:
This is an old PR, but here's a quick suggestion for several methods we could add to `TreiberStack`:

```rust
fn clear(&self) -> TreiberStack<T>;
fn into_iter(self) -> IntoIter<T>;
fn push_batch<I: IntoIterator<Item = T>>(&self, iter: I);
```

Using these methods, one can move all elements from stack `a` to `b` with:

```rust
b.push_batch(a.clear())
```

(method names are bikesheddable)

stjepang:
This is a stale PR. Closing, but if you want to keep working on it, I'll reopen.

PR #145, opened 2017-07-06
Add `Default` implementation for `TreiberStack`

torkleyy:


src/sync/treiber_stack.rs
```diff
@@ -77,6 +77,12 @@ impl<T> TreiberStack<T> {
     }
 }
 
+impl<T> Default for TreiberStack<T> {
+    fn default() -> Self {
+        TreiberStack { head: Atomic::null() }
```
stjepang: It would be more idiomatic to leave `fn new` intact and implement `Default` by calling `TreiberStack::new()`.

torkleyy: Are you sure? I've seen it more often this way actually.

torkleyy: At least clippy recommends it the way it's implemented here: https://github.com/Manishearth/rust-clippy/blob/master/clippy_lints/src/new_without_default.rs#L34

stjepang: Hmm, interesting. Well, pretty much every `impl Default for ...` in rust-lang/rust relies on `Self::new()`. But I don't care too much either way. :)

torkleyy: Oh okay, didn't really contribute to the Rust compiler / libstd yet. If you still want me to change it, just leave another comment.

stjepang: Let's change it.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "crossbeam"
-version = "0.2.10"
+version = "0.2.11"
```
jeehoonkang: I think we can bump the version when releasing one.

torkleyy:
Okay, I think this is ready for merge now.

PR #146, opened 2017-07-06
Fix use-after-free when waking up a blocked thread

### LINKED ISSUE #107: MsQueue Segfault
I am playing around with an experimental new channel design (https://github.com/google/fchan-go), and have a draft implementation in rust (see rust_impl branch). I'm looking to evaluate its performance against the crossbeam blocking `MsQueue` data-structure.

Running the benchmark program (https://github.com/google/fchan-go/blob/rust_impl/fchan-rs/src/bin/bench.rs)  causes the `MsQueue` to segfault. I get this output:
```
fchan-n2	6820574.330741031 ops/s
fchan-n4	10748308.94253941 ops/s
fchan-n8	15876137.253150392 ops/s
fchan-n12	17872916.32695152 ops/s
fchan-n16	17865720.769311864 ops/s
fchan-n24	12006629.789970772 ops/s
fchan-n32	13828055.450307708 ops/s
MsQueue-n2	3623537.4289297326 ops/s
MsQueue-n4	7983964.701022765 ops/s
MsQueue-n8	9037691.379822284 ops/s
MsQueue-n12	8658681.51229544 ops/s
MsQueue-n16	8171181.839549608 ops/s
Segmentation fault (core dumped)
```
The benchmark doesn't include any unsafe code, so I think this means there is a bug somewhere. I get this error if I run the `MsQueue` benchmarks before `fchan` as well. 

This is on a 16-core/32-thread dual-socket x86 machine. I ran this on a windows machine within the WSL linux userland (https://msdn.microsoft.com/en-us/commandline/wsl/about). For what it's worth I am on a developer branch of this distribution, so there is a possibility of bugs there too. 

Let me know if I can provide any more info on the issue.

stjepang:
The problem is that we set `ready` to `true` and then again dereference `signal` to wake up the thread:

```rust
(*signal).ready.store(true, Relaxed);
// Just before we unpark the thread, someone else might do it and
// that thread might destroy the stack-allocated signal.
(*signal).thread.unpark();
```

We must acquire a `Thread` *before* setting `ready` to `true`, and only then unpark the thread without dereferencing `signal`.

cc @Vtec234

This *probably* fixes #107.

### DIFF
--- a/src/sync/ms_queue.rs
+++ b/src/sync/ms_queue.rs
@@ -1,5 +1,5 @@
-use std::sync::atomic::Ordering::{Acquire, Release, Relaxed, SeqCst};
-use std::sync::atomic::{AtomicBool, fence};
+use std::sync::atomic::Ordering::{Acquire, Release, Relaxed};
+use std::sync::atomic::AtomicBool;
 use std::{ptr, mem};
 use std::thread::{self, Thread};
 
@@ -176,7 +176,7 @@ impl<T> MsQueue<T> {
                             (*signal).data = Some(cache.into_data());
                             let thread = (*signal).thread.clone();
 
-                            (*signal).ready.store(true, Relaxed);
+                            (*signal).ready.store(true, Release);
                             thread.unpark();
                             guard.unlinked(head);
                             return;
--- a/src/sync/ms_queue.rs
+++ b/src/sync/ms_queue.rs
@@ -1,5 +1,5 @@
-use std::sync::atomic::Ordering::{Acquire, Release, Relaxed};
-use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering::{Acquire, Release, Relaxed, SeqCst};
+use std::sync::atomic::{AtomicBool, fence};
 use std::{ptr, mem};
 use std::thread::{self, Thread};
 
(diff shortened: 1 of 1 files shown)


src/sync/ms_queue.rs
```diff
@@ -174,8 +174,15 @@ impl<T> MsQueue<T> {
                         unsafe {
                             // signal the thread
                             (*signal).data = Some(cache.into_data());
+                            let thread = (*signal).thread.clone();
+
+                            // It's important that we store the data and get the thread that needs
+                            // to be woken up before setting `ready` to `true`.
+                            // FIXME(stjepang): Use `compiler_fence` when it gets stabilized.
+                            fence(SeqCst);
```
jeehoonkang: why is this fence needed?

stjepang: On another thought, I think using a fence here is fairly convoluted. I've pushed a new commit that simplifies this code a bit.

Basically, I just want `let thread = (*signal).thread.clone()` to happen before `(*signal).ready.store(true, Release)`.

The `Release` ordering will be enough (if we later on load `ready` using `Acquire`).

stjepang:
@jeehoonkang Can you take another look?

jeehoonkang:
Will do in 18 hours.  Sorry-  I am super busy these days...

PR #148, opened 2017-07-19
Steal-half in Chase-Lev

jeehoonkang:
this fixes a bug in calculating the size in the steal_half method in chase-lev deque.

obsletes https://github.com/crossbeam-rs/crossbeam/pull/72

@fitzgen I couldn't add you as a reviewer, but please review it :)

### DIFF
--- a/src/sync/chase_lev.rs
+++ b/src/sync/chase_lev.rs
@@ -285,7 +285,7 @@ impl<T> Deque<T> {
             return StealHalf::Empty;
         }
 
-        let half = size + 1 / 2;
+        let half = (size + 1) / 2;
 
         unsafe {
             let a = self.array.load(Acquire, &guard).unwrap();
--- a/src/sync/chase_lev.rs
+++ b/src/sync/chase_lev.rs
@@ -102,6 +102,19 @@ pub enum Steal<T> {
     Data(T),
 }
 
+/// When stealing half of the data, this is an enumeration of the possible
+/// outcomes.
+#[derive(PartialEq, Eq, Debug)]
+pub enum StealHalf<T> {
+    /// The deque was empty at the time of stealing.
+    Empty,
+    /// The stealer lost the race for stealing data, and a retry may return more
+    /// data.
+    Abort,
+    /// The stealer has successfully stolen some data.
+    Data(Vec<T>),
+}
+
 // An internal buffer used by the chase-lev deque. This structure is actually
 // implemented as a circular buffer, and is used as the intermediate storage of
 // the data in the deque.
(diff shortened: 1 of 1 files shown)


src/sync/chase_lev.rs
```diff
@@ -254,6 +273,40 @@ impl<T> Deque<T> {
...
+            for i in t..t + half {
+                data.push(a.get(i));
+            }
+
+            if self.top.compare_and_swap(t, t + half, SeqCst) == t {
+                StealHalf::Data(data)
+            } else {
+                while data.len() > 0 {
+                    mem::forget(data.pop());
+                }
```
stjepang: You can rewrite this loop in a different, perhaps more idiomatic way:
```rust
while let Some(x) = data.pop() {
    mem::forget(x);
}
```

src/sync/chase_lev.rs
```diff
@@ -254,6 +273,40 @@ impl<T> Deque<T> {
...
+
+        unsafe {
+            let a = self.array.load(Acquire, &guard).unwrap();
+
+            let mut data = Vec::with_capacity(half as usize);
+            for i in t..t + half {
+                data.push(a.get(i));
+            }
+
+            if self.top.compare_and_swap(t, t + half, SeqCst) == t {
```
stjepang: What happens if just before this CAS the worker thread pops most of the elements, so that `t + half` overshoots the bottom?

There is a special if-statement in `fn pop` where we deal with a special racy case when popping the last element. With the presence of the steal-half operation, now `steal_half` and `pop` can race in more complex ways.

The racy case in `fn pop` is currently handled by incrementing top rather than decrementing bottom. With steal-half we can't do a similar trick.

Perhaps this is the difficulty with implementing steal-half we've been missing all along? :)

jeehoonkang: @stjepang is right -- my bad.  This implementation of `steal_half()` is wrong, and I cannot imagine how to fix it.

The reason I thought it was correct is I only considered races among stealers.  As @stjepang said, `b` can be decremented beyond `t + half` before a `steal_half()` is successfully stealing from `t` to `t + half`.  That breaks deque's invariant.

Thank you pointing out that.  I think I should be more humble in front of concurrency...  :face_with_head_bandage:  I am closing this PR.

@fitzgen this PR is finally resolved, but negatively.  I will work more on designing a proper `steal_half()` in deque!

jeehoonkang:
I made another attempt for `steal_half()`: https://github.com/jeehoonkang/crossbeam-deque/tree/half-stealing

After [proving the correctness of Chase-Lev](https://github.com/crossbeam-rs/rfcs/pull/26), I realized that I can use a similar reasoning for the `steal_half()` method I implemented in the branch mentioned above.  I'll also prove the correctness of `steal_half()`, maybe in another RFC, and then send a PR.

jeehoonkang:
I made another attempt for `steal_half()`: https://github.com/jeehoonkang/crossbeam-deque/tree/half-stealing

After [proving the correctness of Chase-Lev](https://github.com/crossbeam-rs/rfcs/pull/26), I realized that I can use a similar reasoning for the `steal_half()` method I implemented in the branch mentioned above.  I'll also prove the correctness of `steal_half()`, maybe in another RFC, and then send a PR.

cc @fitzgen @stjepang

PR #149, opened 2017-07-19
Simplify the control flow of chase-lev deque

jeehoonkang:
The `try_pop` function's control flow is simplified.  I think this version is more readable than before.

A downside is that it diverges from the implementation in this [paper](http://www.di.ens.fr/~zappa/readings/ppopp13.pdf).  But I feel the benefit outweighs the cost in this case.

### DIFF
--- a/src/sync/chase_lev.rs
+++ b/src/sync/chase_lev.rs
@@ -210,16 +210,14 @@ impl<T> Deque<T> {
             Some(data)
         } else {
             // racy case. race against steals.
-            let data = a.get(t);
             let success = self.top.compare_and_swap(t, t + 1, SeqCst) == t;
 
             // set the queue to a canonically empty state.
             self.bottom.store(b, Relaxed);
 
             if success {
-                Some(data)
+                Some(a.get(t))
             } else {
-                mem::forget(data); // someone else stole this value
                 None
             }
         }
--- a/src/sync/chase_lev.rs
+++ b/src/sync/chase_lev.rs
@@ -184,46 +184,43 @@ impl<T> Deque<T> {
             b = self.bottom.load(Relaxed);
         }
         a.put(b, data);
-        fence(Release);
-        self.bottom.store(b + 1, Relaxed);
+        self.bottom.store(b + 1, Release);
     }
 
     unsafe fn try_pop(&self) -> Option<T> {
         let guard = epoch::pin();
 
-        let b = self.bottom.load(Relaxed) - 1;
+        let b = self.bottom.load(Relaxed);
         let a = self.array.load(Relaxed, &guard).unwrap();
-        self.bottom.store(b, Relaxed);
+        self.bottom.store(b - 1, Relaxed);
         fence(SeqCst); // the store to bottom must occur before loading top.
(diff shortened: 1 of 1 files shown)


src/sync/chase_lev.rs
```diff
@@ -184,46 +184,43 @@ impl<T> Deque<T> {
...
+        } else {
+            // racy case. race against steals.
+            let data = a.get(t);
+            let success = self.top.compare_and_swap(t, t + 1, SeqCst) == t;
+
+            // set the queue to a canonically empty state.
+            self.bottom.store(b, Relaxed);
+
+            if success {
+                Some(data)
```
stjepang: Can we actually read the data here and return `Some(a.get(t))`? That way we wouldn't need `mem::forget` later.

jeehoonkang: No we cannot.  After `self.top` is updated at line 213, it may be possible that other thread resizes the buffer, effectively deallocating `a`.  `a.get(t)` should be called before the update on `self.top`.

src/sync/chase_lev.rs
```diff
@@ -184,46 +184,43 @@ impl<T> Deque<T> {
             b = self.bottom.load(Relaxed);
         }
         a.put(b, data);
-        fence(Release);
-        self.bottom.store(b + 1, Relaxed);
+        self.bottom.store(b + 1, Release);
```
jeehoonkang: Yes, @alexcrichton is right -- I silently sneaked in something without explicitly saying it.  It probably should be done in a separate PR.

Having said that, I think release store also does its job: the release fence's sole purpose was releasing the view (think: information) gathered so far to the write of `self.bottom`.  The same job can be done via a release write to `self.bottom`.

In fact, I think several orderings in this implementation of Chase-Lev deque is stronger than what is necessary.  I will make another PR on the orderings.  In the mean time, I will roll back this change.

src/sync/chase_lev.rs
```diff
@@ -191,46 +191,44 @@ impl<T> Deque<T> {
...
-            self.bottom.store(b + 1, Relaxed);
-            None
         }
     }
 
     fn steal(&self) -> Steal<T> {
         let guard = epoch::pin();
 
-        let t = self.top.load(Acquire);
+        let t = self.top.load(Relaxed);
```
jeehoonkang: In fact, I also sneaked in this change in this PR.  This is correct because there is an SC fence right below.  FWIW, it is sound to transform acquire load into relaxed load and acquire fence, and two fences can be merged.

src/sync/chase_lev.rs
stjepang: But... only the worker can call `pop()` and resize the buffer, so that isn't possible.

src/sync/chase_lev.rs
stjepang: Fine with me. It's a bit odd that the paper didn't use `Relaxed` here - seems like an easy case.

src/sync/chase_lev.rs
jeehoonkang: I agree, thank you for the good catch!  I updated the PR.

PR #150, opened 2017-07-31
Bump to version 0.3

stjepang:
The motivation of this release is just to get a few bug fixes and features out so that users don't wait for them forever. 

Crossbeam 0.2.x has several long-standing bugs (e.g. leaks and crashes in `MsQueue`). These have been fixed in the master branch, so I suggest we cut a new release. There were also some compatibility breaking changes, so I'm bumping to version 0.3.

The next steps in development will be rework of EBR and splitting Crossbeam up into multiple crates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Version 0.3
 
 - Introduced `ScopedThreadBuilder` with the ability to name threads and set stack size
-- `Worker` methods don't require mutable access anymore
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+# Version 0.3
+
+- Introduced `ScopedThreadBuilder` with the ability to name threads and set stack size
+- `Worker` methods don't require mutable access anymore
```
jeehoonkang: I think it is worth mentioning it is deque's worker.

PR #151, opened 2017-08-03
Link directly to documentation.

frewsxcv:


jeehoonkang:
@frewsxcv Thank you for the PR!

I am merging it because it is just a patch to the documentation.  FYI, the README's link in this patch is canonical w.r.t. `https://docs.rs/`'s URI scheme.

PR #153, opened 2017-08-23
Fix memory leak in AtomicOption. Closes #142.

### LINKED ISSUE #142: `AtomicOption` should run `Box::drop` on drop.
Right now it's leaking memory and not running destructors. This is necessary:

```rust
impl<T> Drop for AtomicOption<T> {
    fn drop(&mut self) {
        let inner = self.inner.load(Ordering::SeqCst);
        if !inner.is_null() {
            unsafe { Box::from_raw(inner); }
        }
    }
}
```

Playground: https://is.gd/FUK0pr

spacejam:


src/sync/atomic_option.rs
```diff
@@ -9,6 +9,17 @@ pub struct AtomicOption<T> {
     inner: AtomicPtr<T>,
 }
 
+impl<T> Drop for AtomicOption<T> {
+    fn drop(&mut self) {
+        let inner = self.inner.load(Ordering::Relaxed);
+        if !inner.is_null() {
+            unsafe {
+                Box::from_raw(inner);
```
jeehoonkang: How about explicitly calling `drop` here?  `drop(Box::from_raw(inner))`

PR #157, opened 2017-10-04
Improve cache_line alignment on x86_64

Firstyear:
On x86_64 the cache_alignment is 64bytes. To accomodate dwcas, you need 128
bytes. The cache_line alignment therefore should be 18 * usize, which is 128
bytes wide. Other systems still default to 256 bytes (32 * 8).

### DIFF
--- a/src/cache_padded.rs
+++ b/src/cache_padded.rs
@@ -5,7 +5,15 @@ use std::mem;
 use std::ptr;
 use std::ops::{Deref, DerefMut};
 
-// For now, treat this as an arch-independent constant.
+// x86_64 has a 64byte cache alignment, so set this to 8 * sizeof(usize)
+// which happens to be 8, ie 64 bytes. However, the tests fail, so we set
+// this to 16 for 128 bytes to allow dwcas (which is likely the issue)
+#[cfg(target_arch = "x86_64")]
+const CACHE_LINE: usize = 16;
+// Other arches may not - for example, x86 may vary, ppc64 is 128, ppc32 is 32.
+// To be safe, we set a minimum of 32 * sizeof(usize) here, but specific
+// processor versions could be added later.
+#[cfg(not(target_arch = "x86_64"))]
 const CACHE_LINE: usize = 32;
 
 #[cfg_attr(feature = "nightly",


schets:
I think from the comment in the code you're confusing bits and bytes here. The x86 cache line is 64 bytes, not bits, and the alignment here is changed to 16 bytes. I'm not sure why it was ever 32 in the first place

stjepang:
This repository will soon be split into multiple crates (`crossbeam-epoch`, `crossbeam-utils`, `crossbeam-deque`, and so on).

The implementation of `CachePadded` now actually lives in [`crossbeam-utils`](https://github.com/crossbeam-rs/crossbeam-utils) (but will be re-exported in `crossbeam` as well). There's a [PR](https://github.com/crossbeam-rs/crossbeam-utils/pull/1) that revamps `CachePadded` - could you please take a look at it instead?

fitzgen:
> The implementation of CachePadded now actually lives in crossbeam-utils (but will be re-exported in crossbeam as well). There's a PR that revamps CachePadded - could you please take a look at it instead?

Idea: This seems like something useful outside of just `crossbeam`, and its annoying to track down cache line sizes for different architectures, therefore it seems like `CachePadded` could be useful as a standalone crate.

jeehoonkang:
@fitzgen There are two aspects to discuss on the nature of the `crossbeam-utils` crate: scope and name.

The idea of `crossbeam-utils` is providing utilities for concurrent programming in general, which isn't necessarily coupled with Crossbeam.  I prefer not to split `crossbeam-utils` into smaller ones because it will add managerial costs.

In my opinion, the primary reason we named it `crossbeam-utils` is the repository maintained in the Crossbeam organization.  While I prefer the current nomenclature, I understand that maybe someone wants to give a name that better reflects the nature of the crate, e.g. `concurrency-base`.

Firstyear:
Hi,

The alignment is actually multiple of width, that's why the numbers look funny. It was odd to me too until I read the code and did some testing. 

I don't think that splitting every crate into micro-crates is a good idea, because this makes it really hard for system packagers to use these things. Having one crate that handles this is a good thing because it's easier for a distro to distribute (and this is very imprortant to me as I want to use crossbeam in a major open source application). Were this crate split to many smaller crates it may become impractical for us to use this software :(  Please keep this as one crate focused on lockfree/parallel datastructures,

Thanks!

jeehoonkang:
For the same reason as in #168, I'd like to use `crossbeam-utils` instead.

Issue #13, opened 2015-10-18
Destructors


sorear:
I've been thinking about how I might use crossbeam::mem to implement a concurrent hash table.  The current examples of stacks and queues are actually somewhat special, because values can be moved into and out of them, but are never borrowed.  In contrast with a hash table I want to be able to read it without modifying, which requires borrows; but if a value is borrowed from a hash table and then deleted from it in another thread, the `drop()` call needs to be deferred until garbage collection.

Currently crossbeam only supports freeing memory after the grace period.  It seems like it could probably be modified to call destructors, but there might be some interesting caveats / possible unsafeties.  One obvious thing is that a drop call can free an arbitrary amount of memory, so intelligently batching them becomes more difficult (if an object is holding on to 20MiB of memory, we want to free it ASAP, not after 64 epochs).

schets:
I'm working on that [here](https://github.com/schets/crossbeam/tree/destructors)

>  In contrast with a hash table I want to be able to read it without modifying, which requires borrows; but if a value is borrowed from a hash table and then deleted from it in another thread, the drop() call needs to be deferred until garbage collection.

It's only safe to read (and borrow) from a datastructure like the hashtable while you know the data can't be freed - in this case, that means while a Guard is active. Likewise, it's only safe to call drop during garbage collection since it invalidates the region of memory.

> Currently crossbeam only supports freeing memory after the grace period. It seems like it could probably be modified to call destructors

This is pretty simple - 

> (if an object is holding on to 20MiB of memory, we want to free it ASAP, not after 64 epochs)

As it stands now, local garbage is freed whenever a pin call believes that it can and global garbage is freed whenever the epoch is advanced. Threads could be more forceful about advancing the epoch - GC thresholds could consider the garbage size so a large value doesn't sit waiting while no epochs advance and a flag could exist to pressure threads to advance the global epoch when putting lots of memory in global garbage.

However, any calls to free/drop must wait until it's safe (two epochs have advanced) and as of now epochs only advance when a call to pin decides to. In [this branch](https://github.com/schets/crossbeam/blob/gcless_pin/src/mem/epoch/mod.rs#L401), one of the functions I'm adding allows one to force attempts at GC and epoch advancement - if one knowingly unlinks a big object, this function could be called until enough epochs have advanced ( I'm working on the update that returns whether a GC actually ran or not).

[...]

PR #158, opened 2017-11-21
Call destructors when collecting garbage

bossmc:
Related to #13, this PR causes the garbage collector to call drop on the collected `T` nodes.

The previous implementation of the MS queue and the Trieber stack rely on being able to return the popped elements by simply `ptr::read` the `T` embedded in the entry in the structure.  This PR changes this to instead store the `T` in a new rather inelegant structure that holds a `T` but can have that `T` taken away from it (so that it won't drop it when it is dropped).

I'm not sure if the Chase-Lev deque relies on the same trick, since I've not grokked that component yet, similar for the segmented MS queue.  Perhaps someone can confirm/apply the same fix?

### DIFF
--- a/src/epoch/garbage.rs
+++ b/src/epoch/garbage.rs
@@ -40,7 +40,7 @@ impl Bag {
             })
         }
         unsafe fn free<T>(t: *mut u8) {
-            drop(Vec::from_raw_parts(t as *mut T, 0, 1));
+            drop(Box::from_raw(t as *mut T));
         }
     }
 
--- a/src/sync/mod.rs
+++ b/src/sync/mod.rs
@@ -12,3 +12,4 @@ mod treiber_stack;
 mod seg_queue;
 pub mod chase_lev;
 mod arc_cell;
+mod spot;
--- a/src/sync/ms_queue.rs
+++ b/src/sync/ms_queue.rs
@@ -1,9 +1,10 @@
 use std::sync::atomic::Ordering::{Acquire, Release, Relaxed};
 use std::sync::atomic::AtomicBool;
-use std::{ptr, mem};
+use std::mem;
 use std::thread::{self, Thread};
 
 use epoch::{self, Atomic, Owned, Shared};
+use sync::spot;
 use CachePadded;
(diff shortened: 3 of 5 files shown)


bossmc:
See the linked issue (https://github.com/bossmc/pinboard/issues/7) for my motivating issue.

bossmc:
The build failure being reported is a nightly-only failure due to some deprecated syntax already present in the code.

stjepang:
@bossmc Thanks for taking the time to submit this PR.

Note that dropping a value of type `T` like this is basically safe only when `T: 'static`, i.e. `T` does not hold any references to non-static values. This is important because the destructor of `T` might run arbitrarily late and may access the internals of `T`.

This is a known issue in Crossbeam and is fixed in [`crossbeam-epoch`](https://github.com/crossbeam-rs/crossbeam-epoch), which is going to be Crossbeam's new epoch-based GC. We'll release a 0.1.0 version very soon then replace the `epoch` module in `crossbeam` crate with `crossbeam-epoch`.

bossmc:
Good point... that's a shame, I changed a memory leak to a memory unsoundness.  Is there a write-up of the new epoch collection algorithm?  I'd be interested to see how you're bridging this gap.

stjepang:
I'd recommend reading the RFCs: https://github.com/crossbeam-rs/rfcs/tree/master/text

Currently, we have only one function for destroying garbage, called `defer`:
https://github.com/crossbeam-rs/crossbeam-epoch/blob/c22f09dbd930882e724c5033d9f19b74f9692a94/src/guard.rs#L155

This function defers execution of an arbitrary function until each currently pinned thread gets unpinned. You must ensure that executing the function is safe yourself, either by requiring `T: Send + 'static` or by doing other tricks.

bossmc:
Thanks, I've updated by library to use the new epoch collector.  Rather than trying to resolve the leaked references issue, I required `T` to be `'static` for now.

You can see my changes at https://github.com/bossmc/pinboard/pull/10 if you're interested :grinning:.

PR #162, opened 2017-12-31
bump rand to 0.4

ignatenkobrain:


jeehoonkang:
Thank you for this PR!  It seems there's no breaking changes.  In particular, It seems both 0.3 and 0.4 versions of `rand` support Rust 1.15.  I'm merging it.

PR #168, opened 2018-01-14
Let ZerosValid be a regular, non-auto trait, in nightly

### LINKED ISSUE #166: Is auto trait ZerosValid correctly defined?
Using crossbeam 0.3.2, features = nightly, it claims `Vec<i32>` is valid to initialize with all zeros

This should not compile, yet it compiles with both crossbeam 0.3.0 and 0.3.2

It's an auto trait with no opt outs?

```rust
extern crate crossbeam;
#[test]
fn test_cb() {
    use crossbeam::ZerosValid;
    fn is_zeros_valid<T: ZerosValid>() {}

    is_zeros_valid::<Vec<i32>>();
}
```

This nightly extension has never been correct.. maybe remove it?

bluss:
This bug is related to the crate feature "nightly" only.

It's an unintended bug, but ZerosValid was defined for every type and
did not serve its purpose. For example, `Vec<i32>` would implement
ZerosValid when it was an auto trait, and that was a bug -- all-zeros is
not a correct way to create a Vec; its pointer is always nonnull.

It does not seem to be possible to exclude all types by default and then
one by one whitelist types that are valid.

We could attempt a blacklist:

```rust
  impl<T: ?Sized> !ZerosValid for *const T { }
  impl<T: ?Sized> !ZerosValid for *mut T { }
  impl<'a, T: ?Sized> !ZerosValid for &'a T { }
  impl<'a, T: ?Sized> !ZerosValid for &'a mut T { }
```

And this excludes a lot of problematic types, like Vec, Box, Mutex etc!
But it is not a complete solution.

CachePadded::zeroed allows creating values from nothing as long as they
are ZerosValid. Let's say for example `struct UnforgeableToken(u64);`

With CachePadded::zeroed we could then create
`CachePadded<UnforgeableToken>` and use DerefMut to swap one of our valid
tokens with a forged, all-zeros one.

Fixes #166

### DIFF
--- a/src/cache_padded.rs
+++ b/src/cache_padded.rs
@@ -39,29 +39,11 @@ impl<T> fmt::Debug for CachePadded<T> {
 unsafe impl<T: Send> Send for CachePadded<T> {}
 unsafe impl<T: Sync> Sync for CachePadded<T> {}
 
-#[cfg(not(feature = "nightly"))]
-macro_rules! declare_zeros_valid {
-    () => {
-        /// Types for which mem::zeroed() is safe.
-        ///
-        /// If a type `T: ZerosValid`, then a sequence of zeros the size of `T` must be
-        /// a valid member of the type `T`.
-        pub unsafe trait ZerosValid {}
-    }
-}
-
-#[cfg(feature = "nightly")]
-macro_rules! declare_zeros_valid {
-    () => {
-        /// Types for which mem::zeroed() is safe.
-        ///
-        /// If a type `T: ZerosValid`, then a sequence of zeros the size of `T` must be
-        /// a valid member of the type `T`.
-        pub unsafe auto trait ZerosValid {}
-    }
-}
-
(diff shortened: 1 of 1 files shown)


bluss:
Hey, I think I found some reasons to why this just should not be an auto trait.

jeehoonkang:
Instead of updating CachePadded of `crossbeam`, how about using `crossbeam-utils` as far as we can?  This crate will eventually replace with `crossbeam-epoch` and `crossbeam-utils`, and we can start with a small portion of them.

jeehoonkang:
It turned out that we cannot use `crossbeam-utils` without switching to `crossbeam-epoch`: the current `epoch::Participant` is bigger than 64 bytes, so we cannot use `Cachepadded<Participant>`.  So I made the PR #169 that completely switches to `crossbeam-epoch`.

bluss:
Nice

PR #169, opened 2018-01-20
Use subcrates

### LINKED ISSUE #13: Destructors

### LINKED ISSUE #37: Resolve leak of participant data

### LINKED ISSUE #60: CachePadded size and alignment
`CachePadded` seems to massively overestimate the size of a cache line on 64-bit, it's also not clear why the size was chosen.

It seems that the previous estimate of 128-bytes was chosen to avoid interference between adjacent cache lines, but now is assuming a ridiculous 256 bytes due to it being `32 * sizeof(usize)` in the actual structure. Furthermore the reason for picking the sizes isn't stated anywhere, which makes understanding this code frustrating given that almost all CPUs have 64-byte cache lines.

Also the `Padding` struct only aligns to 32-bytes, (4 x 8 (u64) being 32), not the standard 64, which is also confusing.

### LINKED ISSUE #112: Segfault with large messages
Hello,

So, up front this is my first rust project and it's very likely I'm screwing something up. I'm getting segfaults from inside msqueue when I have a large message. Messages sized 0x1000 don't kill it, 0x10000 do.

Code, adapted from a larger project:
```rust
extern crate crossbeam;

use std::sync::Arc;
use std::thread;
use crossbeam::sync::MsQueue;

struct B {
    data : [u8; 0x10000]
}

fn new_b() -> B {
    B { data :  [0; 0x10000] }
}

impl B {
    pub fn rx(
        &mut self,
        rx : Arc<MsQueue<B>>)
    {
        loop {
            let _ = rx.pop();
        }
    }
}


fn main() {
    let b_queue : Arc<MsQueue<B>> = Arc::new(MsQueue::new());

    {
        let mut c = new_b();
        let b_queue = b_queue.clone();

        thread::spawn(move || {
            c.rx(b_queue);
        });
    }

    loop {

    }
}
```

```
$ rust-gdb -q ../target/debug/msq-crash
Reading symbols from ../target/debug/msq-crash...done.
(gdb) r
Starting program: /home/joe/rust/msqueue-crash/target/debug/msq-crash
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff6fff700 (LWP 38768)]

Thread 2 "msq-crash" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff6fff700 (LWP 38768)]
0x000055555556007f in crossbeam::sync::ms_queue::MsQueue<msq_crash::B>::pop<msq_crash::B> (self=<error reading variable: Cannot access memory at address 0x7ffff6dce4e0>)
    at /home/joe/.cargo/registry/src/github.com-1ecc6299db9ec823/crossbeam-0.2.10/src/sync/ms_queue.rs:241
241	    pub fn pop(&self) -> T {
(gdb) bt
#0  0x000055555556007f in crossbeam::sync::ms_queue::MsQueue<msq_crash::B>::pop<msq_crash::B> (self=<error reading variable: Cannot access memory at address 0x7ffff6dce4e0>)
    at /home/joe/.cargo/registry/src/github.com-1ecc6299db9ec823/crossbeam-0.2.10/src/sync/ms_queue.rs:241
#1  0x00005555555635d0 in msq_crash::B::rx (self=0x7ffff6e9e8e8, rx=Arc<crossbeam::sync::ms_queue::MsQueue<msq_crash::B>> = {...}) at /home/joe/rust/msqueue-crash/src/main.rs:21
#2  0x0000555555563d28 in msq_crash::main::{{closure}} () at /home/joe/rust/msqueue-crash/src/main.rs:35
#3  0x000055555556311b in std::panic::{{impl}}::call_once<(),closure> (self=<error reading variable: value of type `std::panic::AssertUnwindSafe<closure>' requires 65544 bytes, which is more than max-value-size>,
    _args=()) at /build/rustc-1.14.0+dfsg1/src/libstd/panic.rs:295
#4  0x000055555555c6c9 in std::panicking::try::do_call<std::panic::AssertUnwindSafe<closure>,()> (data=0x7ffff6f5ebc8 "\000") at /build/rustc-1.14.0+dfsg1/src/libstd/panicking.rs:356
#5  0x000055555557badb in panic_unwind::__rust_maybe_catch_panic () at /build/rustc-1.14.0+dfsg1/src/libpanic_unwind/lib.rs:97
#6  0x000055555555c2fa in std::panicking::try<(),std::panic::AssertUnwindSafe<closure>> (
    f=<error reading variable: value of type `std::panic::AssertUnwindSafe<closure>' requires 65544 bytes, which is more than max-value-size>) at /build/rustc-1.14.0+dfsg1/src/libstd/panicking.rs:332
#7  0x000055555555b5a5 in std::panic::catch_unwind<std::panic::AssertUnwindSafe<closure>,()> (
    f=<error reading variable: value of type `std::panic::AssertUnwindSafe<closure>' requires 65544 bytes, which is more than max-value-size>) at /build/rustc-1.14.0+dfsg1/src/libstd/panic.rs:351
#8  0x000055555556398d in std::thread::{{impl}}::spawn::{{closure}}<closure,()> () at /build/rustc-1.14.0+dfsg1/src/libstd/thread/mod.rs:287
#9  0x000055555555fbda in alloc::boxed::{{impl}}::call_box<(),closure> (self=0x7ffff7044c40, args=()) at /build/rustc-1.14.0+dfsg1/src/liballoc/boxed.rs:595
#10 0x00005555555721b5 in alloc::boxed::{{impl}}::call_once<(),()> () at /build/rustc-1.14.0+dfsg1/src/liballoc/boxed.rs:605
#11 std::sys_common::thread::start_thread () at /build/rustc-1.14.0+dfsg1/src/libstd/sys_common/thread.rs:21
#12 std::sys::imp::thread::{{impl}}::new::thread_start () at /build/rustc-1.14.0+dfsg1/src/libstd/sys/unix/thread.rs:84
#13 0x00007ffff79c0464 in start_thread (arg=0x7ffff6fff700) at pthread_create.c:333
#14 0x00007ffff74ec9df in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:105
```

This crash does not occur when I build with '--release'. I'm guessing this is related to [this?](https://github.com/rust-lang/rust/issues/16012)

Am happy to provide any other info I can. Tested on both macOS 10.12 and debian running linux 4.8.

### LINKED ISSUE #113: Low-level access to the epoch API
I need to make use of the 0x1 pointer, which is not possible under current high-level API. I imagine similar cases exists, and exposing low-level access to the internal `crossbeam::mem::epoch` methods would suffice.

### LINKED ISSUE #116: CAS is wrong (no compare-and-swap)
`Atomic::<T>::cas()` has signature

```rust
fn cas(&self, old: Option<Shared<T>>, new: Option<Owned<T>>, ord: Ordering) -> Result<(), Option<Owned<T>>>
```

But this isn't CAS in the classical sense. In fact, it makes certain things impossible to do. In particular, you don't get the value which didn't match. It acts more as "compare-and-**set**" rather than "compare-and-**swap**" (which is what the short form "CAS" tend to be used for).

I propose a **non-breaking** change (I do so because I think this problem is a major problem that needs to be addressed), which renames

- `cas` → `compare_and_set`
- `cas_and_ref` → `compare_and_set_ref`
-  `cas_shared` → `compare_and_set_shared`

And uses the old names to alias these, but mark them with `#[deprecated]` and changed their doc comments to "Deprecated. Do not use." (kind of like what I did [in chashmap](https://docs.rs/chashmap/2.2.0/chashmap/struct.CHashMap.html#method.filter)).

Futhermore, the crux is to introduce a new method, `compare_and_swap`, with signature

```rust
    pub fn compare_and_swap(&self, old: Option<Shared<T>>, new: Option<Shared<T>>, ord: Ordering)
               -> Result<(), Option<Shared<T>>>
```

As for the semantics, this method compares `self` against `old`, and if it matches, it sets `self` to `new` and returns `Ok(())`. If the two don't match, `Err(val)` is returned, where `val` is the value `self` instead holds.

The change is pretty basic, but is it something of interest? 

ping @aturon @alexcrichton /

### LINKED ISSUE #137: `Shared<T>::deref()` should be marked as unsafe
Related discussion: https://github.com/stjepang/coco/pull/2

The dereference of a `Shared<T>`, may race with the writer of the shared data of type `T`.  For example, suppose a `Box<T>` is written to `Atomic<T>` with `store_ref(Relaxed)`, and later another thread read the `Atomic<T>` with `load(Relaxed)`.  Then the `Box<T>` itself may not have been synchronized, and the dereference of the loaded `Shared<T>` may invoke undefined behavior.

### LINKED ISSUE #159: Store a `Scope` in a struct
Hello,
Is it possible to store a `Scope` in a struct ? The idea is to tie the lifetime of a thread to a struct instead of a function.
Thanks !

### LINKED ISSUE #165: Please set the maintenance status to deprecated
See stjepang/coco#21.

### LINKED ISSUE #166: Is auto trait ZerosValid correctly defined?

jeehoonkang:
This is a long-waited architectural change that makes `crossbeam` use subcrates we've been developing so far: `crossbeam-epoch`, `crossbeam-utils`, and `crossbeam-deque`.

It closes #168, closes #157, closes #70, closes #166, closes #165, closes #159, closes #137, closes #116, closes #113, closes #112, closes #60, closes #37, and closes #13.  If any of them seems still relevant to the new version, please leave a message here.

It contains a huge API breaking change, so I bumped the version to `0.4.0`.  Before publishing a new version, I think we need to discuss how to organize this crate and how to document things, esp. those in the subcrates.  Maybe we need an RFC for making a consensus.


Cargo.toml
```diff
@@ -10,7 +10,14 @@ license = "Apache-2.0/MIT"
 categories = ["concurrency", "data-structures"]
 
 [features]
-nightly = []
+default = ["use_std"]
+nightly = ["crossbeam-epoch/nightly"]
```
Vtec234: Could also include `crossbeam-utils/nightly` and `crossbeam-channel/nightly`.

Cargo.toml
```diff
@@ -10,7 +10,14 @@ license = "Apache-2.0/MIT"
...
+use_std = ["crossbeam-epoch/use_std"]
```
Vtec234: Same here, `crossbeam-utils/use_std`.

Cargo.toml
```diff
@@ -10,7 +10,14 @@ license = "Apache-2.0/MIT"
...
+
+[dependencies]
+crossbeam-epoch = "0.2"
+crossbeam-utils = "0.2"
+crossbeam-deque = "0.1"
```
Vtec234: Do we not want to expose `channel` for now?

jeehoonkang: Eventually, `crossbeam-channel` should replace the two queue implementations in `sync`.  I don't know what's the best way to do it.  I omitted this migration in this PR because it's @stjepang's call to make.

jeehoonkang:
@Vtec234 I adjusted `[features]` in `Cargo.toml` as you said.  Thanks!

jeehoonkang:
@stjepang hmm, do you want to reexport the macros defined in `channel`?  I think it's not stabilized yet: https://github.com/rust-lang/rust/issues/29638


When I just copy-and-pasted your code snippet in `sync/mod.rs`, the compiler says:

```rust
error[E0468]: an `extern crate` loading macros must be at the crate root
  --> src/sync/mod.rs:11:1
   |
11 | pub extern crate crossbeam_channel as channel;
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```


When I copy-and-pasted your code snippet in `lib.rs`, the compiler says:

```rust
warning: unused `#[macro_use]` import
  --> src/lib.rs:30:1
   |
30 | #[macro_use]
   | ^^^^^^^^^^^^
   |
   = note: #[warn(unused_imports)] on by default
```

stjepang:
Does [this](https://github.com/rust-lang/rust/issues/29638#issuecomment-298517765) solution work?

jeehoonkang:
@stjepang It works, but as a side effect, all the members of `crossbeam_channel` is now visible in `crossbeam`, e.g. we have `crossbeam::bounded()`.  I don't think there's a better workaround..

jeehoonkang:
@stjepang The side-effect of visibility (e.g. `crossbeam::bounded()`) seems too big to my eyes.  How about merging this PR without incorporating crossbeam-channel, and then later making another PR for crossbeam-channel?

stjepang:
Yes, I'm okay with leaving `crossbeam-channel` out.

PR #172, opened 2018-02-10
Move /src/bin to /benches

### LINKED ISSUE #111: binaries are too generic
When you do `cargo install crossbeam`, you end up having `bench` and `stress-msq` in your path. This is not very good idea.

Given that it's some bench things, probably it should be only built when running benches and never installed?

jeehoonkang:
I moved `/src/bin` to `/benches`, and adjust it for being used as `cargo +nightly bench`.

Closes #111


stjepang:
There is no more `src/bin` so let's close this.

PR #174, opened 2018-04-18
impl more std traits

HyeonuPark:
Mainly for `Default`s.

Just wonder why only `TreiberStack` implements `Default` :)

### DIFF
--- a/src/sync/arc_cell.rs
+++ b/src/sync/arc_cell.rs
@@ -52,6 +52,24 @@ impl<T> ArcCell<T> {
     }
 }
 
+impl<T: Default> Default for ArcCell<T> {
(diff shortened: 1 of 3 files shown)


jeehoonkang:
It's been open for three months, and I think it's worth merging.  I'm merging it!

jeehoonkang:
According to Travis, this PR breaks the `master` branch: https://travis-ci.org/crossbeam-rs/crossbeam/builds/406810885?utm_source=github_status&utm_medium=notification

But this is merely due to the fact that `crossbeam` currently depends on an old version of `crossbeam-epoch`.  After merging #182, the test failure will be gone.

Issue #179, opened 2018-07-05
crossbeam 0.4 not found


gnzlbg:
The readme says:

> [dependencies]
> crossbeam = "0.4.0"

cargo answers:

> error: no matching version `^0.4.0` found for package `crossbeam`

Vtec234:
Sorry, it's an oversight in the documentation. The latest version is `0.3.2`.

jeehoonkang:
Probably it's good to publish `0.4.0` soon-ish.  Can we do that now?  What're the release blockers?

stjepang:
Just a few simple requests before publishing `0.4.0`.

* Let's simplify the structure of `crossbeam-utils` a bit: https://github.com/crossbeam-rs/crossbeam-utils/pull/22
* `crossbeam-channel` is now basically stable and we can re-export it in `crossbeam`.
* Regarding re-exports, perhaps we should consider using `pub use` instead of `pub extern crate`, just like in Tokio (see [here](https://github.com/tokio-rs/tokio/blob/master/src/lib.rs)). This might lead to cleaner documentation.

PR #182, opened 2018-07-21
Reorganize modules

jeehoonkang:
This PR reorganizes submodules (as suggested in https://github.com/crossbeam-rs/crossbeam/issues/179#issuecomment-405053632).

r? @stjepang


src/lib.rs
```diff
@@ -25,11 +25,32 @@
...
+}
 
 pub mod sync;
-pub use utils::cache_padded::CachePadded;
-pub use utils::scoped::{scope, Scope, ScopedJoinHandle, ScopedThreadBuilder};
+
+// FIXME(jeehoonkang): this is necessary for re-exporting the `select!` macro, but it also
+// introduces a lot of symbols in the global namespace.  We need to find a more precise way to
+// re-export `select!`.
+pub use crossbeam_channel::*;
```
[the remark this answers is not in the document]
jeehoonkang: It seems what you wrote doesn't work: `select!` macro is not re-exported.  Instead, I did:

```rust
extern crate crossbeam_channel;

pub mod channel {
    pub use crossbeam_channel::*;
}

#[doc(hidden)]
pub use crossbeam_channel::*;
```

src/sync/treiber_stack.rs
```diff
@@ -46,7 +46,8 @@ impl<T> TreiberStack<T> {
     /// **Deprecated method**, use try_pop
     ///
     /// Returns `None` if the stack is observed to be empty.
-    #[cfg_attr(any(feature="beta", feature="nightly"), deprecated(note="The pop method has been renamed to try_pop for consistency with other collections."))]
+    #[cfg_attr(any(feature = "beta", feature = "nightly"),
```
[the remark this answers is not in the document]
jeehoonkang: I just removed the `pop()` method.  Anyway, it's marked as deprecated, and let's remove it in the 0.4.0 release...

jeehoonkang:
@stjepang I reorganized things as you suggested.  Thanks!

I think it's better to merge this PR after [releasing a new version of `crossbeam-utils`](https://github.com/crossbeam-rs/crossbeam-utils/pull/23), as after the `crossbeam-utils` 0.5.0 release this crate should be updated accordingly.

jeehoonkang:
.. and now this PR bumps the version of `crossbeam-utils` it depends to 0.5.0!

src/lib.rs
```diff
@@ -25,11 +25,101 @@
...
+mod ms_queue;
+mod seg_queue;
+mod treiber_stack;
+
+
+/// Additional utilities for atomics.
+pub mod atomic {
+    pub use crossbeam_utils::AtomicConsume;
+    pub use arc_cell::ArcCell;
+    // This module will grow in the future.
```
stjepang: Probably don't need this comment. :)

src/lib.rs
```diff
@@ -25,11 +25,101 @@
...
+///
+/// See [the `crossbeam-utils` crate](https://github.com/crossbeam-rs/crossbeam-utils) for more
+/// information.
+pub mod thread {
+    pub use crossbeam_utils::thread::*;
+}
+
+// Export `scope` and `CachePadded` in the crate root in order not to break established patterns.
+pub use thread::scope;
+pub use crossbeam_utils::CachePadded;
```
stjepang: `crossbeam::CachePadded` is not that common (in contrast to `crossbeam::scope`, which everybody knows about) so in this case I'd prefer to simply leave `CachePadded` inside `utils` module only. What do you think?

src/lib.rs
stjepang: Probably don't need this comment. :)

src/lib.rs
jeehoonkang: > Since `crossbeam-utils` is our "catch-all" crate for random things that don't deserve their own crates (or at least not yet), why not scatter its contents around more logically?

I agree with your statement above, so I'm hesitant to introduce `utils` submodule in `crossbeam`.  Maybe someday we will introduce a submodule to which `CachePadded` logically belongs, but for now, I cannot imagine what it will be...  That's why I'd like to put `CachePadded` at the root for the moment.

I think crate organization is an important matter, so even though this PR is approved, I'll wait for us to agree on this matter.

stjepang: Another contender for `utils` I've often wanted is [`Backoff`](https://github.com/crossbeam-rs/crossbeam-channel/blob/ce0b2db0ef5a3609cb6e628e3aff649e8eaf4860/src/internal/utils.rs#L13) struct (also really difficult to put into a more specific category). It automatically yields the processor/thread in CAS loop.

stjepang:
This is more of a hypothetical question, but here it goes: While we currently don't have any basic synchronization primitives, we'll probably implement them in the future. I wonder what module they should belong to.

Some examples of such primitives:
* [`Exchanger`](https://docs.oracle.com/javase/8/docs/api/index.html?java/util/concurrent/package-summary.html)
* [`StampedLock`](https://docs.oracle.com/javase/8/docs/api/index.html?java/util/concurrent/package-summary.html)
* [`Semaphore`](https://docs.oracle.com/javase/8/docs/api/index.html?java/util/concurrent/package-summary.html)
* [`Simple latches`](https://github.com/rayon-rs/rayon/blob/05c8177b05ee78ca1419420ff3a47fd5ae27388a/rayon-core/src/latch.rs#L46)
* `Handoff<T>`, which is basically just an `Option<T>` and `AtomicBool`, which can be used for passing a single value from one thread to another. Very useful ([example](https://github.com/crossbeam-rs/crossbeam-channel/blob/591d3165f8e8892023500cd31195be7949f2235d/src/flavors/zero.rs#L23)).
* Scalable reader-writer lock, where readers don't contend by updating the same atomic value on every lock (bad for cache efficiency).

I'd probably put those in `sync` module (we don't need it right now), but I wonder what you think.

jeehoonkang:
cc @Amanieu 

It seems some of the examples @stjepang mentioned are already in [`parking_lot`](https://github.com/Amanieu/parking_lot).  Personally, I think those primitives are fit into `parking_lot` rather than `crossbeam`.  Though maybe we can re-export `parking_lot` as a submodule of `crossbeam`.

Amanieu:
`parking_lot` is a mechanism to support blocking threads while waiting on a particular memory address. I can definitely see a place for `Semaphore` in there. Not sure about the others, they seem more obscure.

Regarding `StampedLock`, have you looked at my [`seqlock`](https://crates.io/crates/seqlock) crate? It seems to be similar.

stjepang:
On second thought, you might be right -- such utilities could be out of scope of crossbeam. And, we already have crates like `seqlock` and `semaphore`.

Anyways... As far as I'm concerned, after adding `utils` module this PR will be ready.

src/lib.rs
```diff
@@ -44,17 +44,18 @@ pub mod atomic {
...
 
-/// Scoped thread.
+/// Utilities for concurrent programming.
 ///
 /// See [the `crossbeam-utils` crate](https://github.com/crossbeam-rs/crossbeam-utils) for more
 /// information.
-pub mod thread {
-    pub use crossbeam_utils::thread::*;
+pub mod utils {
+    pub use crossbeam_utils::*;
```
stjepang: Maybe put here only `CachePadded` since `AtomicConsume` is inside `atomic` and `thread` module is already exported?

jeehoonkang:
Thanks for the review.  I'm merging it!

PR #184, opened 2018-07-25
Use ManuallyDrop in queues

stjepang:
This will hopefully fix https://github.com/crossbeam-rs/crossbeam-epoch/issues/82

### DIFF
--- a/src/ms_queue.rs
+++ b/src/ms_queue.rs
@@ -1,6 +1,7 @@
-use std::sync::atomic::Ordering::{Acquire, Relaxed, Release};
+use std::mem::{self, ManuallyDrop};
+use std::ptr;
 use std::sync::atomic::AtomicBool;
-use std::{mem, ptr};
+use std::sync::atomic::Ordering::{Acquire, Relaxed, Release};
 use std::thread::{self, Thread};
 
 use epoch::{self, Atomic, Owned, Shared};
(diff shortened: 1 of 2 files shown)


c0gent:
Will test.

stjepang:
@c0gent Published v0.4.1 with this fix.

Thomasdezeeuw:
Maybe I'm missing something, but shouldn't this now manually drop the values? Doesn't the current code leak values?

stjepang:
@Thomasdezeeuw  There are no leaks.

Queue nodes (more precisely, `Segment` and `Payload`) are just temporary storage for values between the moment they are pushed and the moment they are popped. Note that we only destroy empty nodes (not containing any values). Therefore, dropping a node mustn't run any value destructors (we enforce this by wrapping values into `ManuallyDrop`). In other words, just because we have a `ManuallyDrop<T>`, that doesn't necessarily mean a value is currently written there.

Thomasdezeeuw:
@stjepang Ok thanks for the reply.

Shnatsel:
Thanks for the quick resolution and the yank!

This seems to be some kind of a memory safety issue, perhaps a use-after-free, and those are usually exploitable. Could you file this issue in the [Rust security advisory database](https://github.com/RustSec/advisory-db) so that people who still have 0.4.0 in Cargo.lock could find out that they're using a vulnerable version and upgrade?

stjepang:
@Shnatsel Reported: https://github.com/RustSec/advisory-db/pull/75

PR #185, opened 2018-07-28
Drop guard before sleeping

### LINKED ISSUE #103: MsQueue::pop pins epoch until return
The `MsQueue::pop` currently [starts with](https://github.com/aturon/crossbeam/blob/bf1fb5627d67582a53ef712e8cb07fb9ebf878fa/src/sync/ms_queue.rs#L242) a call to `epoch::pin`. This I believe will prevent collection until the method returns, which could potentially take quite a long time (due to the blocking behavior). 

I *think* it may be possible to just drop the epoch [in the blocking case](https://github.com/aturon/crossbeam/blob/bf1fb5627d67582a53ef712e8cb07fb9ebf878fa/src/sync/ms_queue.rs#L298-L301), but that may be naive, so I'd want to think through this some more.

stjepang:
Fixes #103

### DIFF
--- a/src/ms_queue.rs
+++ b/src/ms_queue.rs
@@ -316,7 +316,6 @@ impl<T> MsQueue<T> {
                 Ok(()) => break,
                 Err(n) => {
                     node = n;
-                    continue;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


jeehoonkang:
Thanks!  Also thanks @alexcrichton and @martinhath for discussing this matter!

PR #186, opened 2018-08-30
add AtomicOption

programmerjake:
Add `AtomicOption`, an atomic version of `Option<ReferenceLike<T>>` where `ReferenceLike<T>` is `Box<T>`, `Arc<T>`, `&T`, and `&mut T` for now -- basically anything that has a lay-out like `NonNull<T>`.

I didn't implement `AtomicOption<NonNull<T>>` for now as `NonNull<T>` is not `Send` so `AtomicOption` is mostly useless for that case.

### DIFF
--- a/src/atomic_option.rs
+++ b/src/atomic_option.rs
@@ -1,0 +1,144 @@
+use std::marker::PhantomData;
+use std::mem::transmute;
+use std::sync::{
+    atomic::{AtomicPtr, Ordering},
+    Arc,
+};
+
+/// trait for reference-like objects where `Option<Self>` can be safely transmuted
+/// to a `*mut Self::Target`, sent to another thread (unless `Self` is not `Send`),
+/// and transmuted back
+pub unsafe trait ReferenceLike: Sized {
+    /// the type that `Self` is a reference to; usually `Deref::Target`
+    type Target;
+    /// transmute `Option<Self>` into a `*mut Self::Target` without panicking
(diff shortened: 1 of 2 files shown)


stjepang:
Hi @programmerjake, and thanks for the pull request! :)

We've recently added [`AtomicCell`](https://github.com/crossbeam-rs/crossbeam-utils/blob/61df3e4261336dc927c9fe9d922b0d42fc4d641c/src/atomic/atomic_cell.rs) to `crossbeam-utils`, which is very similar to `AtomicOption` and more general. For example, the equivalent of `AtomicOption<Box<T>>` would be `AtomicCell<Option<Box<T>>>`.

See https://github.com/crossbeam-rs/crossbeam-utils/pull/13 and https://github.com/crossbeam-rs/crossbeam-utils/pull/39 for more details.

Instead of adding `AtomicOption`, we should just publish a new version of `crossbeam-utils` and then reexport `AtomicCell` in `crossbeam`.

What do you think - does `AtomicCell` have everything you need from it?

programmerjake:
`AtomicCell` seems satisfactory. I do think it would be nice to add methods to `AtomicCell` that allow use of more relaxed atomics:
- `Ordering::AcqRel` should be fine from safe code as that's what `Mutex` provides.
- weaker orderings would be nice to have but they would be `unsafe` unless you can prove that the data doesn't contain any references or pointers, maybe using an auto trait to require that.

PR #187, opened 2018-09-08
update crossbeam-deque to 0.6

ignatenkobrain:


ignatenkobrain:
Oh, since crossbeam re-exports all those.. We should bump semver. Should I do so or you prefer to do something else?

PR #192, opened 2018-10-24
Fix UAF in TreiberStack

### LINKED ISSUE #191: TreiberStack seems to double-free the data
I'm currently using crossbeam in a project involving complex cross-thread communications, but I'm currently hitting a problem where the application consistently raises a SIGABRT for a double-free (FYI this is on Android ARMv7 using Rust nightly-2018-09-05 and crossbeam 0.4.1). I can't really share my code, but I believe I managed to create a minimal reproduction case by sticking this code in the tests of TreiberStack:

```rust
#[test]
fn double_free() {
    use std::{
        collections::HashSet,
        sync::{
            atomic::{AtomicUsize, Ordering},
            Arc,
            Mutex,
        },
    };

    static ID: AtomicUsize = AtomicUsize::new(1);

    struct WithId(usize, Arc<Mutex<HashSet<usize>>>);

    impl WithId {
        fn new(ctx: &Arc<Mutex<HashSet<usize>>>) -> Self {
            WithId(ID.fetch_add(1, Ordering::Relaxed), ctx.clone())
        }
    }

    impl Drop for WithId {
        fn drop(&mut self) {
            let mut dropped = self.1.lock().unwrap();
            assert!(dropped.insert(self.0));
        }
    }

    let q: TreiberStack<WithId> = TreiberStack::new();
    let ctx = Arc::new(Mutex::new(HashSet::new()));

    q.push(WithId::new(&ctx));
    q.push(WithId::new(&ctx));
    q.push(WithId::new(&ctx));

    drop(q.try_pop());
    drop(q.try_pop());
    drop(q);
}
```

Interestingly the assertion error is only triggered when the tests are running on multiple threads, with `--test-threads 1` everything runs smoothly.

Vtec234:
Fixes #191. This is also a security issue that could lead to remote code execution via a use-after-free, but crossbeam is still in development, so I wouldn't say this is very serious. Thoughts?

### DIFF
--- a/src/treiber_stack.rs
+++ b/src/treiber_stack.rs
@@ -1,5 +1,5 @@
 use std::sync::atomic::Ordering::{Acquire, Relaxed, Release};
-use std::ptr;
+use std::{ptr, mem};
 
 use epoch::{self, Atomic, Owned};
 
@@ -15,6 +15,13 @@ pub struct TreiberStack<T> {
 struct Node<T> {
     data: T,
     next: Atomic<Node<T>>,
+}
+
+impl<T> Node<T> {
+    /// Deallocates the memory for this node without executing T's destructor.
+    fn finalize(n: Node<T>) {
+        mem::forget(n.data);
+    }
 }
 
 impl<T> TreiberStack<T> {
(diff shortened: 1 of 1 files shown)


Vtec234:
> Oh, this is a classic bug that has appeared in a bunch of our data structures. :)
Welp, that's not great :|

> We would typically fix it by changing value: T to value: ManuallyDrop<T>
Yeah that's much better, initially I coded up the first thing that came into my mind and would work.

PR #195, opened 2018-11-05
Merge remaining subcrates

stjepang:
Continuation of #193. This PR merges:

* `crossbeam-channel`
* `crossbeam-epoch`
* `crossbeam-deque`
* `crossbeam-skiplist`

Closes https://github.com/crossbeam-rs/crossbeam-utils/issues/53


Vtec234:
Is Travis able to build multiple subprojects within a single repo separately? Waiting for all crates to build and scrolling through pages of output seems non-ideal.

stjepang:
@Vtec234 

> Is Travis able to build multiple subprojects within a single repo separately? Waiting for all crates to build and scrolling through pages of output seems non-ideal.

The total build time is not *too* bad at the moment - it takes 10 minutes. For comparison, Rayon takes 15 minutes and Tokio needs 20 minutes.

But yes, we could make builds smarter. I believe it should be relatively simple for our CI scripts to figure out whether the current code is a pull request. If so, then we could take a diff from the master branch, check which crates were changed, and tests only those affected.

jeehoonkang:
But if crossbeam-utils or crossbeam-epoch is touched, probably we wanna test all the crates, right?

.travis.yml
```diff
@@ -1,5 +1,12 @@
 language: rust
 
+rust:
+  - 1.25.0
+  - 1.26.0
```
jeehoonkang: How about testing 1.25 only?

stjepang:
@jeehoonkang 

> But if crossbeam-utils or crossbeam-epoch is touched, probably we wanna test all the crates, right?

That's a good question. I'm not sure. Maybe. :)

> How about testing 1.25 only?

I've just changed the Travis configuration matrix so that each crate is tested on its minimum supported Rust version and on nightly. Each (Rust version, crate) pair gets a separate job on Travis.

Cargo.toml
```diff
@@ -34,6 +34,11 @@ rand = "0.5"
 [workspace]
 members = [
   ".",
+  "crossbeam-channel",
+  "crossbeam-channel/benchmarks",
+  "crossbeam-deque",
+  "crossbeam-epoch",
+  "crossbeam-skiplist",
   "crossbeam-utils",
 ]
```
Vtec234: Can the `TODO`s in this file be resolved now? Seems all subcrates are here now.

stjepang:
By not installing unnecessary `.deb` packages, I brought the total build time down to 8 minutes.

PR #209, opened 2018-11-05
[WIP] Support DSTs

jeehoonkang:
NOTE: this is a copy of https://github.com/crossbeam-rs/crossbeam-epoch/pull/87

Currently, crossbeam-epoch doesn't support dynamically-sized types (DST): https://github.com/crossbeam-rs/rfcs/blob/master/text/2017-05-02-atomic-api.md#fat-pointers-and-dsts

This PR adds the support for DST by generalizing `Box<T>`, which was used as the underlying storage type for atomic pointers, to the `Storage<T>` trait.  For more details, see the comment on the `Storage` trait.  As an example, I added support for [dynamically-sized arrays](https://github.com/crossbeam-rs/crossbeam-epoch/pull/87/files#diff-79833cfe519a852b29369ee44aa044acR10).  Other kinds of DSTs can be supported by implementing `Storage<T>`, even in the client code.

This PR was motivated by the fact that there's an unnecessary indirection in `crossbeam-deque` when accessing its underlying buffer.  As a proof of concept, I removed that indirection on top of PR in my local development: [`concurrent-circbuf`](https://github.com/jeehoonkang/concurrent-circbuf/blob/master/src/sp/mod.rs#L77).

This is a **breaking change**, but I tried to keep it as small as possible.  As far as I understand, the only breaking change is renaming `Atomic::from(*const T)` into `Atomic::from_raw(*const T)` and `Shared::from(*const T)` into `Shared::from_raw(*const T)`.  (Question: I ended up changing the underlying types.  Is it a breaking change?)

This is a WIP.  Probably we want to discuss this PR's design and even write up an RFC.  I just share it for starting a discussion.  Thanks!


crossbeam-epoch/src/storage/array_storage.rs
```diff
@@ -0,0 +1,247 @@
...
+    _marker: PhantomData<T>,
+}
+
+fn div_ceil(a: usize, b: usize) -> usize {
+    (a + b - 1) / b
+}
+
+impl<T> ArrayBox<T> {
+    /// Creates a new array and returns the owning pointer to the new array.
+    pub fn new(num: usize) -> Self {
```
stjepang: We should really switch to using [`alloc`](https://doc.rust-lang.org/nightly/std/alloc/fn.alloc.html) and [`Layout`](https://doc.rust-lang.org/nightly/std/alloc/struct.Layout.html), which became stable in Rust 1.28.
Then the whole function simplifies to just a few lines of code.

jeehoonkang: Thanks :)  Now it's much simpler!

stjepang:
> The main reason is that `Atomic<T>` should really be a word---in order to support atomic operations---but `Box<T>` does not fit in a word when `T` is not `Sized`.

Right. So perhaps we could do the following.

If `T` is a sized type, allocate the value on the heap as usual.

If `T` is unsized, allocate on the heap space for `(&T, T)`. Then we write the value into the second field, and a pointer to it into the first field. It could be implemented roughly like this:

```rust
impl<T: ?Sized> Owned<T> {
    fn new(value: T) -> Owned<T> { // Assumes feature `unsized_locals` is available.
        unsafe {
            if is_sized::<T>() {
                // TODO: Allocate as usual.
            } else {
                let (layout, offset) = Layout::<&T>::new().extend(Layout::for_value(&value)).unwrap();
                let dest = Layout::alloc(layout);

                // Make a fat pointer to `value`.
                let mut ptr = &value as *const T;
                // Change the address field in `ptr` to point to `dest.add(offset)`.
                let ptr_ptr = &mut ptr as *mut _ as *mut usize;
                ptr_ptr.write(ptr as *const u8);

                // TODO: Write `value` into `dest.add(offset)`.
                // TODO: Write `ptr` into `dest`.

                Owned {
                    data: dest as usize,
                    _marker: PhantomData,
                }
            }
        }
    }
}
```

Now, in order to dereference an `Owned<T>`, we check if `T` is sized. If it is sized, we dereference as usual. If it is unsized, we load the `*const T` stored at `self.data` and just dereference it.

jeehoonkang:
I think that's a great idea.  Do you know if it's implementable in today's stable?  If so, I'll try to get it done.

stjepang:
> Do you know if it's implementable in today's stable?

I believe it is. Just a tip: you might need this function when implementing DST support:
https://github.com/rust-lang/rust/pull/55785

jeehoonkang:
@stjepang I think we cannot implement `is_sized::<T>()` in today's stable, and in fact, also in today's nightly.  My Google search didn't give me any useful references...

stjepang:
@jeehoonkang You could hack it with `size_of::<&T>() == size_of::<usize>()`. :)

jeehoonkang:
@stjepang I implemented a prototype here: https://github.com/jeehoonkang/crossbeam/tree/dst

- I'm pretty sure it'll work, but it's not tested at all.  Needs test code to be merged upstream.

- It's tricky to get it compiled in stable Rust.  In order to do so, I need to duplicate 13 functions for stable Rust, because we have to opt-out the support for unsized types in stable Rust.  For example:

  ```rust
   /// Returns a bitmask containing the unused least significant bits of an aligned pointer to `T`.
   #[inline]
   #[cfg(feature = "nightly")]
   fn low_bits<T: ?Sized>() -> usize {
       (1 << T::align_of().trailing_zeros()) - 1
   }

   /// Returns a bitmask containing the unused least significant bits of an aligned pointer to `T`.
   #[inline]
   #[cfg(not(feature = "nightly"))]
   fn low_bits<T>() -> usize {
       (1 << T::align_of().trailing_zeros()) - 1
   }
  ```

  But it'll be very verbose and I don't like to go across that bridge...  I'd rather wait for required nightly features to be stabilized.

  Here are the 13 functions that need to be duplicated:

  ```
  low_bits
  Atomic::new
  fmt::Pointer for Atomic
  From<*const T> for Atomic
  Owned::new
  Owned::from_raw
  Drop for Owned
  Borrow for Owned
  BorrowMut for Owned
  AsRef for Owned
  AsMut for Owned
  Shared::as_raw
  From<*const T> for Shared
  ```

jeehoonkang:
Now I think introducing the storage trait (this PR) is better than utilizing unsized types (https://github.com/jeehoonkang/crossbeam/tree/dst) in supporting dynamically sized types in Crossbeam, even after all necessary features for unsized types are stabilized.

First, storage trait is more general.  In theory, it can support not only dynamically sized types but also allocations in different regions (e.g. those for CPU memory and those for GPU memory) and other use cases.  I believe it's better to treat allocation and storage as a first-class citizen.

Second, the implementation based on the storage trait will be simpler.  In essence, somewhere we need to apply different logics to sized and unsized types.  This PR does that using a trait, which is exactly designed for this purpose, while the implementation based on unsized types will do that using specialization, which is non-parametric and likely introduces more cognitive loads.

For those reasons, I'd like to pursue this PR further.  @stjepang, what do you think?

jeehoonkang:
@stjepang any opinions on this issue?

stjepang:
I'm still unsure whether adding this feature is necessary since the drawbacks currently outweight the benefits. The balance in my mind goes like this:

Benefits:

* Better performance (fewer allocations and less indirection). But we still don't know how much it is improved. I guess a proof that this PR brings us the desired benefits would be to make a simple benchmark: an example that uses `Atomic<Array<T>>` (like in the deque) and then another one without the extra indirection. The second version needs to be meaningfully faster.

Drawbacks:

* More complicated types and increased API surface.
* More code to maintain.
* There are not many places that really need this (the skiplist uses custom allocation, while the performance benefits in deque are dubious, and we could use custom allocation there too). If we used custom allocation (like in the skiplist) in all the currently known use cases this PR is trying to support, the total amount of code would be less that this PR itself!

I'm especially hesitant because this crate is not very easy to use already and would therefore try to avoid adding more complexity to it, if possible.

Feature `unsized_locals` got good progress recently and it seems the lang team is pretty confident we want to stabilize it eventually: https://github.com/rust-lang/rust/issues/48055 So it seems there might be an easier solution in the future.

jeehoonkang:
Superseded by #452

Issue #190, opened 2018-10-22
General versioning consistency of the crossbeam family


silwol:
When packaging crossbeam for Debian, the team stumbled upon some inconsistencies in the crossbeam crate family.
This is a quick overview of the current state of the dependency tree:
```
╭───────────────╮
│crossbeam 0.4.1│
╰───────────────╯
 │   │   │   └───────────────────────────────────────────────┐
 │   ▼   └──────────────────┐                                ▼
 │  ╭─────────────────────╮ │ ╭─────────────────────╮  ╭───────────────────────╮
 │  │crossbeam-deque 0.5.2│ │ │crossbeam-deque 0.6.1│  │crossbeam-channel 0.2.6│
 │  ╰─────────────────────╯ │ ╰─────────────────────╯  ╰───────────────────────╯
 │    │              │      │   │      │                │       │
 │    │              ▼      ▼   ▼      │   ┌────────────┘       ▼
 │    │       ╭─────────────────────╮  │   │    ╭─────────────────────╮
 │    │       │crossbeam-epoch 0.5.2│  │   │    │crossbeam-epoch 0.6.0│
 │    │       ╰─────────────────────╯  │   │    ╰─────────────────────╯
 │    │                            │   │   │   ┌──────────┘
 │    └────────────────────────┐   │   │   │   │
 └─────────────────────────┐   │   │   │   │   │
                           ▼   ▼   ▼   ▼   ▼   ▼
                          ╭─────────────────────╮
                          │crossbeam-utils 0.5.0│
                          ╰─────────────────────╯
```
The major problematic points for us are:
- crossbeam has multiple versions of crossbeam-epoch in it's dependency tree (as mentioned in https://github.com/crossbeam-rs/crossbeam-deque/issues/17)
- The crossbeam-deque dependency of crossbeam is 0.5, not the latest version 0.6 (which might be pulled in by other crates depending on it)

I don't know the details and internals of the crossbeam project, but from a user and packager's perspective it would be easiest and least confusing if the family in itself would have only the latest version of each crate in it's dependency tree.

jeehoonkang:
#210 partially solves this problem.

I think, maybe, it would be convenient to release crossbeam families in lock-step: releasing crossbeam-* 0.x at the same time.  This scheme can be used after reaching 1.0, too.

stjepang:
An interesting issue to follow: https://github.com/tokio-rs/tokio/issues/572

stjepang:
[...]

PR #217, opened 2018-11-05
Debloat crossbeam-utils

stjepang:
This PR accomplishes several things:

* Moves `crossbeam_utils::AtomicCell` and `crossbeam_utils::sync` into `crossbeam`. This is a conservative decision because I'd prefer not to commit to putting those tools into any particular subcrate at this moment, but we should think about eventually moving them into a dedicated crate in the future. We also want to keep `crossbeam-utils` lean so that it doesn't pull in many dependencies.

* Refactors the `#[cfg(...)]` shenanigans related to features `std` and `nightly`. Every crate that has those features now has a standardized template at the top of `lib.rs`. All parts of crates that use `std` are now placed into a dedicated `cfg_if!` block.

* Tweaks `lib.rs` in the main `crossbeam` crate so that docs get nicely rendered. Now modules `channel`, `deque`, and `thread` use the original docs from subcrates, which means we don't have to write those poor substitutes for the real docs anymore.

### DIFF
--- a/crossbeam-utils/src/thread.rs
+++ b/crossbeam-utils/src/thread.rs
@@ -85,10 +85,8 @@ use std::io;
 use std::marker::PhantomData;
 use std::mem;
 use std::panic;
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, mpsc};
 use std::thread;
-
-use sync::WaitGroup;
 
 type SharedVec<T> = Arc<Mutex<Vec<T>>>;
 type SharedOption<T> = Arc<Mutex<Option<T>>>;
@@ -116,9 +114,10 @@ pub fn scope<'env, F, R>(f: F) -> thread::Result<R>
 where
     F: FnOnce(&Scope<'env>) -> R,
 {
+    let (tx, rx) = mpsc::channel();
     let scope = Scope {
         handles: SharedVec::default(),
-        wg: WaitGroup::new(),
+        chan: tx,
         _marker: PhantomData,
     };
 
@@ -126,7 +125,8 @@ where
     let result = panic::catch_unwind(panic::AssertUnwindSafe(|| f(&scope)));
 
     // Wait until all nested scopes are dropped.
-    scope.wg.wait();
+    drop(scope.chan);
+    let _ = rx.recv();
 
     // Join all remaining spawned threads.
     let panics: Vec<_> = {
@@ -157,12 +157,13 @@ where
     }
 }
 
+/// A scope for spawning threads.
 pub struct Scope<'env> {
     /// The list of the thread join handles.
     handles: SharedVec<SharedOption<thread::JoinHandle<()>>>,
 
     /// Used to wait until all subscopes all dropped.
-    wg: WaitGroup,
+    chan: mpsc::Sender<()>,
 
     /// Borrows data with lifetime `'env`.
     _marker: PhantomData<&'env ()>,
@@ -206,6 +207,7 @@ impl<'env> fmt::Debug for Scope<'env> {
 
 /// Scoped thread configuration. Provides detailed control over the properties and behavior of new
 /// scoped threads.
+#[derive(Debug)]
 pub struct ScopedThreadBuilder<'scope, 'env: 'scope> {
     scope: &'scope Scope<'env>,
     builder: thread::Builder,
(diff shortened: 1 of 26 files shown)


crossbeam-utils/src/thread.rs
```diff
@@ -116,17 +114,19 @@ pub fn scope<'env, F, R>(f: F) -> thread::Result<R>
 where
     F: FnOnce(&Scope<'env>) -> R,
 {
+    let (tx, rx) = mpsc::channel();
     let scope = Scope {
         handles: SharedVec::default(),
-        wg: WaitGroup::new(),
+        chan: tx,
```
jeehoonkang: We're reusing mpsc here because wait group is now in `crossbeam`, right?  Generally speaking, putting things in `crossbeam` disallows sub-crates to depend each other...

stjepang: Yeah :( We need to figure out how to solve this problem, but I'd rather do that later than block the new release on the decision.

jeehoonkang: How about the other way?  Specifically, how about putting miscellaneous things in `crossbeam-utils` rather than `crossbeam`, and then continuously migrating subsets of `crossbeam-utils` into their own sub-crate?

stjepang: That strategy is antithetical to backwards compatibility.

If you break off a subset of `crossbeam-utils`, then you break compatibility with its latest version.

However, if you break off a subset of `crossbeam`, then you don't break compatibility with its latest version. The subset goes into a new crate and you reexport it into `crossbeam` so its API stays the same.

Vtec234: I would agree with both - we can't have subcrates depend on `crossbeam`, but we also don't want  to break subcrates with every release. Perhaps an unstable crate for all experimental/miscelaneous stuff could exist, like `crossbeam-utils` but with no backwards compatibility guarnatees? Then other subcrates can depend on it and if it breaks, we can modify them in the same merge/PR, thus not breaking anything (except the experimental crate of course). Does this make any sense?

stjepang: In my view, `crossbeam` already is the unstable/experimental crate. It's totally OK to break compatibility in `crossbeam` because most parts of it (its subcrates) will stay the same between consecutive versions. They usually won't change in the dependency graph.

On the other hand, smaller crates are those that are expected to be more stable. We should break off parts of `crossbeam` into dedicated crates when they mature.

The problem, of course, is that `crossbeam` is big and we can't depend on the whole `crossbeam` within smaller subcrates. That would create cycles anyway.

Personally, I think we should just shrug off minor annoyances like the one that came up here. Replacing `WaitGroup` with `std::sync::mpsc` is not the end of the world. There were several similar instances where we wanted a small crate to depend on bigger one, but it's never been something that isn't trivial to get around.

Yes, it's an eye-sore, but I just think it's such a non-issue that any possible remedy would be worse than the problem itself. :)

jeehoonkang: Yeah, it's not about `WaitGroup`: it's a minor thing.  The discussion is on the crate organization.

It seems `crossbeam-utils` currently serves two roles: (1) collection of *solid* synchronization libraries, and (2) collection of experimental features.  This PR tries to split it and move experimental features into the main crate.

I'd like to propose moving experimental features into a new **unpublished** crate, say `crossbeam-nursery`, whose stability is not a concern.  The main crate will simply re-export features in the nursery, and other crates will use the nursery.  By doing so we can achieve both relative stability of `crossbeam-utils` and availability of experimental features by separating the two roles of `crossbeam-utils` into two crates.

stjepang: > It seems `crossbeam-utils` currently serves two roles: (1) collection of _solid_ synchronization libraries, and (2) collection of experimental features.

Just to clarify, this is the current state of `crossbeam-utils`:
https://docs.rs/crossbeam-utils/0.5.0/crossbeam_utils/

Note that `AtomicCell` and `sync::{ShardedLock,WaitGroup}` have never been published. So `crossbeam-utils` consists of: (1) scoped threads, (2) `CachePadded`, (3) `AtomicConsume`.

This PR won't split the latest version of `crossbeam-utils`. I'm just moving unpublished experimental stuff out of it.

> I'd like to propose moving experimental features into a new **unpublished** crate, say `crossbeam-nursery`, whose stability is not a concern.

Two problems with that:

1. If nursery is an unstable crate other crossbeam crates depend on, then every time nursery breaks compatibility, we have to re-publish all other crates in order not to end up with multiple dependencies of nursery in the graph (see #190).

2. `crossbeam-utils` wouldn't be able to depend on nursery at all because nursery would depend on `parking_lot`. If `crossbeam-utils` has a long list of dependencies, people will not use it in their libraries.

jeehoonkang:
Okay---let's merge it for the time being, and later try to migrate those in the nursery into their own crate.

bors r+

PR #219, opened 2018-11-06
Fix links in readme

senden9:


jeehoonkang:
Good catch, thanks!

bors r+

PR #224, opened 2018-11-08
Prepare v0.5.0

stjepang:
Let's release new versions of crates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Version 0.5.0
+
+- Update `crossbeam-channel` to 0.3.
+- Update `crossbeam-utils` to 0.6.
+- Add `AtomicCell`.
+
 # Version 0.4.1
 
 - Fix a double-free bug in `MsQueue` and `SegQueue`.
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# Version 0.5.0
+
+- Update `crossbeam-channel` to 0.3.
+- Update `crossbeam-utils` to 0.6.
+- Add `AtomicCell`.
```
Vtec234: What about `ShardedLock` and `WaitGroup`?

crossbeam-deque/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Version 0.6.1
 
+- Update `crosbeam-utils` to `0.6`.
+
+# Version 0.6.1
```
Vtec234: Double header?

Vtec234:
Does this fix the two-versions-of-the-same-package issue?

stjepang:
Thanks, fixed!

> Does this fix the two-versions-of-the-same-package issue?

Yes, we even have a check on Travis that verifies this.

PR #226, opened 2018-11-11
Make the lifetime in Scope<'env> invariant

### LINKED ISSUE #225: Soundness problem of thread::Scope::spawn() introduced in 0.5.0
This is the minimal reproduction of a problem I ran into:
```
extern crate crossbeam;

#[derive(Debug, PartialEq)]
struct SomeStruct {
    value: usize
}

fn main() -> Result<(), std::io::Error> {
    crossbeam::thread::scope(|scope| {
        for i in 0..2 {
            let d = SomeStruct { value: i };
            println!("Triggering {:?}", d);
            scope.spawn(|_| {
                println!("{:?}", d);
            });
            println!("Triggered {:?}", d);
        }
    }).unwrap();
    Ok(())
}
```

The output on my machine (Rust 1.30.1 on MacOS) is:
```
Triggering SomeStruct { value: 0 }
Triggered SomeStruct { value: 0 }
Triggering SomeStruct { value: 1 }
SomeStruct { value: 1 }
Triggered SomeStruct { value: 1 }
SomeStruct { value: 4470883725 }
```
My mistake was not moving the struct into the closure. When I use 0.4.1 (with the slightly different API), a compile time error points me to the mistake.

`@muvion` on `#rust-beginners` figured out that the lifetime parameters of spawn() have not just been renamed, but also changed in usage in 95c3e2e46ef151754e2c9dfb316eda9b15754eaa. Now the closure's lifetime is not related to the return value's lifetime anymore.

stjepang:
Fixes #225


crossbeam-utils/src/thread.rs
```diff
@@ -295,7 +295,7 @@ unsafe impl<'scope, T> Send for ScopedJoinHandle<'scope, T> {}
 unsafe impl<'scope, T> Sync for ScopedJoinHandle<'scope, T> {}
 
 /// A handle to a scoped thread
-pub struct ScopedJoinHandle<'scope, T: 'scope> {
+pub struct ScopedJoinHandle<'scope, T> {
```
stjepang: We don't need `T: 'scope` here. Note that [`JoinHandle<T>`](https://doc.rust-lang.org/nightly/std/thread/struct.JoinHandle.html) doesn't require `T: 'static` either.

crossbeam-utils/src/thread.rs
```diff
@@ -115,7 +115,7 @@ where
     F: FnOnce(&Scope<'env>) -> R,
 {
     let (tx, rx) = mpsc::channel();
-    let scope = Scope {
+    let scope = Scope::<'env> {
```
stjepang: Not necessary, but makes things more explicit.

crossbeam-utils/src/thread.rs
```diff
@@ -242,7 +242,7 @@ impl<'scope, 'env> ScopedThreadBuilder<'scope, 'env> {
             let result = Arc::clone(&result);
 
             // A clone of the scope that will be moved into the new thread.
-            let scope = Scope {
+            let scope = Scope::<'env> {
```
stjepang: Not necessary, but makes things more explicit.

crossbeam-utils/src/thread.rs
```diff
@@ -165,8 +165,8 @@ pub struct Scope<'env> {
     /// Used to wait until all subscopes all dropped.
     chan: mpsc::Sender<()>,
 
-    /// Borrows data with lifetime `'env`.
-    _marker: PhantomData<&'env ()>,
+    /// Borrows data with invariant lifetime `'env`.
+    _marker: PhantomData<&'env mut &'env mut ()>,
```
stjepang: This is what fixes the bug.

stjepang:
I'll merge this early and do a minor version jump because it's a rather serious issue.

bors r+

crossbeam-utils/src/thread.rs
```diff
@@ -165,8 +165,8 @@ pub struct Scope<'env> {
     /// Used to wait until all subscopes all dropped.
     chan: mpsc::Sender<()>,
 
-    /// Borrows data with lifetime `'env`.
-    _marker: PhantomData<&'env ()>,
+    /// Borrows data with invariant lifetime `'env`.
+    _marker: PhantomData<&'env mut &'env ()>,
```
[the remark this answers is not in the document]
Vtec234: My understanding of variance in Rust isn't great, does making this a double `&mut &` reference make it invariant w.r.t `'env`?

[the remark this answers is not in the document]
stjepang: The rule is that `&'A mut B` is covariant over `A` and invariant over `B`.

So if we write `&'env mut &'env ()`, then it's invariant over `&'env ()` and therefore also over `'env`.

crossbeam-utils/src/thread.rs
Thomasdezeeuw: @stjepang I think this new for Rust 2018/NLL, so it might be worth it to leave it to support older Rust versions.

PR #247, opened 2018-12-03
Remove Clone for Guard

stjepang:
While `Guard::clone()` is a nice convenience, it's not really a method we *need*.

By removing the clone impl, we enable some kinds of API. For example, a data structure might want to return a `Guard` to the user and then retrieve it back knowing it has never been cloned.

@Amanieu needs to forbid `Guard::clone()` in a certain use case. I don't remember the exact details, so perhaps he could comment what it is about.

### DIFF
--- a/crossbeam-epoch/src/guard.rs
+++ b/crossbeam-epoch/src/guard.rs
@@ -1,6 +1,5 @@
 use core::fmt;
 use core::mem;
-use core::ptr;
 
 use atomic::Shared;
 use collector::Collector;
@@ -63,15 +62,6 @@ use internal::Local;
 /// assert!(epoch::is_pinned());
 /// drop(guard2);
 /// assert!(!epoch::is_pinned());
-/// ```
-///
-/// The same can be achieved by cloning guards:
-///
-/// ```
-/// use crossbeam_epoch as epoch;
-///
-/// let guard1 = epoch::pin();
-/// let guard2 = guard1.clone();
 /// ```
 ///
 /// [`pin`]: fn.pin.html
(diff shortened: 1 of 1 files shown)


stjepang:
> I believe that some code in skiplist is currently using guard clone, but that should be easy to fix.

I vaguely remember having `Guard::clone()` in the skip list, too, but now it seems it isn't used anywhere. :)

Vtec234:
> By removing the clone impl, we enable some kinds of API. For example, a data structure might want to return a Guard to the user and then retrieve it back knowing it has never been cloned.

What invariant is guaranteed by forbidding the `Guard` from being cloned?

stjepang:
@Vtec234 Consider this sendable `MyHandle`:

```rust
pub struct MyHandle {
    local_handle: LocalHandle,
    guard: Option<Guard>,
}

unsafe impl Send for MyHandle {}

impl MyHandle {
    pub fn pin(&self) -> &Guard {
        // ...
    }
}
```

If we allow cloning guards, then this becomes a problem:

```rust
let my_handle: MyHandle = ...;
let guard = my_handle.pin().clone();

thread::spawn(move || {
    let guard2 = my_handle.pin(); // Using `pin()` concurrently from two threads!
}}
```

In that case, `MyHandle::pin()` would have to be an unsafe function, which is a little annoying.

Vtec234:
Oh I see, in that case what's the use case for sending a `LocalHandle`? Since `Collector` is `Sync`, what's different in sending `LocalHandle` vs a `&Collector`, except that `register` had some overhead?

stjepang:
Maybe @Amanieu can answer this question?

Amanieu:
In short, threads in my application can be killed at certain points, and cleanup of that thread's resources will need to be done by another thread. This includes unpinning that thread's `Guard` and freeing its `LocalHandle`. For this to work, `MyHandle` needs to be `Send`, which only works if there is exactly 1 Guard per `MyHandle`.

Vtec234:
Ah ok, so you're GC-ing the GC? :thinking:

PR #248, opened 2018-12-03
Add num_shards argument to ShardedLock::new

stjepang:
This removes the dependency on `num_cpus`. If we also remove dependency on `parking_lot`, then `ShardedLock` can be moved into `crossbeam-utils`.

The `num_shards` arguments is somewhat reminiscent of `concurrencyLevel` in [this](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html#ConcurrentHashMap-int-float-int-) `ConcurrentHashMap` constructor.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,6 @@ std = ["crossbeam-epoch/std", "crossbeam-utils/std"]
 [dependencies]
 cfg-if = "0.1"
 lazy_static = "1.1.0"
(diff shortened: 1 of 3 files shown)


jeehoonkang:
@stjepang I'd like to merge this PR.  Do you want to wait for comments or reviews from others?

stjepang:
Thanks for taking a look!

bors r+

PR #262, opened 2018-12-11
Move AtomicCell into crossbeam-utils

stjepang:
At this point, I feel pretty confident about the current state of `AtomicCell` in the sense that we probably won't make any breaking changes to its public API in the future.

It's also a reasonably small utility without dependencies that deserves living in `crossbeam-utils`. The Rust compiler would like to use it in place of its `LockCell`, and already has `crossbeam-utils` whitelisted. See [this PR](https://github.com/rust-lang/rust/pull/56614#issuecomment-446134372).

Let's move `AtomicCell` into `crossbeam-utils` so that people can use it without depending on whole `crossbeam`.


stjepang:
travis r+

bjorn3:
@stjepang ^ s/travis/bors/ I think.

stjepang:
@bjorn3 Hah yes, I've deleted that comment and posted a new one :)

PR #263, opened 2018-12-13
crossbeam-epoch: Add Shared::deref_mut()

tsnoam:
It is possible that the user would like to have one thread to be able to
update the value of the object pointed by Atomic. For that, they will
need to have a mutable refernce towards that object.

This is an unsafe function because there is no guarantee that there are
more threads attempting to read/write from/to the actual object in the
same time.
In order to proivde safety the user must be able to know that there are
no concurrent accesses towards the object itself and/or use
crossbeam_utils::atomic::AtomicCell to provide the needed guarantees.

### DIFF
--- a/crossbeam-epoch/src/atomic.rs
+++ b/crossbeam-epoch/src/atomic.rs
@@ -942,23 +942,12 @@ impl<'g, T> Shared<'g, T> {
     ///
     /// # Safety
     ///
-    /// * Dereferencing a pointer is unsafe because it could be pointing to invalid memory.
-    ///
-    /// * Another concern is the possiblity of data races due to lack of proper synchronization.
-    ///   For example, consider the following scenario:
-    ///
-    ///   1. A thread creates a new object: `a.store(Owned::new(10), Relaxed)`
-    ///   2. Another thread reads it: `*a.load(Relaxed, guard).as_ref().unwrap()`
-    ///
-    ///   The problem is that relaxed orderings don't synchronize initialization of the object with
-    ///   the read from the second thread. This is a data race. A possible solution would be to use
-    ///   `Release` and `Acquire` orderings.
-    ///
     /// * There is no guarantee that there are no more threads attempting to read/write from/to the
     ///   actual object at the same time.
     ///
-    ///   The user must be know that there are no concurrent accesses towards the object itself
-    ///   and/or use `crossbeam_utils::atomic::AtomicCell` to wrap the object.
+    ///   The user must know that there are no concurrent accesses towards the object itself.
+    ///
+    /// * Other than the above, all safety concerns of `deref()` applies here.
     ///
     /// # Examples
     ///
--- a/crossbeam-epoch/src/atomic.rs
+++ b/crossbeam-epoch/src/atomic.rs
@@ -936,6 +936,57 @@ impl<'g, T> Shared<'g, T> {
         &*self.as_raw()
     }
 
+    /// Dereferences the pointer.
+    ///
+    /// Returns a mutable reference to the pointee that is valid during the lifetime `'g`.
+    ///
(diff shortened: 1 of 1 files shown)


crossbeam-epoch/src/atomic.rs
```diff
@@ -936,6 +936,57 @@ impl<'g, T> Shared<'g, T> {
...
     }
 
+    /// Dereferences the pointer.
+    ///
+    /// Returns a mutable reference to the pointee that is valid during the lifetime `'g`.
+    ///
+    /// # Safety
+    ///
+    /// * Dereferencing a pointer is unsafe because it could be pointing to invalid memory.
+    ///
```
stjepang: Maybe add another item saying "It is only valid to mutably dereference a pointer if no other threads will dereference it at the same time."

stjepang: Oh, I see you've added something similar below.

crossbeam-epoch/src/atomic.rs
```diff
@@ -936,6 +936,57 @@ impl<'g, T> Shared<'g, T> {
...
+    ///   2. Another thread reads it: `*a.load(Relaxed, guard).as_ref().unwrap()`
+    ///
+    ///   The problem is that relaxed orderings don't synchronize initialization of the object with
+    ///   the read from the second thread. This is a data race. A possible solution would be to use
+    ///   `Release` and `Acquire` orderings.
+    ///
+    /// * There is no guarantee that there are no more threads attempting to read/write from/to the
+    ///   actual object at the same time.
+    ///
+    ///   The user must be know that there are no concurrent accesses towards the object itself
```
stjepang: Without "be" - "The user must know..."

crossbeam-epoch/src/atomic.rs
```diff
@@ -936,6 +936,57 @@ impl<'g, T> Shared<'g, T> {
...
+    ///   and/or use `crossbeam_utils::atomic::AtomicCell` to wrap the object.
```
stjepang: Well, if another thread may access this object at the same time, it is not valid to call `.deref_mut()` even if we have `Shared<AtomicCell<_>>`. I'd just delete this sentence. :)

crossbeam-epoch/src/atomic.rs
```diff
@@ -936,6 +936,57 @@ impl<'g, T> Shared<'g, T> {
...
+    ///
+    /// # Safety
+    ///
+    /// * Dereferencing a pointer is unsafe because it could be pointing to invalid memory.
+    ///
+    /// * Another concern is the possiblity of data races due to lack of proper synchronization.
+    ///   For example, consider the following scenario:
+    ///
+    ///   1. A thread creates a new object: `a.store(Owned::new(10), Relaxed)`
+    ///   2. Another thread reads it: `*a.load(Relaxed, guard).as_ref().unwrap()`
```
Vtec234: `as_ref().unwrap()` -> `deref_mut()`?

crossbeam-epoch/src/atomic.rs
tsnoam: oic. I'll delete this sentence. I will also move it to be first in the list because it's the important addition on top of `deref()`

crossbeam-epoch/src/atomic.rs
tsnoam: Actually, this part of the documentation is copy/paste from `deref()`.
I think that instead of explicitly copy/pasting the `deref()` documentation, I'll refer to them.

tsnoam:
I've pushed documentation fixes, per the CR

Vtec234:
bors r+

Nightly failure seems to be an unrelated linker issue.

Issue #246, opened 2018-12-03
Use std::alloc to allocate skip list nodes


stjepang:
Let's resolve [this TODO](https://github.com/crossbeam-rs/crossbeam/blob/e7d9c298317fa985ae8ca39bc108457d57e9b8ad/crossbeam-skiplist/src/base.rs#L101) stating and replace this silly allocation hack to use `std::alloc` instead.

cc @Amanieu @SimonSapin

stjepang:
Resolved in https://github.com/crossbeam-rs/crossbeam/pull/268.

PR #268, opened 2018-12-16
skiplist: switch to alloc API

tobz:
This addresses #246 by removing the allocation hack that was in place before.

Signed-off-by: Toby Lawrence <toby@nuclearfurnace.com>

### DIFF
--- a/crossbeam-skiplist/src/base.rs
+++ b/crossbeam-skiplist/src/base.rs
@@ -1,6 +1,6 @@
 //! TODO: docs
 
-use alloc::vec::Vec;
+use alloc::alloc::{alloc, dealloc, Layout};
 use core::borrow::Borrow;
 use core::cmp;
 use core::fmt;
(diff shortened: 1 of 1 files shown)


tobz:
@stjepang I think this is all set now.

tobz:
Hmm, not really sure why it thinks that `std::alloc` doesn't exist... that's straight out of the Rust docs themselves. O_o

stjepang:
It's because of [these shenanigans](https://github.com/crossbeam-rs/crossbeam/blob/711fd376b99aefaea9a3167033bd56a354d87dc2/crossbeam-skiplist/src/lib.rs#L13-L22) we use to support no_std + alloc environments. You probably need `use alloc::alloc` instead of `use std::alloc`.

tobz:
Ahhhh, alright.  Lemme whip up the fix.

tobz:
Looks like it's an unrelated `crossbeam-channel` failure at this point?

stjepang:
That is an unrelated regression on nightly Rust: https://github.com/rust-lang/rust/issues/56867

crossbeam-skiplist/src/base.rs
```diff
@@ -99,11 +99,8 @@ impl<K, V> Node<K, V> {
     /// with null pointers. However, the key and the value will be left uninitialized, and that is
     /// why this function is unsafe.
     unsafe fn alloc(height: usize, ref_count: usize) -> *mut Self {
-        // TODO(stjepang): Use the new alloc API instead of this hack once it becomes stable.
-        let cap = Self::size_in_u64s(height);
-        let mut v = Vec::<u64>::with_capacity(cap);
-        let ptr = v.as_mut_ptr() as *mut Self;
-        mem::forget(v);
+        let layout = Self::get_layout(height);
+        let ptr = alloc(layout) as *mut Self;
```
SimonSapin: This needs to check for null and call `handle_alloc_error(layout)`.

stjepang: Oh, thanks for pointing this out! I'm used to `Box::new()` always working so the possibility of a failure didn't even cross my mind.

@tobz Would you like to submit a fix for this?

tobz: I'll submit a fix, yeah.

PR #272, opened 2018-12-18
channel: implementation for receiving multiple items at once

tobz:
This PR implements `try_recv_many`, which allows a caller to receive multiple items in a single call.  It is provided generally for all channels, but has a specific implementation override for bounded channels, and a few caveats:

- there are no timeout-based or blocking variants
- this will do almost nothing for some channels like after/zero because it uses the non-blocking try semantics
- callers must allocate an `Extend` capable container themselves
- mixed recv/recv_many performance may be suboptimal

This method fills a niche use case: single consumers that can benefit from grabbing and processing multiple items at a time.

### DIFF
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -1,7 +1,7 @@
 //! The channel interface.
 
 use std::fmt;
-use std::iter::FusedIterator;
+use std::iter::{FusedIterator, once};
 use std::mem;
 use std::panic::{RefUnwindSafe, UnwindSafe};
 use std::sync::Arc;
@@ -662,6 +662,76 @@ impl<T> Receiver<T> {
                 }
             }
             ReceiverFlavor::Never(chan) => chan.try_recv(),
+        }
+    }
+
+    /// Attempts to receive multiple messages from the channel without blocking.
+    ///
+    /// This method will either receive multiple messages from the channel immediately, up to `limit`,
+    /// or return an error if the channel is either empty or disconnected.
+    ///
+    /// If called on a zero-capacity channel, this method will receive a message only if there
+    /// happens to be a send operation on the other side of the channel at the same time.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use crossbeam_channel::{unbounded, TryRecvError};
+    ///
+    /// let mut items = Vec::new();
+    ///
+    /// let (s, r) = unbounded();
+    /// assert_eq!(r.try_recv_many(&mut items, 2), Err(TryRecvError::Empty));
+    ///
+    /// s.send(4).unwrap();
+    /// s.send(5).unwrap();
+    /// s.send(6).unwrap();
+    /// s.send(7).unwrap();
+    /// drop(s);
+    ///
+    /// assert_eq!(r.try_recv_many(&mut items, 2), Ok(2));
+    /// assert_eq!(r.try_recv_many(&mut items, 2), Ok(2));
+    /// assert_eq!(r.try_recv_many(&mut items, 2), Err(TryRecvError::Disconnected));
+    /// ```
+    pub fn try_recv_many<E: Extend<T>>(&self, buf: &mut E, limit: usize) -> Result<usize, TryRecvError> {
+        let mut n = 0;
+
+        loop {
+            if n == limit {
+                return Ok(n)
+            }
+
+            let result = match &self.flavor {
+                ReceiverFlavor::Array(chan) => return chan.try_recv_many(buf, limit),
+                ReceiverFlavor::List(chan) => chan.try_recv(),
+                ReceiverFlavor::Zero(chan) => chan.try_recv(),
+                ReceiverFlavor::After(chan) => {
+                    let msg = chan.try_recv();
+                    unsafe {
+                        mem::transmute_copy::<Result<Instant, TryRecvError>, Result<T, TryRecvError>>(
+                            &msg,
+                        )
+                    }
+                }
+                ReceiverFlavor::Tick(chan) => {
+                    let msg = chan.try_recv();
+                    unsafe {
+                        mem::transmute_copy::<Result<Instant, TryRecvError>, Result<T, TryRecvError>>(
+                            &msg,
+                        )
+                    }
+                }
+                ReceiverFlavor::Never(chan) => chan.try_recv(),
+            };
+
+            match result {
+                Ok(msg) => {
+                    buf.extend(once(msg));
+                    n += 1;
+                },
+                Err(_) if n != 0 => return Ok(n),
+                Err(err) => return Err(err),
+            }
         }
     }
 
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -702,6 +702,7 @@ impl<T> Receiver<T> {
             }
 
             let result = match &self.flavor {
+                //ReceiverFlavor::Array(chan) => chan.try_recv(),
                 ReceiverFlavor::Array(chan) => return chan.try_recv_many(buf, limit),
                 ReceiverFlavor::List(chan) => chan.try_recv(),
                 ReceiverFlavor::Zero(chan) => chan.try_recv(),
--- a/crossbeam-channel/src/flavors/array.rs
+++ b/crossbeam-channel/src/flavors/array.rs
@@ -333,10 +333,8 @@ impl<T> Channel<T> {
 
         loop {
             // Load the head/tail and deconstruct them.
-            let tail = self.tail.load(Ordering::Relaxed);
             let index = head & (self.mark_bit - 1);
             let lap = head & !(self.one_lap - 1);
-            let tail_index = tail & (self.mark_bit - 1);
 
             // Inspect the corresponding head slot.
             let slot = unsafe { &*self.buffer.add(index) };
@@ -344,6 +342,9 @@ impl<T> Channel<T> {
 
             // If the stamp is ahead of the head by 1, we may attempt to pop.
             if head + 1 == stamp {
+                let tail = self.tail.load(Ordering::Relaxed);
+                let tail_index = tail & (self.mark_bit - 1);
+
                 // See how many messages we have available to read.
                 let msgs_avail = if index < tail_index {
                     tail_index - index
--- a/crossbeam-channel/src/flavors/array.rs
+++ b/crossbeam-channel/src/flavors/array.rs
@@ -14,6 +14,8 @@
 //!   - http://www.1024cores.net/home/code-license
 
 use std::cell::UnsafeCell;
+use std::cmp;
+use std::iter::once;
 use std::marker::PhantomData;
 use std::mem;
 use std::ptr;
@@ -268,7 +270,7 @@ impl<T> Channel<T> {
             let slot = unsafe { &*self.buffer.add(index) };
             let stamp = slot.stamp.load(Ordering::Acquire);
 
-            // If the the stamp is ahead of the head by 1, we may attempt to pop.
+            // If the stamp is ahead of the head by 1, we may attempt to pop.
             if head + 1 == stamp {
                 let new = if index + 1 < self.cap {
                     // Same lap, incremented index.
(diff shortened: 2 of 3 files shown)


tobz:
Open questions: should we provide a blocking and/or timeout-able variant of this?

Thomasdezeeuw:
@tobz other that being convenient, is there any performance benefit to this?

tobz:
The intention is to explore an optimized solution that is faster than the naive looping, which would allow it to be used in `select!`.

My use case is that I cannot perform client-side batching -- sending batches themselves over the channel -- due to liveness concerns, but my workload benefits from operating on batches.

I could do the "build a batch by looping" approach in my application itself, but I'm pretty sure we can achieve this on the crossbeam-channel side more efficiently.

tobz:
Running numbers to illustrate why I'm doing this:

### Do-it-yourself batching

    let mut items = Vec::with_capacity(self.batch_size);
    loop {
        select! {
            recv(upkeep_rx) -> _ => {
                ...

                for item in &items {
                    <process item>
                }

                items.clear();
            },
            recv(self.data_rx) -> result => match result {
                Ok(msg) => {
                    // Doesn't matter if I process here or only do it in the upkeep routine: identical perf.
                    if items.len() == self.batch_size {
                        for item in &items {
                            <process item>
                        }
                    }

                    items.push(msg);
                },
                Err(e) => eprintln!("error receiving data message: {}", e),
            },
        }
    }

    INFO benchmark: rate: 4103182.2645107238 samples per second
    INFO benchmark: latency (ns): p50: 367 p90: 420 p99: 516 p999: 3545 max: 31588351

These numbers consistently hold despite the capacity of the underlying channel or the batch size.  It seems that I'm being limited by how fast the `select!` macro is able to churn through checking each case, which adds up.

### crossbeam-channel batches for me

    let mut items = Vec::with_capacity(self.batch_size);
    let upkeep_rx = tick(Duration::from_millis(100));
    loop {
        match upkeep_rx.try_recv() {
            Ok(_) => {
                let now = Instant::now();
                self.histogram.upkeep(now);

                while let Ok(msg) = self.control_rx.try_recv() {
                    self.process_control_msg(msg);
                }
            },
            _ => {},
        };

        match self.data_rx.try_recv_many(&mut items, self.batch_size) {
            Ok(_) => {
                for item in &items {
                    self.counter.update(item);
                    self.gauge.update(item);
                    self.histogram.update(item);
                }

                items.clear();
            },
            Err(TryRecvError::Empty) => {},
            Err(e) => eprintln!("error receiving data message: {}", e),
        }
    }

    INFO benchmark: rate: 12968473.320599513 samples per second
    INFO benchmark: latency (ns): p50: 88 p90: 118 p99: 1923 p999: 2509 max: 846335

As you can see, nearly 3x the throughput and a better latency profile nearly across the board.  These numbers are based on the current (923b057) code.  With the naive batching approach in b2311c5, performance is about ~10-15% less, but still just about 3x faster than the DIY batching approach.

tobz:
I've pushed my rough implementation of what I think is the most optimal receive many algorithm.  You can obviously see the code, but this was my thought process:

- grab head/tail, and return if we're disconnected/empty
- if not, figure out how far we can read ahead from our current position, up to `limit`
- calculate the new head index based on max possible read ahead
- try shoving the head pointer all the way past that target read ahead slot
- if we don't win CAS, loop and go back to 1
- if we win, loop through each slot between old head and new head, and extend the buffer
- adjust the slot stamp as you normally would

I mentioned this to @stjepang in Gitter, but it looks like there's a bug where the code goes into an endless loop if there's no print statements in `try_recv_many_optimized`, but if I uncomment a single one of those print statements, the tests seem to pass.

tobz:
I finally got this working!

Long story short, I should have followed the outline of `start_recv` more faithfully. :)

There's a single test for this in `tests/array.rs` as well as a doctest.  I still need to test the behavior/performance of having multiple readers where `recv`/`recv_many` operations are being mixed. Based on the code, though, I'm confident they work together just fine.

I also had some tests for development that were sync, async, varied the channel size, read size, etc... but that seems a little gratuitous now.  Happy to add any other tests you think would be worth adding to exercise this particular chunk of behavior.

crossbeam-channel/src/flavors/array.rs
```diff
@@ -312,6 +313,88 @@ impl<T> Channel<T> {
...
+                if self
+                    .head
+                    .compare_exchange_weak(head, new_head, Ordering::SeqCst, Ordering::Relaxed)
+                    .is_ok()
+                {
+                    // We've claimed the full range of messages.  Read each one and store it.
+                    let mut index = head_index;
+                    while index != new_head_index {
+                        let slot = unsafe { &*self.buffer.add(index) };
+                        let msg = unsafe { slot.msg.get().read() };
```
stjepang: I believe the problem lies here.

Recall that:
* If `head == head_stamp`, that means a value has been written into the slot.
* If `head != head_stamp`, that means a value is not written into the slot. If the queue is not empty, a value will be written very soon, so we just need to spin-wait until `head == head_stamp`.

Before this while loop, we've only checked that the first slot is safe for reading. But for each of the following slots, we have to spin-wait until the stamp indicates that the value has been written into the slot.

tobz:
Looking at the CI, it's not clear how the listed error could possibly be real...?

stjepang:
`Sender`, `Receiver`, and `Channel` went through refactoring in https://github.com/crossbeam-rs/crossbeam/pull/283, so maybe that's the cause of the compilation errors.

tobz:
That part is fixed now, but I'm seeing an addition overflow during sending now, it seems... still trying to figure out what changed in master that might cause that.

tobz:
This is now ready for another round of review.  There was just a snag in implementation after merging in the `mark_bit`/`one_lap` changes that my code didn't have yet.

I haven't hit a deadlock yet with `hotmic`, so I'm confident in the code as it stands.

crossbeam-channel/src/channel.rs
```diff
@@ -665,6 +665,81 @@ impl<T> Receiver<T> {
...
+                }
+                ReceiverFlavor::Never(chan) => chan.try_recv(),
+            };
+
+            match result {
+                Ok(msg) => {
+                    buf.extend(Ok::<_, ()>(msg));
+                    n += 1;
+                },
+                Err(e) => {
```
Thomasdezeeuw: This be collapsed to:

```rust
Err(ref _) if n != 0 => return Ok(n),
Err(err) => return Err(err),
```

crossbeam-channel/src/channel.rs
```diff
@@ -665,6 +665,81 @@ impl<T> Receiver<T> {
...
+                            &msg,
+                        )
+                    }
+                }
+                ReceiverFlavor::Never(chan) => chan.try_recv(),
+            };
+
+            match result {
+                Ok(msg) => {
+                    buf.extend(Ok::<_, ()>(msg));
```
Thomasdezeeuw: I think it's clearer if you use `std::iter::once`, since `extend` needs an `IntoIterator`.

crossbeam-channel/src/flavors/array.rs
```diff
@@ -410,6 +518,16 @@ impl<T> Channel<T> {
         }
     }
 
+    /// Attempts to receive messages without blocking.
+    pub fn try_recv_many<E: Extend<T>>(&self, buf: &mut E, limit: usize) -> Result<usize, TryRecvError> {
```
Thomasdezeeuw: Why the two seperate functions? (planning to reuse start_recv_many in future?)

crossbeam-channel/src/channel.rs
```diff
@@ -665,6 +665,81 @@ impl<T> Receiver<T> {
         }
     }
 
+    /// Attempts to receive multiple messages from the channel without blocking.
+    ///
+    /// This method will either receive multiple messages from the channel immediately, up to `limit`,
+    /// or return an error if the channel is empty.
```
Thomasdezeeuw: This can also return a `Disconnected` error, so maybe change the wording of this a little.

crossbeam-channel/src/channel.rs
```diff
@@ -665,6 +665,81 @@ impl<T> Receiver<T> {
...
+                return Ok(n)
+            }
+
+            let result = match &self.flavor {
+                ReceiverFlavor::Array(chan) => return chan.try_recv_many(buf, limit),
+                ReceiverFlavor::List(chan) => chan.try_recv(),
+                ReceiverFlavor::Zero(chan) => chan.try_recv(),
+                ReceiverFlavor::After(chan) => {
+                    let msg = chan.try_recv();
+                    unsafe {
```
Thomasdezeeuw: This is not ideal, but I can't think of a better way to do it...

crossbeam-channel/src/flavors/array.rs
tobz: Yeah, I'm not sure there's a good reason to have it be separate.

I started out just trying to mimic the existing method structure, but @stjepang mentioned that the `select!` is already too complex for their liking, so it's not likely that this would ever have support added... I'll go ahead and collapse this down to just `try_recv_many`.

tobz: Actually, I think it makes more sense maybe to just rename `start_recv_many` to `recv_many`.. my thought here is that I like being able to see `try_recv_many` as a separate function to show the logic of "if we got any items, we now have more free space, so notify the next waiting sender".

With everything compressed, those few lines of code get lost easily within the rest of the actual logic.

crossbeam-channel/src/channel.rs
tobz: What are you referring to specifically?

crossbeam-channel/src/channel.rs
tobz: Hmm, nice!  I'll switch it.

crossbeam-channel/src/channel.rs
tobz: Sounds good. 👍

crossbeam-channel/src/channel.rs
tobz: Done.

crossbeam-channel/src/channel.rs
Thomasdezeeuw: The use of of the unsafe `transmute_copy`.

crossbeam-channel/src/flavors/array.rs
Thomasdezeeuw: @tobz then maybe the functions need different names. Just looking at the function names we have `recv_many` and `try_recv_many`, to me this says both functions do the same thing, but `try_recv_many` will return an error if it can't recveive anything, while the `recv_many` will block/panic. However this is not the case.

Maybe `pub(crate) fn try_recv_many` and `fn recv_many_internal`? To indicate you have to call `try_recv_many` to ensure senders are notified.

tobz: Hmmm.... I think I can live with that, yeah. :)

crossbeam-channel/src/channel.rs
tobz: Ah, yeah... I just copypasta'd what `try_recv` does. :(

tobz:
Looks like grabbing nightly from rustup is broken so all the tests running under nightly fail hard.

tobz:
@stjepang I think this is ready for another/final pass?

stjepang:
So I tried benchmarking `try_recv_many()` by changing the `spsc()` function in `benchmarks/crossbeam-channel.rs`:

```rust
fn spsc(cap: Option<usize>) {
    let (tx, rx) = new(cap);

    crossbeam::scope(|scope| {
        scope.spawn(|_| {
            for i in 0..MESSAGES {
                tx.send(message::new(i)).unwrap();
            }
        });

        let mut i = 0;
        let cap = 100;
        let mut v = Vec::with_capacity(cap);
        while i < MESSAGES {
            let cnt = rx.try_recv_many(&mut v, cap).unwrap_or(0);
            if cnt == 0 {
                rx.recv().unwrap();
                i += 1;
            } else {
                i += cnt;
                v.clear();
            }
        }
    }).unwrap();
}
```

The test invocation in the `main()` function is:

```rust
run!("bounded_spsc", spsc(Some(MESSAGES)));
```

Results with `try_recv_many()`:

```
bounded_spsc              Rust crossbeam-channel   0.309 sec
```

Results with `recv()`:

```
bounded_spsc              Rust crossbeam-channel   0.161 sec
```

The reason why `try_recv_many()` is slower here is because it loads the tail on every invocation, which means the sending and the receiving thread contend on the cache line that holds the tail index.

It's probably possible to avoid contention *somehow*, but it's likely to be non-trivial and I wonder if it's really worh it in the end. What do you think, @tobz? Do you see a way to fix the slowdown due to cache line contention?

On the other hand, the `seq()` benchmark gets better with `try_recv_many()`:

```rust
fn seq(cap: Option<usize>) {
    let (tx, rx) = new(cap);

    for i in 0..MESSAGES {
        tx.send(message::new(i)).unwrap();
    }

    let mut i = 0;
    let cap = 100;
    let mut v = Vec::with_capacity(cap);
    while i < MESSAGES {
        let cnt = rx.try_recv_many(&mut v, cap).unwrap_or(0);
        if cnt == 0 {
            rx.recv().unwrap();
            i += 1;
        } else {
            i += cnt;
            v.clear();
        }
    }
}
```

The test invocation in the `main()` function is:

```rust
run!("bounded_seq", seq(Some(MESSAGES)));
```

With `try_recv_many()`:

```
bounded_seq               Rust crossbeam-channel   0.160 sec
```

With `recv()`:

```
bounded_seq               Rust crossbeam-channel   0.261 sec
```

Very nice improvement, but the point of channels is that they work well for sending messages between threads, which is the more interesting case. :)

tobz:
Let me try and take another look, with these specific benchmarks in mind.

tobz:
So, you're absolutely right: by the numbers, spsc is markedly slower using `try_recv_many` rather than just the straight line of single item `recv`s.

The problem is that in my workload, the batched receives are _much_ faster.

Receiving and processing a single item at a time: 5M messages/sec
Receiving and processing a batch (naive try_recv_many in `src/channel.rs`): 7.5M messages/sec
Receiving and processing a batch (true try_recv_many impl in `src/flavors/array.rs`): 10M messages/sec

It seems like this can be attributed to the batch receive clearing the way to be able to ram through a lot of sends in big bursts, and fwiw, this is in SPSC mode in my application.  The time spent doing the work overshadows the latency introduced by having to read the tail... while the channel itself becomes slower, it's unlocking a lot of efficiency in how I can process items.

I totally understand if you're not willing to merge this out of fear of giving people a method that they might shoot themselves in the foot with, performance-wise, but the numbers for my own workload are too good to pass up, so I can just utilize my own fork going forward.

stjepang:
Interesting, that seems promising! Can I perhaps reproduce the benchmark?

tobz:
Sure!

Checkout both hotmic (https://github.com/nuclearfurnace/hotmic, branch `feature/recv_many`) and crossbeam (branch `tobz/channel_recv_many`) in the same directory level.  Go into `hotmic` and build the benchmark example: `cargo build --example benchmark --release && RUST_LOG=debug target/release/examples/benchmark`.

By default, it will use the non-batching implementation, which should yield in the ~5M messages/sec range.

To then run it utilizing `try_recv_many`, build with the `recv_many` feature flag.  By default, it will use the "full" `try_recv_many` implementation (aka `recv_many_internal` in `src/flavors/array.rs`) but if you look on `src/channel.rs:705`. there's a commented out line that does the batching naively, same as all the other channel flavor match arms.  Just uncomment that, and comment the line below, or vise versa, to toggle between the naive and full implementation.

As mentioned in the past comment, the naive recv_many implementation should yield 7.5-8M messages/sec, while the full implementation yields 9.5-10M messages/sec.

# No batching
    [2019-01-19T13:33:51Z INFO  benchmark] rate: 5180259.256140807 samples per second
    [2019-01-19T13:33:51Z INFO  benchmark] latency (ns): p50: 381 p90: 402 p99: 489 p999: 651 max: 1396735
    [2019-01-19T13:33:52Z INFO  benchmark] total metrics pushed: 305322780

# Naive batching
    [2019-01-19T13:45:41Z INFO  benchmark] rate: 8453679.517593566 samples per second
    [2019-01-19T13:45:41Z INFO  benchmark] latency (ns): p50: 117 p90: 148 p99: 4167 p999: 6159 max: 1513471
    [2019-01-19T13:45:42Z INFO  benchmark] total metrics pushed: 501934458

# Full batching
    [2019-01-19T13:38:23Z INFO  benchmark] rate: 10356512.486930337 samples per second
    [2019-01-19T13:38:23Z INFO  benchmark] latency (ns): p50: 112 p90: 144 p99: 2575 p999: 3413 max: 921087
    [2019-01-19T13:38:24Z INFO  benchmark] total metrics pushed: 606955426

stjepang:
Thanks for the benchmarks and the writeup! <3

The ~20% improvement in the benchmarks is really nice. However, the PR in this current state is not yet a clear "slam dunk":

* Batched receives don't scale well under high contention.

* While extracting the batch using the `Extend` trait works, I'm unsure if that's the idiomatic style here. Some bikeshedding may be needed.

* This method is stretching the bounds of the already large API surface. It's a niche method and doesn't have a `try_send_many` counterpart, thus breaking the symmetry of `Sender` and `Receiver`. :)

Sorry if I'm being nitpicky, but I do appreciate all the effort you've put into this PR!

tobz:
I think this one can be closed at this point. :)

PR #276, opened 2018-12-23
Rustfix edition

jeehoonkang:
Just ran `cargo fix --edition` for all crates.


Vtec234:
LGTM, but it does make the older compilers fail. If we want to keep 1.26 support, this may need to be held off, unless there is some way to make them support Rust 2018.

PR #277, opened 2018-12-25
Introduce crossbeam-circbuf

jeehoonkang:
I'm proposing to introduce `crossbeam-circbuf`, which implements bounded/unbounded SPSC/SPMC queues based on circular buffer.  Bounded queues are implemented with fixed-sized circular buffer.  Unbounded queues are implemented with dynamically growable/shrinkable circular buffer.

Currently I'm not sure how performant this crate is.  A few benchmark will be helpful in evaluating the performance.  I'm thinking of re-purposing the benchmark used in `crossbeam-channel`.

The implementation is sub-optimal in that it has an unnecessary indirection when accessing the underlying buffer.  In order to remove the indireciton we need to [support DST](#209), which is not yet implemented in Crossbeam, though.

We discussed to introduce `crossbeam-circbuf` in [another PR](https://github.com/crossbeam-rs/crossbeam/pull/189), and there we also discussed tentative crate organizations.  Speaking of organizations, I think `crossbeam-deque` and `crossbeam-arrayqueue` can also be incorporated in this crate for the following reasons:

- All of them share [the same buffer struct](https://github.com/jeehoonkang/crossbeam/blob/circbuf/crossbeam-circbuf/src/buffer.rs).

- `crossbeam-arrayqueue` is actually bounded MPMC queue based on circular buffer.

- `crossbeam-deque` is actually a slight variant of unbounded SPMC queue based on circular buffer.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,10 @@ path = "./crossbeam-deque"
 default-features = false
 optional = true
 
+[dependencies.crossbeam-circbuf]
+version = "0.1.0"
+path = "./crossbeam-circbuf"
+
 [dependencies.crossbeam-epoch]
 version = "0.8"
 path = "./crossbeam-epoch"
@@ -85,5 +89,6 @@ members = [
   "crossbeam-epoch",
   "crossbeam-queue",
   "crossbeam-skiplist",
+  "crossbeam-circbuf",
   "crossbeam-utils",
 ]
--- a/crossbeam-circbuf/.travis.yml
+++ b/crossbeam-circbuf/.travis.yml
@@ -1,0 +1,12 @@
+language: rust
+
+rust:
+  - stable
+  - beta
+  - nightly
+
+script:
+  - cargo build
+  - cargo build --release
+  - cargo test
+  - cargo test --release
--- a/crossbeam-circbuf/Cargo.toml
+++ b/crossbeam-circbuf/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "crossbeam-circbuf"
+# When publishing a new version:
+# - Update CHANGELOG.md
+# - Update README.md
+# - Create "crossbeam-circbuf-X.Y.Z" git tag
+version = "0.1.0"
+authors = ["The Crossbeam Project Developers"]
+license = "MIT OR Apache-2.0"
+readme = "README.md"
+repository = "https://github.com/crossbeam-rs/crossbeam"
+homepage = "https://github.com/crossbeam-rs/crossbeam"
+documentation = "https://docs.rs/crossbeam-circbuf"
+description = "Concurrent queues based on circular buffer"
+keywords = ["circular-buffer", "ring-buffer", "spsc", "spmc", "concurrent-queue"]
+categories = ["concurrency", "data-structures"]
+
+[dependencies]
+crossbeam-epoch = { path = "../crossbeam-epoch" }
+crossbeam-utils = { path = "../crossbeam-utils" }
+memoffset = "0.5.4"
+
+[dev-dependencies]
+rand = "0.7"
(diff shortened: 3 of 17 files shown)


jeehoonkang:
It seems CI fails due to unrelated issue: https://travis-ci.org/crossbeam-rs/crossbeam/jobs/472150001#L455

bors retry

stjepang:
By extending the `crossbeam-channel` benchmarks, I get these results for `crossbeam-circbuf`:

```
bounded_mpmc              Rust crossbeam-circbuf   0.496 sec
bounded_mpsc              Rust crossbeam-circbuf   0.400 sec
bounded_seq               Rust crossbeam-circbuf   0.138 sec
bounded_spmc              Rust crossbeam-circbuf   0.233 sec
bounded_spsc              Rust crossbeam-circbuf   0.123 sec
```

And here is [MPMC queue](https://github.com/stjepang/queue):

```
unbounded_mpmc            Rust queue        0.179 sec
unbounded_mpsc            Rust queue        0.185 sec
unbounded_seq             Rust queue        0.242 sec
unbounded_spmc            Rust queue        0.204 sec
unbounded_spsc            Rust queue        0.112 sec
```

By bumping the block size from 32 to 256 we get even better results:

```
unbounded_mpmc            Rust queue        0.166 sec
unbounded_mpsc            Rust queue        0.190 sec
unbounded_seq             Rust queue        0.223 sec
unbounded_spmc            Rust queue        0.162 sec
unbounded_spsc            Rust queue        0.101 sec
```

And this is `crossbeam-deque`:

```
unbounded_seq             Rust crossbeam-deque   0.185 sec
unbounded_spmc            Rust crossbeam-deque   0.286 sec
unbounded_spsc            Rust crossbeam-deque   0.196 sec
```

My takeaways are:

* The bounded SPSC should be a very simple typical SPSC queue instead, not something based on Chase-Lev.

* Unbounded SPSC might be faster with something like [this](http://www.1024cores.net/home/lock-free-algorithms/queues/unbounded-spsc-queue), and wouldn't need to use `crossbeam-epoch`. We could try allocating nodes in blocks to squeeze out more performance from it.

* The MPMC queue seems to be faster in the MPMC and MPSC cases.

* The case where `crossbeam-circuf` excels is SPMC, which is unsurprising because that's pretty much the use case Chase-Lev was invented for. But even then the MPMC queue is competitive (maybe because it doesn't need `crossbeam-epoch`).

* Currently, `crossbeam-circbuf` is a clear winner only in the "seq" (single thread) case, which is uninteresting for a concurrent queue anyway. :( I wonder if we could optimize things some more to find its niche.

I should also add that it's possible that these benchmarks are not measuring the right thing and might be unfair to `crossbeam-deque`. If you have a suggestion on how to do benchmarks differently, do say!

jeehoonkang:
In my unscientific experiment, it seems the numbers will become much better if we could remove an indirect pointer access using #209.  It'll be roughly on par with #338 for the bounded spsc case.

stjepang:
@jeehoonkang

Do you have any numbers or perhaps benchmarks I could run to reproduce them?

In my [first comment](https://github.com/crossbeam-rs/crossbeam/pull/277#issuecomment-451650150) on this PR, I noted it takes 0.123 sec for the bounded SPSC variant of `crossbeam-circbuf` to finish the `spsc` benchmark. 5 million messages are sent so that means sending & receiving a single message takes 25 ns.

The [new bounded SPSC](https://github.com/crossbeam-rs/crossbeam/pull/338#issue-257249385) queue spends 0.029 sec on that same benchmark, which is 6 ns per message. That's 4x faster than `crossbeam-circbuf`.

I honestly doubt the double indirection costs much, if anything (my guess is less than 1 ns per message). The new bounded SPSC channel uses double indirection too, so it definitely cannot cost more than 6 ns per message.

Even if we conservatively assume the cost of double indirection is the whole 6 ns per message, that would mean removing it brings the result for `crossbeam-circbuf` from 25 ns down to 19 ns, which is still 3x slower than the new queue.

jeehoonkang:
@stjepang I'll soon provide benchmark numbers.  Sorry for delay.

PR #279, opened 2018-12-28
Optimize unbounded channels

stjepang:
Use a different queue for unbounded channels.

This gives us:

* Performance improvements (see benchmarks below).
* Lower memory consumption (memory reclamation is not deferred, it's *eager*).
* Fewer dependencies (no more `crossbeam-epoch`).

Before:

```
unbounded_mpmc            Rust crossbeam-channel   0.392 sec
unbounded_mpsc            Rust crossbeam-channel   0.373 sec
unbounded_select_both     Rust crossbeam-channel   0.536 sec
unbounded_select_rx       Rust crossbeam-channel   0.589 sec
unbounded_seq             Rust crossbeam-channel   0.462 sec
unbounded_spsc            Rust crossbeam-channel   0.235 sec
```

After:

```
unbounded_mpmc            Rust crossbeam-channel   0.266 sec
unbounded_mpsc            Rust crossbeam-channel   0.250 sec
unbounded_select_both     Rust crossbeam-channel   0.449 sec
unbounded_select_rx       Rust crossbeam-channel   0.438 sec
unbounded_seq             Rust crossbeam-channel   0.333 sec
unbounded_spsc            Rust crossbeam-channel   0.210 sec
```

### DIFF
--- a/crossbeam-channel/Cargo.toml
+++ b/crossbeam-channel/Cargo.toml
@@ -20,10 +20,6 @@ parking_lot = "0.7"
 rand = "0.6"
 smallvec = "0.6.2"
 
-[dependencies.crossbeam-epoch]
-version = "0.7"
-path = "../crossbeam-epoch"
-
 [dependencies.crossbeam-utils]
 version = "0.6.3"
 path = "../crossbeam-utils"
--- a/crossbeam-channel/src/context.rs
+++ b/crossbeam-channel/src/context.rs
@@ -94,7 +94,8 @@ impl Context {
                 select.into(),
                 Ordering::AcqRel,
                 Ordering::Acquire,
-            ).map(|_| ())
+            )
+            .map(|_| ())
             .map_err(|e| e.into())
     }
 
--- a/crossbeam-channel/src/flavors/list.rs
+++ b/crossbeam-channel/src/flavors/list.rs
@@ -4,10 +4,9 @@ use std::cell::UnsafeCell;
 use std::marker::PhantomData;
 use std::mem::{self, ManuallyDrop};
 use std::ptr;
-use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+use std::sync::atomic::{self, AtomicPtr, AtomicUsize, Ordering};
 use std::time::Instant;
 
-use crossbeam_epoch::{self as epoch, Atomic, Guard, Owned, Shared};
 use crossbeam_utils::CachePadded;
 
 use context::Context;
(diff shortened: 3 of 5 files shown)


stjepang:
@jeehoonkang Here's a reference implementation: https://github.com/stjepang/queue

This unbounded MPMC queue is very similar to Dmitry Vyukov's [bounded MPMC](http://www.1024cores.net/home/lock-free-algorithms/queues/bounded-mpmc-queue) - it's the same idea, except we have a linked list of blocks rather than one big circular array.

This queue is not 100% lock-free (but neither is `SegQueue`) - there are a few locking sections of code, but they're very small so chances of hitting them are slim. They don't seem to have a big effect on scalability, throughput, nor latency, which is good.

In my benchmarks (those in `crossbeam-channel/benchmarks`), this queue seems to be faster than `MsQueue`, `SegQueue`, and `crossbeam-deque`. I'll benchmark with `crossbeam-circbuf`, too - should be interesting. Eliminating `crossbeam-epoch` is a noticeable perfomance win because pinning incurs one TLS access and one `SeqCst` fence per operation. Tokio benchmarks show better results with this queue than with `SegQueue` and `crossbeam-channel`, too. Finally, latency distribution of operations on this queue is better than with other queues. 

Another benefit of this queue is that memory reclamation is not deferred at all - it's fully eager and there is no concept of garbage! As soon as the last reference to a block is dropped, it gets destroyed. I did some tests and measured lower memory overhead in high concurrency scenarios than with queues using epochs. Hopefully this fixes [the problem](https://github.com/firecracker-microvm/firecracker/pull/677#pullrequestreview-179258729) where Firecracker tests were failing on Amazon's internal CI due to excessive memory use by `crossbeam-channel`.

So how does this avoid epochs? The queue has head and tail indices of type `AtomicUsize`, and there are also head and tail pointers to blocks of type `AtomicPtr<Block<T>>`. Each block has a slot where a message can be stored and each slot has a state with three bits: `READ`, `WRITE`, and `DESTROY`. All bits are initially zero.

To send a message, load `tail.index` and `tail.block`, then do CAS to advance `tail.index` forward. If we succeeded, now we "own" the slot at that index in that block. Now it's safe to dereference the block pointer and write the message into the slot. When done, we do a `fetch_or` to set the `WRITE` bit in the slot's state.

To receive a message, similarly load `head.index` and `head.block`, then do CAS to advance `head.index` forward. If we succeeded, now we "own" the slot at that index in that block. Now it's safe to dereference the block pointer and read the message from the slot (but first wait until `WRITE` is set). When done, we do a `fetch_or` to set the `READ` bit in the slot's state. If we notice that the `DESTROY` bit is set too, we must call `destroy(block)`.

Here comes block destruction. If our receive operation got the last slot in the block, after reading the message and setting the `READ` bit, we must call `destroy(block)`. The `destroy` function iterates over slots in the block and sets the `DESTROY` bit in each slot's state. If we notice during `fetch_or` that the `READ` bit is set too, that means nobody is using the slot anymore. However, if the `READ` bit is not set, that means some other thread is still using the slot, but now that thread became responsible for destroying the block so we just return.

That's the gist of it. I'm omitting a few less important remaining details, but hopefully they make sense when reading the code.

I'm very excited about this queue. Although it doesn't seem super interesting from the CS perspective, in practice it seems to be outperforming pretty much everything else as a general-purpose unbounded MPMC queue.

jeehoonkang:
Your queue seems very interesting.  I'll look into it real soon.

---

> Although it doesn't seem super interesting from the CS perspective,

AFAICT, everything practical is interesting form the CS perspective :)  To [quote Bertrand Meyer](https://arxiv.org/abs/1712.05078):

> The outcomes that matter in research are not numerous publications, best-paper awards, completed PhD theses, keynote invitations, software tools, citations and other measurable signs of progress. I was after real success, in the sense of changing the way the IT industry develops software.

stjepang:
I haven't tried with that many cores yet, let's do this!

So I'm thinking about writing a comprehensive benchmark suite for queues similar to `crossbeam-channel/benchmarks`, except better. The benchmarks would be organized as follows.

There are scenarios with non-blocking operations:
*  sequential (send N messages, receive N messages)
* mp (send N messages with T threads)
* mc (receive N messages with T threads)

There are scenarios with blocking operations (we spin+yield when the queue is full or empty):
* spsc (send and receive N messages with 2 threads)
* spmc (send and receive N messages with T+1 threads)
* mpsc (send and receive N messages with T+1 threads)
* mpmc (send and receive N messages with 2T threads)

We measure:
* Total time to complete all operations - that's throughput.
* The time to complete each operation. Then sort the times and print percentiles: 50, 90, 99, 999 - that's latency.

We do this for a bunch of different Ts (number of threads), ranging from 1 to 100 or so.

A few notes about bounded queues:
* In scenarios with non-blocking operations, we set capacity to N so that the queue never gets full.
* In scenarios with blocking operations, we set capacity to 1, 10, 100, and N.

How does this plan sound? What do benchmarks in published papers on concurrent queues do? Should we do something differently?

PR #281, opened 2018-12-30
channel: update benchmarks

### LINKED ISSUE #231: Update Go in channel benchmark
It seems `crossbeam-channel`'s benchmark result is from Go 1.10:
https://github.com/crossbeam-rs/crossbeam/tree/master/crossbeam-channel/benchmarks

The newest version is 1.11, released on 24 Aug 2018.  How about updating the version in the benchmark result?

stjepang:
Closes #231


Thomasdezeeuw:
This also contains a bug fix, does that have a test somewhere?

stjepang:
@Thomasdezeeuw Good idea - I've added a test in https://github.com/crossbeam-rs/crossbeam/pull/284

PR #283, opened 2018-12-30
Reduce the size of Receiver

stjepang:
This changes the size of `Receiver`s from 32 bytes to 16 bytes and the size of `Sender` from 8 bytes to 16 bytes.

The amount of reference counting going on is reduced too, which makes channel construction and destruction slightly faster.


stjepang:
Ah, not a bug in the actual code - it's a bug in a test.

bors r+

PR #299, opened 2019-01-22
utils: add Backoff for exponential backoff

stjepang:
This is one of those simple tools like `CachePadded` that comes up so often it is useful to have it in the library, despite being easy to implement. `Backoff` calls `spin_loop_hint()` and `yield_now()` with good default settings that work well across a wide spectrum of workloads.

Currently, `crossbeam-channel`, `crossbeam-deque`, and `SegQueue` have their own `Backoff`s so this PR just removes duplication.

There's even a place in Tokio where this might be useful as we previously struggled with tuning backoff manually and got it wrong: https://github.com/tokio-rs/tokio/pull/534

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ This crate provides a set of tools for concurrent programming:
     * `epoch` module contains epoch-based garbage collection.
 
 * Utilities
+    * `Backoff` performs exponential backoff in spin loops.
     * `CachePadded<T>` pads and aligns a value to the length of a cache line.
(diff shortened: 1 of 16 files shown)


stjepang:
@Vtec234 It may make `Backoff` easier to use in certain contexts. I don't have a great example at the moment, but imagine some code structured roughly like this:

```rust
struct Operation {
    backoff: Backoff,
    // ...
}

impl Operation {
    fn try(&self) -> bool {
        loop {
            // ...
            self.backoff.spin();
        }
    }
}
```

PR #300, opened 2019-01-23
Use `target_has_atomic` in `impl_arithmethic`

jrobsonchase:
The lack of larger atomics on thumbv7m-none-eabi (and I'm guessing
similar embedded platforms as well) was causing build errors.


stjepang:
Thanks for the PR! Travis is failing because literals in macros are only supported on very recent versions of Rust. Is there a workaround for this?

jrobsonchase:
Ah, should probably work with just a `tt`. I'll give that a shot.

stjepang:
Nice!
bors r+

PR #308, opened 2019-01-27
Support for 128-bit integers in AtomicCell

stjepang:


stjepang:
Yes, that is correct. Locks will be used, but these metods are still useful.

PR #316, opened 2019-01-30
Fix deque::Injector doc link

alexheretic:


stjepang:
Ah, copy/paste errors. Thanks for the fix! :)

bors r+

PR #320, opened 2019-02-05
Implement PartialEq and Eq for channels 

### LINKED ISSUE #319: Implement PartialEq + Eq for channel
As the title says, would it possible/acceptable to implement `PartialEq` and `Eq` for the channel types?

I'm wrapping a `Sender` in another type which I would like to implement `Eq` for. If approved I'll take a look at implementing it.

Thomasdezeeuw:
Currently for the never receiver always returns false as it is not
possible to determine if two receivers are equal using the current
structure.

I've added a test to compare never receivers, but it is disabled as it
currently doesn't pass.

Closes #319.


Thomasdezeeuw:
Only thing I would like to add is test for `same_channel` with different flavours of channels but I don't know in which file to put that.

Thomasdezeeuw:
@stjepang I've rebased on master and ran rustfmt, the CI should pass now.

crossbeam-channel/src/channel.rs
```diff
@@ -531,6 +531,34 @@ impl<T> Sender<T> {
             SenderFlavor::Zero(chan) => chan.capacity(),
         }
     }
+
+    /// Returns true if senders send to the same channel.
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # fn main() {
```
stjepang: I believe `# fn main() {` is not required in these examples.

stjepang:
> Only thing I would like to add is test for `same_channel` with different flavours of channels but I don't know in which file to put that.

I'm okay with adding a new test file (e.g. `tests/same_channel.rs`) to test this function.

stjepang:
@BurntSushi @glaebhoerl How do you feel about this PR? Are you fine with adding `Sender::same_channel()` and `Receiver::same_channel()`?

Thomasdezeeuw:
I've move the tests to there own file and added a test for `same_channel` with different flavours. Also removed main from the examples

glaebhoerl:
@stjepang I don't have any "dog in this hunt" in that I'm not currently (or in the foreseeable future) using this for anything, so I'm quite fine with any solution. :) FWIW, one point I don't remember being brought up is that an `impl Eq`/`PartialEq` makes it easier to compose with any algorithms or data structures that have an `Eq`/`PartialEq` requirement. (I don't know how many of those there are which don't also require `Ord`/`PartialOrd` or `Hash` though, and it's just a newtype away even with the `same_channel` methods.)

Thomasdezeeuw:
Its been two weeks any more comments on this PR?

stjepang:
Ah apologies - looks like this PR fell off my radar :(

@Thomasdezeeuw This PR looks ready for merging to me. Could you just run rustfmt one last time to make the CI pass?

stjepang:
One last bikeshedding round - what do you think about the following naming scheme?

```rust
impl<T> Sender<T> {
    fn identical_to(&self, s: &Sender<T>) -> bool;
    fn connected_to(&self, r: &Receiver<T>) -> bool;
}

impl<T> Receiver<T> {
    fn identical_to(&self, r: &Receiver<T>) -> bool;
    fn connected_to(&self, s: &Sender<T>) -> bool;
}
```

(this PR is not introducing `connected_to`, but we could add such a function in the future)

Thomasdezeeuw:
I've rebased on master and ran rustfmt. I'm ok with `identical_to` as name too, but that sounds somewhat the same as `eq` (from `PartialEq` and `Eq`) to me.

stjepang:
> but that sounds somewhat the same as `eq` (from `PartialEq` and `Eq`) to me.

I was going for structural vs identity equality (`==` is structural, `identical_to` is identity) :)

@BurntSushi Your thoughts on this?

BurntSushi:
@stjepang I like your suggestion. Making the distinction between identity and structural more explicit in the docs would be good too.

Thomasdezeeuw:
So do we agree on `identical_to`? Personally I'm ok with it.

stjepang:
@Thomasdezeeuw Yeah let's do that!

Thomasdezeeuw:
@stjepang done. Ready for a final review.

Issue #321, opened 2019-02-05
Various test failures on macOS for the channel crate


Thomasdezeeuw:
The following tests fail on my macOS machine.

```
use_while_exiting
channel_tests::oneshot_single_thread_recv_chan_close
channel_tests::oneshot_multi_task_recv_then_close
channel_tests::oneshot_multi_thread_send_close_stress
channel_tests::oneshot_multi_thread_recv_close_stress
sync_channel_tests::oneshot_single_thread_recv_chan_close
sync_channel_tests::oneshot_multi_task_recv_then_close
sync_channel_tests::oneshot_multi_thread_send_close_stress
sync_channel_tests::oneshot_multi_thread_recv_close_stress
```

I've added some stacktraces, I'll look into a bit further myself.

`use_while_exiting`

```
thread '<unnamed>' panicked at 'use of std::thread::current() is not possible after the thread's local data has been destroyed', src/libcore/option.rs:1038:5
stack backtrace:
   0:        0x103be2c33 - std::sys::unix::backtrace::tracing::imp::unwind_backtrace::h6b67c23e58603609
                               at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:39
   1:        0x103bdeee2 - std::sys_common::backtrace::_print::hd7c2b572e0c4e656
                               at src/libstd/sys_common/backtrace.rs:70
   2:        0x103be1ae6 - std::panicking::default_hook::{{closure}}::hbfd2b812b74e7516
                               at src/libstd/sys_common/backtrace.rs:58
                               at src/libstd/panicking.rs:200
   3:        0x103be17f1 - std::panicking::default_hook::h2e0df5c26761382f
                               at src/libstd/panicking.rs:215
   4:        0x103be226e - <std::panicking::begin_panic::PanicPayload<A> as core::panic::BoxMeUp>::get::h68f156da32bb6727
                               at src/libstd/panicking.rs:478
   5:        0x103be1d8c - std::panicking::continue_panic_fmt::he9962abdc5d06fe1
                               at src/libstd/panicking.rs:385
   6:        0x103be1c78 - std::panicking::try::do_call::h62d8346540aa00a7
                               at src/libstd/panicking.rs:312
   7:        0x103bf3b51 - core::char::methods::<impl char>::escape_debug::hb08634abedefbc98
                               at src/libcore/panicking.rs:85
   8:        0x103bf3bbd - core::char::methods::<impl char>::escape_debug::hb08634abedefbc98
[...]

PR #322, opened 2019-02-05
Join all threads created in tests

Thomasdezeeuw:
A lot of tests created threads but never waited (joined) on them. As a
result if one of those threads panicked they test still passed and the
test runner exits with a success exit code.

Updates #321.


Thomasdezeeuw:
@stjepang done.

Thomasdezeeuw:
The failure should be fixed in #323.

PR #323, opened 2019-02-05
 Cleanup panicking test threads

Thomasdezeeuw:
Depends on #322.

Updates  #321.

### DIFF
--- a/crossbeam-channel/tests/mpsc.rs
+++ b/crossbeam-channel/tests/mpsc.rs
@@ -23,6 +23,7 @@ extern crate crossbeam_channel as cc;
 
 use std::sync::mpsc::{RecvError, RecvTimeoutError, TryRecvError};
 use std::sync::mpsc::{SendError, TrySendError};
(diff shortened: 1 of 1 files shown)


Thomasdezeeuw:
There seems to be a race condition in the `oneshot_multi_thread_send_close_stress` test. In a loop it creates two threads, on one thread it drops the receiver and on the other thread it send a value, which means that sending sometimes succeeds and sometimes fails. @stjepang any ideas on what would be the best way forward?

stjepang:
Looks like that test makes is nonsensical. I'd say let's just pull `tx.send(1)` out of `assert_eq!(tx.send(1), Err(SendError(1)))` - so remove the assert but keep the send operation.

And add a comment on top of the file saying "Remove assertion from `oneshot_multi_thread_send_close_stress`."

Thomasdezeeuw:
Fixed the race condition and added the change to the list.

stjepang:
lgtm, thanks!! :)

bors r+

PR #325, opened 2019-02-06
Enable testing on macOS (OS X) on Travis CI

Thomasdezeeuw:
Note that macOS on Travis is not the fastest thing in the world and as crossbeam already has a huge number of tests still will greatly increase the time it takes for the check to complete.

Updates #321.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,11 @@ matrix:
   - rust: 1.26.0
     name: "crossbeam-utils on 1.26.0"
     script: ./ci/crossbeam-utils.sh
+  # Run all test with the highest minimum Rust version on macOS.
+  - rust: 1.28.0
+    os: osx
+    name: "all of crossbeam on 1.28.0 on macOS"
+    script: ./ci/all.sh
 
   # Test crates on nightly Rust.
   - rust: nightly
@@ -68,6 +73,11 @@ matrix:
   - rust: nightly
     name: "crossbeam-utils on nightly"
     script: ./ci/crossbeam-utils.sh
+  # Run all test on macOS.
+  - rust: nightly
+    os: osx
+    name: "all of crossbeam on nightly on macOS"
+    script: ./ci/all.sh
 
   # Check for duplicate dependencies.
   - rust: stable
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ matrix:
   - rust: 1.28.0
     os: osx
     name: "all of crossbeam on 1.28.0 on macOS"
-    script: ./ci/all.sh
+    script: ./ci/all.sh --test-threads=1
 
   # Test crates on nightly Rust.
   - rust: nightly
@@ -77,7 +77,7 @@ matrix:
   - rust: nightly
     os: osx
     name: "all of crossbeam on nightly on macOS"
-    script: ./ci/all.sh
+    script: ./ci/all.sh --test-threads=1
 
   # Check for duplicate dependencies.
   - rust: stable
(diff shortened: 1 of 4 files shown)


Thomasdezeeuw:
Hmmm, this didn't work as I though it would. I think I have to duplicate the entire matrix and set the OS per entry, which seems a bit much to me. It might be fine to just run all the tests with a single compiler version, i.e. minimal or nightly. @stjepang any thoughts?

stjepang:
Sounds good! We can add a single item to the matrix that runs `./ci/all.sh` on macOS with stable Rust. The `all.sh` script could look like this (same as `crossbeam.sh` except uses `--all` in all `cargo` invocations):

```bash
#!/bin/bash

cd "$(dirname "$0")"/..
set -ex

export RUSTFLAGS="-D warnings"

cargo check --all --no-default-features
cargo check --all --bins --examples --tests
cargo test --all

if [[ "$TRAVIS_RUST_VERSION" == "nightly" ]]; then
    cargo check --all --no-default-features --features nightly
    cargo test --all --features nightly
fi
```

Thomasdezeeuw:
@stjepang added the script and rebased on master.

stjepang:
Looks like `all.bash` should've been named `all.sh` :)

Thomasdezeeuw:
Whoops, fixed now :)

stjepang:
All right, the tests should pass now...

bors r+

Thomasdezeeuw:
I don't know about the bors' errors, but for Travis one failure is edition related and the other one is timeout related.

`atomicring` v1.1.2 seems to make use of edition 2018 (see https://github.com/eun-ice/atomicring/blob/b0cf3b0c8e0ca2df8cf578c4484dd4d2255d5ec5/Cargo.toml), which requires Rust 1.31. I don't know how the Linux build is able to proceed. Maybe its using an old version of atomicring? It seems v1.0.4 doesn't use 2018 edition (see https://github.com/eun-ice/atomicring/blob/dbe8b89a929de0808b5913a1db5cd911777770c0/Cargo.toml) and it was changed in v1.0.5 (see https://github.com/eun-ice/atomicring/commit/d1c174305133bd5c435a6cbe51bdab1c7b839862#diff-80398c5faae3c069e4e6aa2ed11b28c0).

Should I increase the minimum Rust version to 1.31 for the stable build?

I've seen the timeout failures before, my solution was to skip all tests with (small) timeouts. I don't know the exact problem but it seems the macOS machines Travis uses are either very slow, or shared between too many builds. Some of my prs fixing it in another project: https://github.com/Thomasdezeeuw/mio-st/pull/3, https://github.com/Thomasdezeeuw/mio-st/pull/36.

@stjepang What would your preferred way to fix this be?

stjepang:
Regarding the `atomicring` compilation error: I think it's best to just delete the `"crossbeam-channel/benchmarks"` line from `Cargo.toml` in order to remove benchmarks from the workspace. Benchmarks are still checked by CI on nightly Rust (see `ci/crossbeam-channel.sh`) so that's okay.

About timeouts... ugh, that's a tough one. :( Let's try adding `-- --test-threads=1` to the failing `cargo test` command and see if that helps. If not, then I'm afraid the Travis macOS environment is just too unreliable for these kinds of tests. I tried enabling macOS tests once before and ran into the same problem.

Thomasdezeeuw:
@stjepang I've removed the benchmarks from the workspace and set the test threads to 1. Unfortunately the timeouts are still to strict, I'll push some commits to increase the margins but that will take some time as Travis first has to fail on each one.

Thomasdezeeuw:
@stjepang Travis-ci macOS boxes are just too slow, I'm going to close this.

PR #329, opened 2019-02-15
Use acquire-release orderings in AtomicCell

### LINKED ISSUE #317: AtomicCell: Why do you use SeqCst?
This [recent great blogpost](https://stjepang.github.io/2019/01/29/lock-free-rust-crossbeam-in-2019.html) said that `AtomicCell` doesn't let the user choose the memory ordering, which is fair, but of course I had to immediately check which one it uses ;) .  I was very surprised to see it use `SeqCst`.  I would have expected `Acq`/`Rel`/`AcqRel`.  This allows reasoning by ownership transfer, which is strong enough for almost everything.

I wouldn't usually care enough to actually submit an issue about this, but now I found [this comment](https://github.com/crossbeam-rs/crossbeam-utils/pull/39#discussion_r206885280):

> So AtomicCell<T> is basically Mutex<Cell<T>>, except faster. That's it.

This is not in the documentation as far as I can see, but I agree it makes sense as a principle.  The thing is, from all I know it is perfectly legal to implement locks with release-acquire spinlocks, so by this argument, the accesses in `AtomicCell` should definitely be release-acquire!

The post claims that
```rust
let x = AtomicCell::new(false);
let y = AtomicCell::new(false);
let z = AtomicCell::new(0usize);

scope(|s| {
    s.spawn(|| x.store(true));
    s.spawn(|| y.store(true));

    s.spawn(|| {
        while !x.load() {}
        if y.load() { z.fetch_add(1); }
    });

    s.spawn(|| {
        while !y.load() {}
        if x.load() { z.fetch_add(1); }
    });
}).unwrap();

assert!(z.load() > 0);
```
will pass when using `Mutex<Cell>`, but (assuming this was tested) I think that is just an accident of the current implementation.  Nothing in the C11 memory model requires this to pass when using locks, and you absolutely shouldn't rely on it.

Cc @jeehoonkang @stjepang

stjepang:
Closes #317

### DIFF
--- a/crossbeam-utils/src/atomic/atomic_cell.rs
+++ b/crossbeam-utils/src/atomic/atomic_cell.rs
@@ -15,8 +15,12 @@ use Backoff;
 /// global locks otherwise. You can call [`AtomicCell::<T>::is_lock_free()`] to check whether
 /// atomic instructions or locks will be used.
 ///
+/// Atomic loads use the [`Acquire`] ordering and atomic stores use the [`Release`] ordering.
+///
 /// [`Cell`]: https://doc.rust-lang.org/std/cell/struct.Cell.html
 /// [`AtomicCell::<T>::is_lock_free()`]: struct.AtomicCell.html#method.is_lock_free
+/// [`Acquire`]: https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.Acquire
+/// [`Release`]: https://doc.rust-lang.org/std/sync/atomic/enum.Ordering.html#variant.Release
 pub struct AtomicCell<T> {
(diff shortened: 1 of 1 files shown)


Amanieu:
Do we need to change `load` and `store` to use `AcqRel` ordering as well (to match the semantics of `Mutex<T>`)? This can be done by using `fetch_add(0, AcqRel)` and `swap(x, AcqRel)`, however we would lose a lot in performance.

stjepang:
@Amanieu Do you think we could use `SeqCst` fences instead of `fetch_add`/`swap` to achieve the same effect?

PR #331, opened 2019-02-15
CachePadded: use 128-byte alignment on x86-64

stjepang:
Motivation: https://www.reddit.com/r/rust/comments/akzas0/lockfree_rust_crossbeam_in_2019/efa26r1/

### DIFF
--- a/crossbeam-utils/src/cache_padded.rs
+++ b/crossbeam-utils/src/cache_padded.rs
@@ -9,14 +9,17 @@ use core::ops::{Deref, DerefMut};
 /// CPU cores. Use `CachePadded` to ensure updating one piece of data doesn't invalidate other
 /// cached data.
 ///
-/// Cache lines are assumed to be 64 bytes on all architectures.
-///
 /// # Size and alignment
 ///
-/// The size of `CachePadded<T>` is the smallest multiple of 64 bytes large enough to accommodate
+/// Cache lines are assumed to be of N bytes, depending on the architecture:
+///
+/// * On x86-64, N = 128.
+/// * On all others, N = 64.
+///
+/// The size of `CachePadded<T>` is the smallest multiple of N bytes large enough to accommodate
 /// a value of type `T`.
 ///
-/// The alignment of `CachePadded<T>` is the maximum of 64 bytes and the alignment of `T`.
+/// The alignment of `CachePadded<T>` is the maximum of N bytes and the alignment of `T`.
 ///
 /// # Examples
 ///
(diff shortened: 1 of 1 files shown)


crossbeam-utils/src/cache_padded.rs
```diff
@@ -9,14 +9,17 @@ use core::ops::{Deref, DerefMut};
...
 /// cached data.
 ///
-/// Cache lines are assumed to be 64 bytes on all architectures.
-///
 /// # Size and alignment
 ///
-/// The size of `CachePadded<T>` is the smallest multiple of 64 bytes large enough to accommodate
+/// Cache lines are assumed to be of N bytes, depending on the architecture:
+///
+/// * On x86-64, N = 128.
```
Vtec234: Isn't it more the case that x64 CPUs fetch two cache lines at a time to align them to 128-bit chunks rather than that a single cache line is 128-bit?

crossbeam-utils/src/cache_padded.rs
```diff
@@ -49,7 +52,14 @@ use core::ops::{Deref, DerefMut};
 /// }
 /// ```
 #[derive(Clone, Copy, Default, Hash, PartialEq, Eq)]
-#[repr(align(64))]
+// Starting from Intel's Sandy Bridge, spatial prefetcher is now pulling pairs of 64-byte cache
```
Vtec234: Is this also the case for AMD?

stjepang: Judging by a quick skim through AMD's optimization manual, I'd say no.
https://developer.amd.com/wordpress/media/2013/12/55723_SOG_Fam_17h_Processors_3.00.pdf

crossbeam-utils/src/cache_padded.rs
stjepang: Yes. Maybe the wording is not great, but `N` is just a pessimistic *guess* of how long a cache line is. So `N` is at least the length of a cache line, but may be larger than that in order to accommodate architectures with longer cache lines or ones that have unusual prefetching strategies.

Vtec234: Okay, perhaps this could be reworded slightly to avoid confusion later? In any case LGTM.

stjepang: Sure - just added an additional comment for clarification.

crossbeam-utils/src/cache_padded.rs
taiki-e: @Mart-Bogdan I would accept a PR to adjust the placement of CachePadded based on the CPU when the user specifies -C target-cpu in rustflag.

PR #334, opened 2019-02-21
Rustfmt everything

stjepang:


Vtec234:
Thoughts on taking this opportunity to add `rustfmt --check` to Travis as a lint that must pass? That way we can avoid 'Rusfmt everything' to be the latest commit for every file :P

stjepang:
@Vtec234 Sure, let's give it a shot!

Vtec234:
`rustfmt` could have failed due to missing `cargo fmt`, since it's not installed by default by `rustup`.

stjepang:
Thanks for fixing this! :)
bors r+

PR #337, opened 2019-02-27
Release `RefEntry` when dropping `Entry`

### LINKED ISSUE #273: Memory leak or gc not triggered?
Hello, i'm not sure if there's a bug in code, or whether the epoch-based gc is not triggerly frequently enough, but my friend says this code can easily run out of system memory.

```rust
extern crate crossbeam_skiplist;
use crossbeam_skiplist::SkipMap;

struct TestDrop{
    i:i64
}
impl Drop for TestDrop {
    fn drop(&mut self) {
        println!("TestDrop Closing port, {}", self.i);
    }
}

fn main() {
    let m:SkipMap<String, TestDrop> = SkipMap::new();
    loop{
        m.clear();
        //for entry in m.iter(){
        //    entry.remove();
        //}
        for i in 0 .. 5000000{
            m.insert(format!("{}", i), TestDrop{i:i});
        }
        println!("map.len():{:?}", m.len());
    }
}
```

topecongiro:
This PR fixes #273. ~in the case where you are using `clear`. The memory leak still persists when you use `Entry::remove()`~. EDIT: both cases are fixed with https://github.com/crossbeam-rs/crossbeam/commit/a6de0ca23b5364d2acc876f6e3d74cb8ab4d7b4a.

I find that when creating a new `Entry`, a reference counter to the node gets incremented, but the counter won't get decremented when `Entry` gets dropped. This could be why garbage were not collected, though I am not entirely sure if it is right. Looking forward to see reviews from the crossbeam maintainers.

Thank you for creating great piece of code and documentation. It was very fun to read both of them.

### DIFF
--- a/crossbeam-skiplist/src/base.rs
+++ b/crossbeam-skiplist/src/base.rs
@@ -230,7 +230,7 @@ impl<K, V> Node<K, V> {
     /// Decrements the reference count of a node, pinning the thread and destoying the node
     /// if the count become zero.
     #[inline]
-    unsafe fn decremnt_with_pin<F>(&self, pin: F)
+    unsafe fn decrement_with_pin<F>(&self, pin: F)
     where
         F: FnOnce() -> Guard,
     {
@@ -1431,7 +1431,7 @@ impl<'a, K: 'a, V: 'a> RefEntry<'a, K, V> {
     where
         F: FnOnce() -> Guard,
     {
-        unsafe { self.node.decremnt_with_pin(pin) }
+        unsafe { self.node.decrement_with_pin(pin) }
     }
 
     /// Tries to create a new `RefEntry` by incrementing the reference count of
--- a/crossbeam-skiplist/src/base.rs
+++ b/crossbeam-skiplist/src/base.rs
@@ -1401,6 +1401,12 @@ impl<'a, K: 'a, V: 'a> RefEntry<'a, K, V> {
         self.parent
     }
 
+    /// Releases the reference on the entry.
+    pub fn release(self, guard: &Guard) {
+        self.parent.check_guard(guard);
+        unsafe { self.node.decrement(guard) }
+    }
+
     /// Tries to create a new `RefEntry` by incrementing the reference count of
     /// a node.
     unsafe fn try_acquire(
@@ -1529,12 +1535,6 @@ where
                 }
             }
         }
-    }
-
-    /// Releases the reference on the entry.
-    pub fn release(self, guard: &Guard) {
-        self.parent.check_guard(guard);
-        unsafe { self.node.decrement(guard) }
     }
 }
 
--- a/crossbeam-skiplist/src/base.rs
+++ b/crossbeam-skiplist/src/base.rs
@@ -230,7 +230,7 @@ impl<K, V> Node<K, V> {
     /// Decrements the reference count of a node, pinning the thread and destoying the node
     /// if the count become zero.
     #[inline]
-    unsafe fn decrement_with_pin<F>(&self, pin: F)
+    unsafe fn decrement_with_pin<F>(&self, parent: &SkipList<K, V>, pin: F)
     where
         F: FnOnce() -> Guard,
     {
@@ -241,7 +241,9 @@ impl<K, V> Node<K, V> {
             == 1
         {
             fence(Ordering::Acquire);
-            pin().defer_unchecked(move || Self::finalize(self));
+            let guard = &pin();
+            parent.check_guard(guard);
+            guard.defer_unchecked(move || Self::finalize(self));
         }
     }
 
@@ -1431,7 +1433,7 @@ impl<'a, K: 'a, V: 'a> RefEntry<'a, K, V> {
     where
         F: FnOnce() -> Guard,
     {
-        unsafe { self.node.decrement_with_pin(pin) }
+        unsafe { self.node.decrement_with_pin(self.parent, pin) }
     }
 
     /// Tries to create a new `RefEntry` by incrementing the reference count of
--- a/crossbeam-skiplist/src/base.rs
+++ b/crossbeam-skiplist/src/base.rs
@@ -227,6 +227,24 @@ impl<K, V> Node<K, V> {
         }
     }
 
+    /// Decrements the reference count of a node, pinning the thread and destoying the node
+    /// if the count become zero.
+    #[inline]
+    unsafe fn decremnt_with_pin<F>(&self, pin: F)
+    where
+        F: FnOnce() -> Guard,
+    {
+        if self
+            .refs_and_height
+            .fetch_sub(1 << HEIGHT_BITS, Ordering::Release)
+            >> HEIGHT_BITS
+            == 1
+        {
+            fence(Ordering::Acquire);
+            pin().defer_unchecked(move || Self::finalize(self));
+        }
+    }
+
     /// Drops the key and value of a node, then deallocates it.
     #[cold]
     unsafe fn finalize(ptr: *const Self) {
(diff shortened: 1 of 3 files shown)


crossbeam-skiplist/src/base.rs
```diff
@@ -1401,6 +1401,12 @@ impl<'a, K: 'a, V: 'a> RefEntry<'a, K, V> {
         self.parent
     }
 
+    /// Releases the reference on the entry.
+    pub fn release(self, guard: &Guard) {
+        self.parent.check_guard(guard);
+        unsafe { self.node.decrement(guard) }
+    }
```
stjepang: Was this function modified in any way or just moved?

topecongiro: It was just moved so that it can be called from `drop` without any trait bound.

topecongiro:
> Shouldn't we instead implement `Drop` for `base::RefEntry` rather than `map::Entry`? :)

I though that implementing `Drop` for `base::RefEntry` will be inconsistent with existing APIs. Currently interfaces of `RefEntry` takes `Guard` as an argument, but if we were to implement `Drop` for it, we need to somehow get `Guard` from `RefEntry` or call `epoch::pin()` inside `drop()`. I could not come up with a simple way without introducing a lot of rewriting, so I opted for a simpler approach.

stjepang:
That makes sense!

crossbeam-skiplist/src/map.rs
```diff
@@ -248,6 +249,20 @@ impl<'a, K, V> Entry<'a, K, V> {
...
+impl<'a, K, V> Drop for Entry<'a, K, V>
+{
+    fn drop(&mut self) {
+        let guard = &epoch::pin();
+        let dummy: [usize; 2] = [1; 2]; // 1 because a refrence to 0 will crash :)
+        unsafe {
+            // Safe because the size of `RefEntry` is same as two pointers,
+            // which in turn is equal to two `usize`s.
+            let transmuted = mem::transmute::<[usize; 2], base::RefEntry<K, V>>(dummy);
+            mem::replace(&mut self.inner, transmuted).release(guard);
```
stjepang: Perhaps a cleaner way to do this would be to wrap `inner` into a `ManuallyDrop` and then here do something like:

```rust
let guard = &epoch::pin();
unsafe {
    ManuallyDrop::take(&mut self.inner).release(guard);
}
```

topecongiro: Thanks, I was not aware of `ManuallyDrop`, seems like a great fit here.

`ManuallyDrop::take` is nightly-only API, so I have added a feature attribute to enable it.

topecongiro:
https://github.com/crossbeam-rs/crossbeam/pull/337/commits/a6de0ca23b5364d2acc876f6e3d74cb8ab4d7b4a The commit fixes memory leak when using iterator.

topecongiro:
Ah, should crossbeam build on both stable and nightly? Looks like there is a test on stable on Travis, but is that allowed to fail?

crossbeam-skiplist/src/map.rs
```diff
@@ -248,6 +249,16 @@ impl<'a, K, V> Entry<'a, K, V> {
     }
 }
 
+impl<'a, K, V> Drop for Entry<'a, K, V>
+{
+    fn drop(&mut self) {
+        let guard = &epoch::pin();
+        unsafe {
+            ManuallyDrop::take(&mut self.inner).release(guard);
```
stjepang: Oh, I didn't realize `ManuallyDrop::take` was unstable! :( Fortunately, we can replace it with:

```suggestion
            ManuallyDrop::into_inner(ptr::read(&mut self.inner)).release(guard);
```

topecongiro: Thanks! Fixed this in https://github.com/crossbeam-rs/crossbeam/pull/337/commits/7f4f564bc67f275ec43209274d48bb18de394720.

stjepang:
> The memory leak still persists when you use `Entry::remove()`.

Why does the memory leak persist with `Entry::remove()`? This PR should fix the leak in my understanding...

topecongiro:
@stjepang The memory leak was caused by an iterator, not `Entry::remove`. I initially misunderstood it because the original example used an iterator and `Entry::remove` together:

```rust
        for entry in m.iter(){
            entry.remove();
        }
```

With https://github.com/crossbeam-rs/crossbeam/commit/a6de0ca23b5364d2acc876f6e3d74cb8ab4d7b4a, the above code won't leak memory.

topecongiro:
Thanks, applied the suggestion in https://github.com/crossbeam-rs/crossbeam/pull/337/commits/3f1390d7ab3b8913dc4aa3a4c323f276f4f4ec79.

crossbeam-skiplist/src/base.rs
```diff
@@ -227,6 +227,24 @@ impl<K, V> Node<K, V> {
         }
     }
 
+    /// Decrements the reference count of a node, pinning the thread and destoying the node
+    /// if the count become zero.
+    #[inline]
+    unsafe fn decremnt_with_pin<F>(&self, pin: F)
```
stjepang: Typo: `decremnt` -> `decrement`

crossbeam-skiplist/src/base.rs
```diff
@@ -1401,6 +1419,21 @@ impl<'a, K: 'a, V: 'a> RefEntry<'a, K, V> {
...
+        unsafe { self.node.decrement(guard) }
+    }
+
+    /// Releases the reference of the entry, pinning the thread only when
+    /// the reference count of the node becomes 0.
+    pub fn release_with_pin<F>(self, pin: F)
+    where
+        F: FnOnce() -> Guard,
+    {
+        unsafe { self.node.decremnt_with_pin(pin) }
```
stjepang:
```suggestion
        unsafe { self.node.decrement_with_pin(self.parent, pin) }
```

crossbeam-skiplist/src/base.rs
stjepang:
```suggestion
    unsafe fn decrement_with_pin<F>(&self, parent: &SkipList<K, V>, pin: F)
```

crossbeam-skiplist/src/base.rs
```diff
@@ -227,6 +227,24 @@ impl<K, V> Node<K, V> {
...
+        F: FnOnce() -> Guard,
+    {
+        if self
+            .refs_and_height
+            .fetch_sub(1 << HEIGHT_BITS, Ordering::Release)
+            >> HEIGHT_BITS
+            == 1
+        {
+            fence(Ordering::Acquire);
+            pin().defer_unchecked(move || Self::finalize(self));
```
stjepang: Function `release()` has this `check_guard` call that will make sure the passed guard is the right one (belongs to the same GC instance the parent skip list uses). We need to do the same check here:

```rust
let guard = &pin();
parent.check_guard(guard);
defer_unchecked(move || Self::finalize(self));
```

Note that we'll also need to pass a reference to the parent skip list as an argument.

crossbeam-skiplist/src/base.rs
topecongiro: Oops, thank you. Fixed in https://github.com/crossbeam-rs/crossbeam/pull/337/commits/d07e0eb83bc33451b6587224467e8851982be103.

crossbeam-skiplist/src/base.rs
topecongiro: Thanks! Added a call to `check_guard` in https://github.com/crossbeam-rs/crossbeam/pull/337/commits/46c8b530b528eef27f86ded3c153e2ed3650e159.

topecongiro:
> Sorry for so many requests for changes, I promise this is the last one. 😅

No worries. Thank you for taking time to review this PR!

Issue #312, opened 2019-01-29
Crossbeam channel v0.3.7 fails to compile


LivingInSyn:
including Crossbeam channel v0.3.7 fails to compile with the error:

```
  Downloaded crossbeam-channel v0.3.7
  Downloaded crossbeam-utils v0.6.1
   Compiling quote v0.6.11
   Compiling crossbeam-channel v0.3.7
error[E0432]: unresolved import `crossbeam_utils::Backoff`
```

this appears to be because of a mismatch between channel and utils

stjepang:
Gah, it's because it requires the latest version of `crossbeam-utils`. :(
You can fix the problem with `cargo update -p crossbeam-utils`.

I'll publish a new minor version with a fix.

BurntSushi:
@stjepang You might find testing with minimal versions to be helpful. For example: https://github.com/rust-lang/regex/blob/60d087a23025e045ae754a345b04003c31d83d93/ci/script.sh#L53-L56

stjepang:
Ok, I've published 0.3.8 that bumps the `crossbeam-utils` dependency.

@BurntSushi Thank you, that's exactly what I'm looking for! :)

LivingInSyn:
Thanks, all!

jhpratt:
@stjepang Works for me! Looks like this can be closed.

stjepang:
@BurntSushi Transitive dependencies break under `-Z minimal-versions`:

```
    Checking void v1.0.0
error[E0259]: the name `core` is defined multiple times
  --> /home/stjepan/.cargo/registry/src/github.com-1ecc6299db9ec823/void-1.0.0/src/lib.rs:14:1
   |
14 | extern crate core;
   | ^^^^^^^^^^^^^^^^^^ `core` reimported here
   |
   = note: `core` must be defined only once in the type namespace of this module
help: you can use `as` to change the binding name of the import
   |
14 | extern crate core as other_core;
   |

error: aborting due to previous error
```

Lots of other folks reported similar problems with - it looks like the whole crates ecosystem is broken in this aspect because no one cares about minimal versions. :( Did you have better luck with `-Z minimal-versions`?

BurntSushi:
[...]

PR #341, opened 2019-03-10
Add minimal versions check to travis

taiki-e:
Related: https://github.com/crossbeam-rs/crossbeam/issues/312, https://github.com/rust-lang/cargo/issues/5657

### DIFF
--- a/ci/dependencies.sh
+++ b/ci/dependencies.sh
@@ -11,4 +11,9 @@ cargo tree
 cargo tree --duplicate
 cargo tree --duplicate || exit 1
 
+# Check minimal versions.
+cargo update -Zminimal-versions
+cargo tree
+cargo check --all --all-features --exclude benchmarks
+
 exit 0
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,7 @@ matrix:
     script: ./ci/crossbeam-utils.sh
 
   # Check for duplicate dependencies.
-  - rust: stable
+  - rust: nightly
     name: "dependency tree check"
     script: ./ci/dependencies.sh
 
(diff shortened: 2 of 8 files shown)


taiki-e:
(If adding `rand` (0.6 or later) to dependencies (not `dev-dependencies`) in the future, I think [this hack](https://github.com/rust-lang-nursery/futures-rs/blob/0.3.0-alpha.13/futures-util/Cargo.toml#L36) in `futures-rs` is helpful: https://github.com/rust-lang-nursery/futures-rs/pull/1425)

ci/dependencies.sh
```diff
@@ -11,4 +11,8 @@ cargo tree
 cargo tree --duplicate
 cargo tree --duplicate || exit 1
 
+# Check minimal versions.
+cargo update -Zminimal-versions
+cargo tree
```
stjepang: If we add a full check here, then we don't need the checks in other scripts (like `crossbeam-deque.sh`, `crossbeam-epoch.sh`, and so on):

```suggestion
cargo check --all --all-features --bins --examples --tests
```

taiki-e: Tests and examples are not published to crate.io as crates, so I think that they are unnecessary. And the changes included in this PR do not allow tests and examples to be compiled with minimal versions.

stjepang: That makes sense, sounds good to me! 👍

stjepang:
Looks like the tests are failing. Perhaps we need to drop `--bins` too?

taiki-e:
Yes, in addition, it seems necessary to exclude the benchmarks (crossbeam-channel/benchmarks).

taiki-e:
AFAIK benchmarks are not published, but is it OK to add "publish = false" field to `Cargo.toml` for clarity?

stjepang:
Yes, `publish = false` is a good idea since we don't want to publish that crate accidentally.

PR #348, opened 2019-03-23
Add timestamps, cyclic timestamp implementation

emc2:
Adds a timestamp trait, as well as an implementation of cyclic timestamps using a rock-paper-scissors counter combined with a traditional integer counter.  Unlike a naive integer counter, these these counters do not overflow; rather, they tolerate 2^n-2 concurrent successors, after which the ordering breaks down.

This is useful for implementing versioned values, which are in turn used in many lock-free techniques.

### DIFF
--- a/crossbeam-utils/src/cyclic_timestamp.rs
+++ b/crossbeam-utils/src/cyclic_timestamp.rs
@@ -1,0 +1,529 @@
+use timestamp::Timestamp;
+
+/// A cyclic timestamp value.  For a numeric type having 2^n bits, a
+/// CyclicTimestamp guarantees that if a timestamp is less than its
+/// next 2^k - 1 successors, where k = n - 2.
+///
+/// For large numeric (such as u32 or u64), this can be used as a
+/// version timestamp which will tolerate up to 2^30 and 2^62
+/// concurrent successor values respectively.
+///
+/// This is particularly useful for versioned pointers in lock-free
+/// protocols, where the number of concurrent successors is determined
+/// by the maximum time any thread can take to attempt an lock-free
+/// transaction (which is largely a function of how long it might be
+/// stalled).  These tolerance levels guarantee the integrity of the
+/// timestamp counter with overwhelming probability for local
+/// threading models (and even many distributed ones).
+#[derive(Copy, Clone, Debug, Hash, Eq, PartialEq)]
+pub struct CyclicTimestamp<S: Sized>(S);
+
+// These are implemented by splitting the number into two sections.
+// The lower 2^n-2 bits are a traditional counter.  The top 2 bits are
+// used to implement a "rock-paper-scissors" (RPS) counter.
+//
+// The RPS ordering looks like 0 < 1 < 2 < 0.  Thus, the RPS counter
+// tolerates at most two concurrent timestamps.
+//
+// We bump the RPS counter when the lower bit counter overflows,
+// resulting in a tolerance of 2^n-2 concurrent values.
+
+impl Timestamp for CyclicTimestamp<u8> {
+    fn new() -> CyclicTimestamp<u8> {
+        CyclicTimestamp(0)
+    }
+
+    fn earlier(&self, other: &CyclicTimestamp<u8>) -> bool {
+        let self_number = 0x3f & self.0;
+        let other_number = 0x3f & other.0;
+        let self_rps = self.0 & 0xc0;
+        let other_rps = other.0 & 0xc0;
+
+        (self_rps == other_rps && self_number < other_number) ||
+            ((self_rps == 0x00) && (other_rps == 0x40)) ||
+            ((self_rps == 0x40) && (other_rps == 0x80)) ||
+            ((self_rps == 0x80) && (other_rps == 0x00))
+    }
+
+    fn later(&self, other: &CyclicTimestamp<u8>) -> bool {
+        let self_number = 0x3f & self.0;
+        let other_number = 0x3f & other.0;
+        let self_rps = self.0 & 0xc0;
+        let other_rps = other.0 & 0xc0;
+
+        (self_rps == other_rps && self_number > other_number) ||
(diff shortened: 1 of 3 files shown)


crossbeam-utils/src/cyclic_timestamp.rs
```diff
@@ -0,0 +1,529 @@
+use timestamp::Timestamp;
+
+/// A cyclic timestamp value.  For a numeric type having 2^n bits, a
+/// CyclicTimestamp guarantees that if a timestamp is less than its
```
Vtec234: > if a timestamp is less than its 2^k1 - 1 successors, where k = n - 2
then.. what?

crossbeam-utils/src/cyclic_timestamp.rs
```diff
@@ -0,0 +1,529 @@
...
+fn test_u128_wrap() {
+    assert_eq!(CyclicTimestamp(0 as u128),
+               CyclicTimestamp(0xbfffffffffffffffffffffffffffffff as u128)
+                 .succ())
+}
+
+#[test]
+fn test_u8() {
+    let values: [CyclicTimestamp<u8>; 9] = [
+        CyclicTimestamp(0x00),
```
Vtec234: For this and other examples, it would be good to give some explanation of why these specific bit patterns are chosen (or that they're simply random) and maybe give them names if they're significant.

crossbeam-utils/src/timestamp.rs
```diff
@@ -0,0 +1,16 @@
+/// Trait representing basic operations on a timestamp value.
+/// Timestamps are often used in lock-free protocols to implement
+/// versioned values.
+pub trait Timestamp {
+    /// Create a new timestamp.
+    fn new() -> Self;
+
+    /// Check `self` is earlier than `other`.
+    fn earlier(&self, other: &Self) -> bool;
```
Vtec234: Does it make sense to implement `PartialOrd` for timestamps instead of custom `earlier`/`later` methods?

emc2: These don't observe transitivity.

emc2:
The most immediate use case for CyclicTimestamp is Lamport's bakery algorithm.  More generally, a common technique in lock-free data structures is to utilize "double-wide" compare-and-set instructions to create versioned values.  This is done for atomic snapshots, several lock-free data structures, several solutions to the lock-free memory reclamation problem, and Maged Michael's lock-free malloc.  In some of these cases, we compare timestamps for order, not just equality.  Finally, some distributed protocols use sequence numbers.

In all these cases where we end up comparing for order, there is an inevitable failure that occurs when integer addition wraps, as this will result in order comparisons incorrectly reporting the earlier timestamps to be later than the ones created after the addition wraps.

CyclicTimestamp eliminates this inevitable failure mode, replacing it with a failure mode that occurs only when the number of concurrently-existing timestamp exceeds a (typically very large) value.  In many protocols, there is at most one concurrent timestamp value per pending operation; therefore, failure would require a ridiculously huge number of threads to exist (which, in practice, cannot actually happen).  Even when this is not the case, CyclicTimestamp replaces a failure mode which is guaranteed to happen eventually with one that happens with a vanishingly small probability.

The "rock-paper-scissors" technique was taken from one of Herlihy's course lectures, and I believe is described in Herlihy and Shavit's book.

stjepang:
Thanks for the clarification! I see what rock-paper-scissors is doing now.

I was not able to find this technique in "The Art of Multiprocessor Programming" book. So what I'm failing to understand is where this is really used in practice. It's true that many lock-free algorithms use cyclic timestamps, but is the rock-paper-scissors technique used anywhere?

Another thing I'm not understanding is what are the cases where the following technique is failing, while rock-paper-scissors isn't?

```rust
fn earlier(&self, other: &Self) -> bool {
    let a = self.0;
    let b = other.0;
    a.wrapping_sub(b) > b.wrapping_sub(a)
}

fn later(&self, other: &Self) -> bool {
    let a = self.0;
    let b = other.0;
    a.wrapping_sub(b) < b.wrapping_sub(a)
}

fn succ(&self) -> Self {
    let a = self.0;
    let b = a.wrapping_add(1);
    Self(b)
}
```

emc2:
The official name was something like "trinary counters", or something similar, if I recall correctly.  I opted for the "rock-paper-scissors" lingo in the documentation, because it's easier to see what's going on.  I recall this technique from one of Herlihy's course lectures, I want to think it was alongside Lamport's Bakery algorithm (that would be a logical place for it), but it's been quite some time.

The biggest problem with your examples using wrapping_sub is that it's not clear what the specification _should_ say.  The characteristic wrapping case is where 0xffffffff wraps to 0 (on a 32-bit counter).  Should 0xffffffff < 0 be true?  If you're considering the call to succ which wraps, then yes.  But I can also argue for 0xffffffff > 0.

The trinary counter breaks out of this ambiguity by declaring that we only get valid results when comparing so many successors ahead.  Thus, we can write a consistent specification for how earlier and later should behave.

stjepang:
> The trinary counter breaks out of this ambiguity by declaring that we only get valid results when comparing so many successors ahead. Thus, we can write a consistent specification for how earlier and later should behave.

Doesn't the same specification also work with `wrapping_sub`, i.e. any timestamp is less than it's next `2^k - 1` successors?

If we implemented `CyclicTimestamp` using `wrapping_sub`, would we break any of its guarantees?

emc2:
Sorry for the delay, I've been tied up with a new job.

The wrapping_sub solution would need to use unsigned numbers, otherwise it breaks on 0-to-1. (0.wrapping_sub(1) = 0xffffffff > 1.wrapping_sub(0) = 1, which fails the later check).

This in turn fails on 0xffffffff-to-0 (since 0xffffffff = -1 in signed numbers): 0.wrapping_sub(-1) = 1 > -1.wrapping_sub(0) = -1, which also fails the check.

Issue #345, opened 2019-03-18
Ridiculous corner case on AtomicCell reads


mtak-:
For `AtomicCell`'s that use seqlocks, the `usize` in the lock can, in some extreme scenario, wrap around in the middle of a read, causing the read validation to succeed even though the read is torn.

Copying the repro from here: https://github.com/Amanieu/seqlock/issues/3

> Thread 1:
> Starts a read, getting a sequence number, and reads some of the data.
>
> Other threads:
> Successfully perform exactly `2^(mem::size_of::<usize>() - 1)` writes, leaving the sequence number effectively unchanged.
>
> Thread 1:
> Finishes reading the data, and sees the sequence number unchanged. It then returns the torn read.
>
> On 64-bit machines, this would take a few lifetimes to occur, on 32-bit machines it might require the reader thread to be inactive for around one second.

I'm not sure if this worth caring about, but thought I'd post it anyway.

stjepang:
> on 32-bit machines it might require the reader thread to be inactive for around one second.

A single update operation is somewhere on the order of around 10 ns, which means 2^32 updates in a loop would take 43 seconds. So in order for a torn read to happen, a reader thread would have to be paused for 43 seconds and then continue running at just the right moment in between two update operations.

Considering how OS schedulers work and how unlikely that scenario is to happen, I wouldn't worry about it.

Do my calculations seem correct?

mtak-:
1 bit is reserved as the lock bit, so 2^31 updates is the correct number. I benched `fetch_add`s at around 4ns on my (64bit) machine, and conservatively dropped that to about 0.5ns (2GHz) giving the 1s number. Using `10ns`, it would be about `21.5s`.

Full disclosure: I have a similar issue in swym's stm implementation, and I'm trying to get a feel for what, if anything, there is to do about it.

stjepang:
Perhaps we should just use [`AtomicU64`](https://doc.rust-lang.org/nightly/std/sync/atomic/struct.AtomicU64.html) (will be stable in Rust 1.34, to be released on April 11th)?

mtak-:
`AtomicU64::load` on `i686` compiles to a `lock cmpxchg8b` [godbolt](https://rust.godbolt.org/z/gnx5Ng), which would be a substantial hit to performance.

[...]

PR #359, opened 2019-04-23
Implement wide sequence lock

jeehoonkang:
In #345, @mtak- discusses the problem of using `AtomicUsize` as a wrapping counter: wrapping takes only ~10 seconds, which will render the Crossbeam implementation of sequence lock wrong.  Using `AtomicU64` was proposed as a solution, but it is compiled to an expensive compare-exchange instruction in 32-bit x86 architectures.

Here is another solution: using two `AtomicUsize` as a counter.  Observation is that we don't need the full atomicity of `AtomicU64` for the counter of sequence lock.  We just can use two `AtomicUsize` variables and synchronize accesses to them properly.

r? @stjepang


jeehoonkang:
https://travis-ci.org/crossbeam-rs/crossbeam/jobs/523343511#L464

Tempting to bump the minimum Rust version..  @stjepang @Vtec234 what do you think?

stjepang:
Oh, I like this change! Will take another look and do a proper review today :)

Regarding bumping the minimum Rust version... don't worry about it, that's fine.

crossbeam-utils/src/atomic/mod.rs
```diff
@@ -1,5 +1,16 @@
 //! Atomic types.
 
+cfg_if! {
+    // TODO(jeehoonkang): want to express `target_pointer_width >= "64"`, but it's not expressible
```
Vtec234: In principle Rust also supports [a 16-bit target](https://github.com/rust-lang/rust/blob/9ebf47851a357faa4cd97f4b1dc7835f6376e639/src/librustc_target/spec/msp430_none_elf.rs) for which two `AtomicUsize`s would make a 32-bit counter. I think this can safely be ignored as it's very primitive hardware, but mentioning it here for completeness. There are no 128-bit targets currently as far as I can tell from a quick search.

mtak-:
iOS has a very weird scheduler where this issue is more likely to crop up under certain scenarios (see [Update: A Warning for iOS](https://stackoverflow.com/questions/5869825/when-should-one-use-a-spinlock-instead-of-mutex/5870415#5870415)). Thanks for the fix!

stjepang:
Just so we don't forget, in all the other places where we have atomic timestamps, we should use `AtomicU64` (I believe we can't use wide sequences like in this PR, unfortunately):

- epoch representation
- head/tail index in deque, channel, and queues

But that can come later in another PR. It will also require Rust 1.34 minimum.

crossbeam-utils/src/atomic/seq_lock_wide.rs
```diff
@@ -0,0 +1,106 @@
...
+            state_lo: AtomicUsize::new(0),
+        }
+    }
+
+    /// If not locked, returns the current stamp.
+    ///
+    /// This method should be called before optimistic reads.
+    #[inline]
+    pub fn optimistic_read(&self) -> Option<(usize, usize)> {
+        let state_hi = self.state_hi.load(Ordering::Acquire);
```
Vtec234: Any reason why this is `Acquire` while other loads of `state_hi` are `Relaxed`?

crossbeam-utils/src/atomic/mod.rs
jeehoonkang: I added some documentation: https://github.com/crossbeam-rs/crossbeam/pull/359/files#diff-7e838416d7e6b94eeac80063d0b6a446R10

crossbeam-utils/src/atomic/seq_lock_wide.rs
jeehoonkang: I left some explanation on this and other orderings.

jeehoonkang:
@stjepang @Vtec234 Thank you for your comments.  I addressed all of them.

jeehoonkang:
@stjepang any opinions on this PR?

jeehoonkang:
ping? @stjepang

PR #361, opened 2019-04-23
Fix CI errors

jeehoonkang:
The CI says (https://travis-ci.org/crossbeam-rs/crossbeam/jobs/523343513#L485):

```
error: variable does not need to be mutable
  --> crossbeam-channel/src/flavors/tick.rs:63:21
   |
63 |                 let mut delivery_time = self.delivery_time.load();
   |                     ----^^^^^^^^^^^^^
   |                     |
   |                     help: remove this `mut`
   |
   = note: `-D unused-mut` implied by `-D warnings`
error: aborting due to previous error
```

(Not sure why warning is issued as an error.)  Fixing this error.

### DIFF
--- a/crossbeam-channel/src/flavors/tick.rs
+++ b/crossbeam-channel/src/flavors/tick.rs
@@ -60,7 +60,7 @@ impl Channel {
         loop {
             // Compute the time to sleep until the next message or the deadline.
             let offset = {
-                let mut delivery_time = self.delivery_time.load();
+                let delivery_time = self.delivery_time.load();
                 let now = Instant::now();
 
                 // Check if we can receive the next message.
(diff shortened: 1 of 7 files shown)


stjepang:
Warnings are issued as errors on CI because of lines like this one:

https://github.com/crossbeam-rs/crossbeam/blob/2df0d5d02a41935769308ceb2b6c1effe74bda1b/ci/crossbeam-channel.sh#L6

This is not really necessary, but good to have in CI so that no warning accidentally slips through.

It looks like the latest nightly is a bit smarter about unused `mut`s and was able to detect a few cases it wasn't able before. Which is great!

Thanks for the PR! I just pushed a change removing a few more `mut`s. 

Hopefully this works now...

bors r+

PR #363, opened 2019-04-24
[WIP] Add push/pop lockless

Restioson:
This PR will add non-deadlocking push/pop operations to queues -- they will never deadlock even if a concurrent push/pop stalls for whatever reason indefinitely.

### DIFF
--- a/crossbeam-queue/src/array_queue.rs
+++ b/crossbeam-queue/src/array_queue.rs
@@ -216,6 +216,84 @@ impl<T> ArrayQueue<T> {
         }
     }
 
+    /// Attempts to push an element into the queue without deadlocking.
+    ///
+    /// If the queue is full *or* if another operation would cause a potentially endless retry loop
+    /// (if the other operation is stalled) the element is returned back as an error.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use crossbeam_queue::{ArrayQueue, PushError};
+    ///
+    /// let q = ArrayQueue::new(1);
+    ///
+    /// assert_eq!(q.push(10), Ok(()));
+    /// assert_eq!(q.push(20), Err(PushError(20)));
+    /// ```
+    pub fn push_lockless(&self, value: T) -> Result<(), PushError<T>> {
+        let backoff = Backoff::new();
+        let mut tail = self.tail.load(Ordering::Relaxed);
+
+        loop {
+            // Deconstruct the tail.
+            let index = tail & (self.one_lap - 1);
+            let lap = tail & !(self.one_lap - 1);
+
+            // Inspect the corresponding slot.
+            let slot = unsafe { &*self.buffer.add(index) };
+            let stamp = slot.stamp.load(Ordering::Acquire);
+
+            // If the tail and the stamp match, we may attempt to push.
+            if tail == stamp {
+                let new_tail = if index + 1 < self.cap {
+                    // Same lap, incremented index.
+                    // Set to `{ lap: lap, index: index + 1 }`.
+                    tail + 1
+                } else {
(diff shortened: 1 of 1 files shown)


crossbeam-queue/src/array_queue.rs
```diff
@@ -216,6 +216,84 @@ impl<T> ArrayQueue<T> {
...
+    ///
+    /// ```
+    /// use crossbeam_queue::{ArrayQueue, PushError};
+    ///
+    /// let q = ArrayQueue::new(1);
+    ///
+    /// assert_eq!(q.push(10), Ok(()));
+    /// assert_eq!(q.push(20), Err(PushError(20)));
+    /// ```
+    pub fn push_lockless(&self, value: T) -> Result<(), PushError<T>> {
```
jeehoonkang: The code is very much similar with `push()`.  Do you think can we somehow merge the code of the two functions?

Restioson: Yes, I did think this and tried to merge them. Unfortunately it made the tests hang... So I will have another go at that when I have time. Before it is merged that should be done though.

stjepang:
@Restioson I have a high-level question about your use case with interrupts pushing keyboard events into the queue.

So if your interrupt handler is defined like this:

```rust
fn handler(event: KeyboardEvent) {
    queue.push(event);
}
```

What happens if an event handler (call it A) is interrupted *before* `push` is called by another interrupt handler (call it B)? Does that mean keyboard event from A will end up in the queue *after* event from B?

That sounds wrong to me so I wonder how keyboard drivers handle situations like these? Are you sure this is the way to go?

My guess would be that event handlers shouldn't be interruptable and maybe the hardware guarantees so, but I've no idea if what I'm saying is true.

Restioson:
Yes, event handlers cannot be interrupted (interrupts are disabled by the os). This is because if nesting is allowed, excessive nesting depth could cause a stackoverflow. However, the local APIC is able to buffer one IRQ to be dispatched if the cpu has disabled interrupts, which will be dispatched when the cpu re-enables them.

I do believe that I still need the function though, for instance in case the push interrupts a pop from the handler. And nevertheless, it could be useful for others.

stjepang:
@Restioson Thanks for clarifying!

And just a couple more questions to see if I understood the problem domain correctly:

1. Is it true that there is only a single consumer?
2. Is it true that there is only a single producer?
3. Is it true that even though you're using the queue in the SPSC fashion, you want a MPSC or MPMC interface only to get around Rust's ownership/sharing rules?

Restioson:
1) yes
2) yes
3) yes pretty much :) it has to be static unfortunately

stjepang:
In that case, the way to go would really be https://github.com/crossbeam-rs/crossbeam/pull/338, which is ~5 times faster than `ArrayQueue` anyway :)

It's unfortunate that the SPSC queue cannot be used in your use case without `unsafe {}` blocks, but that is really the nature of the problem of interrupts. So my suggestion would be to use `static mut` variable then...

Restioson:
Aren't those slated to be removed? Perhaps a RefCell. I will consider it though. Do you still think that this has a use case?

stjepang:
> Aren't those slated to be removed?

Possibly -- I'm not sure. The alternative to `static mut` would be a static `UnsafeCell<T>` variable.

> Do you still think that this has a use case?

I don't think so. The SPSC queue is really designed to be wait-free and applicable in embedded systems, device drivers, and similar use cases. Others queues are a bit more high-level and can't provide strong progress guarantees without jumping through hoops or having tricky caveats.

Restioson:
Sorry, I haven't had much time to work on this because of exams... I will get back to it when they are finished though.

PR #364, opened 2019-04-26
Make crossbeam no std

Restioson:
This makes the crossbeam-queue crate no_std/alloc and fixes no_std in crossbeam as a whole (I couldn't get it working without the fix).


Restioson:
This still has a bug apparently - will fix soon. Don't merge yet.

crossbeam-queue/Cargo.toml
```diff
@@ -15,6 +15,14 @@ description = "Concurrent queues"
 keywords = ["queue", "mpmc", "lock-free", "producer", "consumer"]
 categories = ["concurrency", "data-structures"]
 
+[features]
+default = ["std"]
+nightly = ["crossbeam-utils/nightly"]
```
Vtec234: If you're adding a `nightly` feature to `crossbeam-queue`, I believe you should modify `ci/crossbeam-queue.sh` to match the structure of e.g. `ci/crossbeam-epoch.sh` wrt. handling features and `TRAVIS_RUST_VERSION`.

Restioson: Sure. I don't know whether it necessarily needs it - I just copied the setup from crossbeam-epoch.

stjepang: It would be good to test `crossbeam-queue` without the default features and with the `nightly` feature, just in case we add something to the crate that does require `std` or `nightly` :)

So you need to change the contents of `ci/crossbeam-queue.sh` to the following:

```bash
#!/bin/bash

cd "$(dirname "$0")"/../crossbeam-queue
set -ex

export RUSTFLAGS="-D warnings"

cargo check --no-default-features
cargo check --bins --examples --tests
cargo test

if [[ "$TRAVIS_RUST_VERSION" == "nightly" ]]; then
    cargo check --no-default-features --features nightly
    cargo test --features nightly
fi
```

Restioson:
I couldn't actually get crossbeam to work in no_std even with the following setup:

```toml
[dependencies.crossbeam]
path = "../../crossbeam"
default-features = false
features = ["nightly"]
```

I've restructured how no std works (necessary for no std lazy_static). I don't know if this is correct or desired. It might count as breaking too. I've pushed it now so it can be reviewed (and also in the meanwhile used in [flower](https://github.com/flower-os/flower), because we couldn't figure out how to get it to compile on no std without these changes).

stjepang:
I think commit `2639822` was fine, we just needed to add more tests to CI. :)

Are you sure that the setup you mentioned doesn't work?

```toml
[dependencies.crossbeam]
version = "0.7"
default-features = false
features = ["nightly"]
```

Is there perhaps a way for me to reproduce the problem?

Restioson:
@stjepang -- it will work/fail silently on platforms with std, as the dependencies will just link it in -- including the subcrates. However, on platforms where std is not present at all (like, in my case, for Flower's kernel's custom target) it will fail, as std does not exist. See [this stackexchange question](https://stackoverflow.com/questions/55330493/why-is-randrng-able-to-work-in-a-no-std-environment-even-when-i-have-not-set-d) for a bit more background.

stjepang:
@Restioson So if I understood correctly, it seems like Crossbeam is properly setting `#![no_std]` for itself, but Crossbeam's dependencies are not, therefore pulling `std` in?

In that case perhaps we should fix those dependencies or conditionally disable them if the `std` feature is not enabled? Do you know which crates are at fault here?

Restioson:
I think it's mostly the subcrates. In the commit I pushed I have fixed it and verified that it works (we are currently using the fork in flower on local repo).

taiki-e:
The current crossbeam's CI does not check for no_std environment, so we need to add some checks to CI and make some corrections, but I don't think a big change like this PR is necessary.

Restioson:
@taiki-e if you can figure out a way to get crossbeam to work truly without std without these kinds of changes I will be impressed lol. Currently the fact that it isn't using `std` isn't propagated to the subcrates nor its dependencies.

Restioson:
After some discussion (in that PR) closing in favour of #368.

PR #365, opened 2019-04-27
Require 1.26.0 in deque and utils

stjepang:
`crossbeam-deque` and `crossbeam-utils` don't really need 1.28.0, we can support 1.26.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,8 @@ matrix:
   - rust: 1.28.0
     name: "crossbeam-channel on 1.28.0"
     script: ./ci/crossbeam-channel.sh
-  - rust: 1.28.0
-    name: "crossbeam-deque on 1.28.0"
+  - rust: 1.26.0
+    name: "crossbeam-deque on 1.26.0"
     script: ./ci/crossbeam-deque.sh
   - rust: 1.28.0
     name: "crossbeam-epoch on 1.28.0"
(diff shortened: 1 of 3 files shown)


cuviper:
Ah, the latest *published* versions work fine with 1.26, but master is using `#[repr(transparent)]`.

stjepang:
@cuviper Thanks for tracking this down, I forgot about `#[repr(transparent)]`.

What do you think about introducing `legacy` feature flag to `crossbeam-deque` and `crossbeam-utils` and always supporting whatever the Rust version is used by Rayon?

cuviper:
If you're willing to carry such conditions in the code, I'd rather see that auto-detected, so rayon doesn't have to be an inhibitor. It could either use a version check or probing like `autocfg`, though I don't have an API to probe for attributes yet. Also note that we can't probe crossbeam features from the rayon side, because build scripts can't affect cargo-level features.

If you're open to this, I can take it on myself to make a PR.

stjepang:
@cuviper Yeah, that'd be a cleaner solution because then we wouldn't have to add new features to `Cargo.toml`.

I'd be happy to accept a PR for this :) Thanks!

stjepang:
@cuviper How about this? https://github.com/dtolnay/select-rustc

cuviper:
Yeah, select-rustc looks really cool -- but it's a 2018 edition crate, so it needs at least 1.31 itself.

jeehoonkang:
I'm proposing to revert #360, and support Rust 1.26 (or any version that is supported by Rayon).  #360 is good, but I think its benefit doesn't outweigh its drawback, which is breaking the compatibility with Rayon.

stjepang:
@jeehoonkang Yeah, maybe we should revert that. In fact, only `#[repr(transparent)]` bumps the minimum Rust version so it's enough if we just remove that line.

PR #368, opened 2019-04-29
Fix no_std build

taiki-e:
Closes https://github.com/crossbeam-rs/crossbeam/pull/364


ci/crossbeam-epoch.sh
```diff
@@ -20,4 +20,9 @@ if [[ "$TRAVIS_RUST_VERSION" == "nightly" ]]; then
         --target x86_64-unknown-linux-gnu \
         --features sanitize,nightly \
         --example sanitize
+
+    # Check for no_std environment.
+    cd .. && cargo run --manifest-path ci/Cargo.toml --bin remove-dev-dependencies */Cargo.toml && cd -
+    cargo check --target thumbv7m-none-eabi --no-default-features
```
taiki-e: If build with Linux as a target, it will link to std, so it need to specify a target that does not have std.

ci/crossbeam-epoch.sh
```diff
@@ -20,4 +20,9 @@ if [[ "$TRAVIS_RUST_VERSION" == "nightly" ]]; then
         --target x86_64-unknown-linux-gnu \
         --features sanitize,nightly \
         --example sanitize
+
+    # Check for no_std environment.
+    cd .. && cargo run --manifest-path ci/Cargo.toml --bin remove-dev-dependencies */Cargo.toml && cd -
```
taiki-e: If there is a package that requires std in development dependency, it may not be checked properly in no_std environment. See https://github.com/rust-lang-nursery/futures-rs/issues/1396#issue-395720940 for more.

taiki-e: This may not be necessary with the current crossbeam, so I will try to remove it once Travis green.

Restioson:
I see what you've done here & agree that it's a better solution to what I have done... however, I'm not sure how it handles lazy_static. Lazy_static is available for `no_std` with the `spin_no_std` feature (which by the way was originally why I changed the feature to `no_std`.

taiki-e:
AFAIK the current no_std crossbeam does not depend on lazy_tactic.

The crossbeam currently uses lazy_static in two places:

https://github.com/crossbeam-rs/crossbeam/blob/86eff71c893644d5150c209fe903472472a15189/crossbeam-utils/src/sync/sharded_lock.rs#L556-L562

https://github.com/crossbeam-rs/crossbeam/blob/86eff71c893644d5150c209fe903472472a15189/crossbeam-epoch/src/default.rs#L10-L13

However, these are both in modules that require std. Also, they seem to be used with features not available on no_std.

https://github.com/crossbeam-rs/crossbeam/blob/86eff71c893644d5150c209fe903472472a15189/crossbeam-utils/src/sync/sharded_lock.rs#L7-L10

https://github.com/crossbeam-rs/crossbeam/blob/86eff71c893644d5150c209fe903472472a15189/crossbeam-epoch/src/default.rs#L15-L18

So I don't think crossbeam needs to think about lazy_static's no_std support.

taiki-e:
Also, I believe many crates in the ecosystem (e.g., [serde](https://github.com/serde-rs/serde/blob/28ce892617573d8da6d86bcb8a504aa1d8097b07/serde/Cargo.toml#L40), [futures](https://github.com/rust-lang-nursery/futures-rs/blob/50f3f71758bc295cff76d157b909eba058efde3d/futures/Cargo.toml#L40), [itertools](https://github.com/bluss/rust-itertools/blob/70988ff4f8a8b52f00d4109fbf987fecfc24c3c2/Cargo.toml#L35)) support the no_std environment in the same way as the current crossbeam.

lazy_static supports no_std in the other way as these crates, because it adds `spin` dependency.
(I think it can be solved by using their own `Once` and providing no_std support with the same way as the current crossbeam. However, I do not know if the lazy_static developer agrees with this. Also, to not break crates that currently use the spin_no_std feature, it needs to do no_std support on both `spin_no_std` and no-default-features instead of removing `spin_no_std` immediately, but I think doing this is not very difficult.)

----

FYI, you can always use the lazy_static/spin_no_std feature or configure the feature to allow users to opt-in like the following:

```toml
[dependencies]
lazy_static = [ version = "1", optional = true ]
[features]
std = ["lazy_static"]
spin_no_std = ["lazy_static/spin_no_std"]
```

The point to keep in mind when always using the lazy_static/spin_no_std feature is that it has not been tested on older versions than the latest stable ([lazy_static](https://github.com/rust-lang-nursery/lazy-static.rs/blob/88994b26744fac2ceb713a9285985f76195c9c89/.travis.yml#L8), [spin](https://github.com/mvdnes/spin-rs/blob/d0dbe6b2b0381f674e49075beda824d475e7bd95/.travis.yml#L4)).

crossbeam-epoch/src/lib.rs
```diff
@@ -68,31 +68,33 @@ cfg_if! {
...
 }
 
-extern crate arrayvec;
-extern crate crossbeam_utils;
-#[macro_use]
-extern crate memoffset;
-#[macro_use]
-extern crate scopeguard;
+cfg_if! {
+    if #[cfg(any(feature = "std", feature = "nightly"))] {
```
taiki-e: `alloc` crate is already stable on rust 1.36, I think it makes sense to add the `alloc` feature and replace it with `#[cfg(any(feature = "alloc", feature = "std"))]`.

ci/crossbeam-epoch.sh
taiki-e: Well, this was not necessary because the current crossbeam development dependency was the only rand. However, in the future, it may change.

@stjepang Do you think we should have this from now?

crossbeam-epoch/src/lib.rs
taiki-e: Added 22d1442.

Restioson:
> FYI, you can always use the lazy_static/spin_no_std feature or configure the feature to allow users to opt-in like the following:

Yeah I just didn't want to fragment the options.

stjepang:
@taiki-e This looks great to me, thanks so much! :)

The only thing left to do is to update the readmes -- check out how we're currently explaining which parts of Crossbeam are available with individual features. We should probably also mention the `alloc` feature there.

taiki-e:
@stjepang 
I updated readmes and I added e3507f9 so that even targets that do not support CAS(compare-and-swap) operations such as thumv6m can be properly compiled. (Although `#![cfg_attr(feature = "nightly", feature(cfg_target_has_atomic))]` was originally valid for some crates, [errors will occur if this configuration is not propagated](https://github.com/rust-lang-nursery/futures-rs/pull/1505).)

Cargo.toml
```diff
@@ -22,30 +22,42 @@ travis-ci = { repository = "crossbeam-rs/crossbeam" }
 [features]
 default = ["std"]
```
stjepang: Should the `default` set of features also contain `"alloc"`?

taiki-e: I think it is not necessary because this PR enables Items requiring `alloc` under the following conditions.

https://github.com/crossbeam-rs/crossbeam/blob/61e15de65fcb70f82dbade006efef857885032d9/crossbeam-skiplist/src/lib.rs#L26

taiki-e: In addition to this, this PR does `extern crate alloc` when `alloc` feature is enabled (The minimum required version of the `alloc` feature is Rust 1.36, so it is newer than the minimum required version of crossbeam).

We can change this to prioritize the std feature, but it is not recommended to include `alloc` as the default, as it may add dependencies that have similar implementations (that breaks the minimum required version).

crossbeam-epoch/README.md
```diff
@@ -20,7 +20,7 @@ immediately. Epoch-based GC is an efficient mechanism for deferring destruction
 shared objects until no pointers to them can exist.
 
 Everything in this crate except the global GC can be used in `no_std` environments that implement
-`alloc`.
+`alloc`. It needs to enable the `alloc` feature of this crate to use this crate, in the `no_std` environments.
```
stjepang: Perhaps we could rephrase this as:

```suggestion
`alloc`. The `alloc` feature of this crate needs to be enabled in `no_std` environments.
```

Cargo.toml
stjepang: Makes sense, thanks!

crossbeam-epoch/README.md
taiki-e: Yeah.

stjepang:
Thanks again, this looks ready for landing now! :)

bors r+

PR #371, opened 2019-05-08
epoch: Add Atomic::into_owned

### LINKED ISSUE #369: crossbeam_epoch::Atomic::into_owned
Hello

I'm wondering why the `Atomic` doesn't have an `unsafe fn into_owned(self) -> Owned<T>` method. I'd find it handy in a data structure destructor, when I'm sure that no concurrent accesses are being made to the structure any more, as it's in the destructor which is `&mut self`.

Now, I can do a load on the atomic with `epoch::unprotected` and convert that shared. But is there a particular reason why the above method doesn't exist? If not, is it OK if I send a PR?

vorner:
To mimic Shared::into_owned.

Closes #369

### DIFF
--- a/crossbeam-epoch/CHANGELOG.md
+++ b/crossbeam-epoch/CHANGELOG.md
@@ -1,3 +1,5 @@
+- Add `Atomic::into_owned()`.
+
 # Version 0.7.1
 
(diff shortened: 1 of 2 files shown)


stjepang:
Thanks!

bors r+

Issue #315, opened 2019-01-30
Soundness of AtomicCell::compare_exchange is dubious


RalfJung:
`AtomicCell<T>::compare_exchange` will, if I understand the code correctly, use `AtomucU*::compare_exchange_weak` if `T` has the right size.  But this means that e.g. for `(u8, u16)`, which has size 4, it will transmute these pairs into `u32` and compare them as integers.  However, such pairs contain a padding byte, meaning that the `u32` contains some uninitialized memory.  Having uninitialized memory in a  `u32` is [uncharted territory at best](https://github.com/rust-rfcs/unsafe-code-guidelines/issues/71), but performing actual operations on such a datum (as opposed to just load/store) brings us very close to UB.  I am not sure what the rules are in LLVM for this case, but under the proposed `poison` semantics I think the comparison will return `poison` and hence using it in an `if` is UB.  For Rust, I think the intention is to make any operation on uninitialized data UB (like it is in C), meaning the comparison would already return UB.

Strictly speaking, this affects not just the fast path, but also the arbitrarily-sized case: `bytes_eq` might compare uninitialized bytes.

BurntSushi:
> But this means that e.g. for `(u8, u16)`, which has size 4, it will transmute these pairs into u32 and compare them as integers.

If you don't mind me asking, is the `transmute` itself UB as well? In particular, I had thought that the in-memory representation of tuples isn't guaranteed? Or is it just the ordering of fields that isn't guaranteed?

RalfJung:
Well, the code checks that the sizes match.  Ignoring complicated issues around pointers, the only way in which transmuting a 4-byte piece of data to `u32` is UB is if there are bit patterns that are not valid for `u32` -- which boils down to the question of whether uninitialized bytes are allowed in `u32`.

jeehoonkang:
It seems C/C++ standard committee members tend to agree that all bit patterns are valid for `u32` and other unsigned integer types: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0907r4.html "For unsigned integer types, the bits of the object representation are ..."  I think it's reasonable, because (once we agree to use two's complement) there's no other sane choice.

stjepang:
@RalfJung Fair enough, this is difficult. :(

It might be helpful to look at what C/C++ do on atomic compare-exchange with arbitrary (non-POD) data. This is interesting:

[...]

PR #372, opened 2019-05-10
De-deprecate AtomicCell::get_mut

jeehoonkang:
PR #332 deprecated after the discussion of issue #315, but it is a
measure not yet necessary to take. Revert the deprecation and wait
for us to reach a consensus.

### DIFF
--- a/crossbeam-utils/src/atomic/atomic_cell.rs
+++ b/crossbeam-utils/src/atomic/atomic_cell.rs
@@ -169,8 +169,11 @@ impl<T: ?Sized> AtomicCell<T> {
     ///
     /// assert_eq!(a.load(), 8);
     /// ```
-    #[doc(hidden)]
-    #[deprecated(note = "this method is unsound and will be removed in the next release")]
+    ///
+    /// # Caveat
+    ///
+    /// This method might be unsound and removed in a future release, but we haven't decided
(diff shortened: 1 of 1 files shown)


stjepang:
Since we're not sure whether this is unsound, could we perhaps change the deprecation message to `"this method might be unsound and removed in a future release, but we haven't decided yet"` and then un-deprecate it if we decide it is sound?

jeehoonkang:
How about leaving the message and pointing to #315 in the comment, but still de-deprecating the function?

stjepang:
Sure! That sounds fine as well :)

jeehoonkang:
@stjepang I updated the PR.  Thanks for feedback!

stjepang:
Thanks!

bors r+

Centril:
> Since we're not sure whether this is unsound, [...]

@stjepang If you are not sure this is sound, then I think the right thing to do is to deprecate and eventually remove it until such time that you are sure. In my view, the onus is on you to show why some abstraction using `unsafe` is sound. Pointing towards the absence of a counter-example is not a good argument. Meanwhile, I think @RalfJung has made a good case for why the soundness is dubious.

stjepang:
@Centril I have reverted this commit and published a new version. Thank you for your feedback!

Centril:
Thank you!

PR #373, opened 2019-05-10
Support Rust 1.26

jeehoonkang:
Closes #365 

Remaining question: should we commit to support the minimum supported Rust version of Rayon?


cuviper:
> should we commit to support the minimum supported Rust version of Rayon?

It would be nice for me. :smile: 

Or at least, if you always bump semver when increasing your minimum, Rayon won't be affected.

stjepang:
Thanks for resolving this!

bors r+

PR #376, opened 2019-05-14
Reduce stall in list iteration

jeehoonkang:
In list iteration, even though the CAS to delete a node fails, the iteration may be able to proceed by re-reading the predecessor's next pointer.  A more precise explanation is in the comments.

### DIFF
--- a/crossbeam-epoch/src/sync/list.rs
+++ b/crossbeam-epoch/src/sync/list.rs
@@ -248,10 +248,11 @@ impl<'g, T: 'g, C: IsElement<T>> Iterator for Iter<'g, T, C> {
                 debug_assert!(self.curr.tag() == 0);
 
                 // Try to unlink `curr` from the list, and get the new value of `self.pred`.
-                let succ = self
+                let succ = match self
                     .pred
                     .compare_and_set(self.curr, succ, Acquire, self.guard)
-                    .map(|_| {
+                {
+                    Ok(_) => {
                         // We succeeded in unlinking `curr`, so we have to schedule
                         // deallocation. Deferred drop is okay, because `list.delete()` can only be
                         // called if `T: 'static`.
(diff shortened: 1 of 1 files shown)


jeehoonkang:
ping? @Vtec234 @stjepang

crossbeam-epoch/src/sync/list.rs
```diff
@@ -247,32 +247,39 @@ impl<'g, T: 'g, C: IsElement<T>> Iterator for Iter<'g, T, C> {
                 // node leaves the list in an invalid state.
```
Vtec234: Shouldn't this comment actually say "The tag should always be zero, .."?

Vtec234:
Although I can't really convince myself that this is definitely correct, it *seems* okay and I trust that you know what you're doing :) Have you observed performance diffs with this change?

crossbeam-epoch/src/sync/list.rs
jeehoonkang: Thanks!  just fixed it.

crossbeam-epoch/src/sync/list.rs
```diff
@@ -243,36 +243,43 @@ impl<'g, T: 'g, C: IsElement<T>> Iterator for Iter<'g, T, C> {
...
                 let succ = succ.with_tag(0);
 
-                // The tag should never be zero, because removing a node after a logically deleted
+                // The tag should always be zero, because removing a node after a logically deleted
                 // node leaves the list in an invalid state.
                 debug_assert!(self.curr.tag() == 0);
 
-                match self
+                // Try to unlink `curr` from the list, and get the new value of `self.pred`.
+                let succ = self
```
stjepang: Can you perhaps rewrite this as `let succ = match self.pred.compare_and_set...`?

I find matching easier to understand than combinators like `.map(...).unwrap_or_else(...)` -- I have to pause and think for a moment what is actually going on here.

jeehoonkang: I agree.  I changed it and rebased on `master`.  Thanks~

stjepang:
This looks good to me!

bors r+

PR #377, opened 2019-05-16
Fix list's finalize API

jeehoonkang:
When an entry is unlinked from list, the current implementation calls `defer_unchecked()` the invocation of the entry's finalizer.  This PR proposes to fix it in such a way that when an entry is unlinked from list, the finalizer is immediately called.

Here is an example why the current implementation is wrong, and this PR fixes it.  Suppose an element is in two (intrusive) linked lists, and their corresponding two finalizers (1) check if the element is unlinked from both lists, and if so, (2) defer the deallocation of the element.  The current implementation cannot express this idea because the invocation of the entry's finalizer is deferred.  On the other hand, this PR can express this idea directly.

### DIFF
--- a/crossbeam-epoch/src/internal.rs
+++ b/crossbeam-epoch/src/internal.rs
@@ -45,7 +45,7 @@ use core::sync::atomic::Ordering;
 use arrayvec::ArrayVec;
 use crossbeam_utils::CachePadded;
 
-use atomic::Owned;
+use atomic::{Shared, Owned};
 use collector::{Collector, LocalHandle};
 use deferred::Deferred;
 use epoch::{AtomicEpoch, Epoch};
@@ -501,9 +501,8 @@ impl IsElement<Local> for Local {
         &*local_ptr
     }
 
-    unsafe fn finalize(entry: &Entry) {
-        let local = Self::element_of(entry);
-        drop(Owned::from_raw(local as *const Local as *mut Local));
+    unsafe fn finalize(entry: &Entry, guard: &Guard) {
+        guard.defer_destroy(Shared::from(Self::element_of(entry) as *const _));
     }
 }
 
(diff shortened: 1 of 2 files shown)


Vtec234:
I'm a little confused as to the purpose of this. Surely if a value of some type `T` is in two intrusive lists, you couldn't call `finalize` from both anyway, because then `T::drop` would run twice on the same value?

jeehoonkang:
@Vtec234 `finalize()` doesn't need to drop the element.  More concretely, here's an example implementation of `finalize()` when an element can be inserted to multiple intrusive lists.  Say there's a counter that represents how many times an element is inserted in intrusive lists, and each `finalize()` decrements the counter.  If the counter reaches zero, then `defer_destroy(element)` is called.

jeehoonkang:
ping? @Vtec234 @stjepang

stjepang:
I don't think this matters much because the intrusive linked list is not a publicly available API. Can we just say that an element cannot be in two intrusive linked lists and call it a day? Does anyone else feel like the problem we're solving here does not really exist? (: 

As a side note, to be honest, I personally find it easier to implement a linked list from scratch than learn how to use the intrusive linked list API correctly. It's a kind of leaky abstraction that adds complexity instead of simplifying the original problem. *shrug*

jeehoonkang:
@stjepang While it's not exposed as public, it's an internal API whose bug is still better to be fixed.  A buggy API hinders future evolution, e.g., I'm experimenting with putting `Local` in two intrusive lists, which is impossible in the current API.

I know intrusive list is a difficult topic, and I agree with you in that it's worth exploring with other data structures (e.g. array as mentioned in `internal.rs` comment).  But I disagree with you on the complexity of our intrusive list abstraction.  Yes it's leaky, but still it is much simpler than implementing one in `internal.rs`, for example.  `Finalize()` is practically the only API the user should be concerned with, whose safety condition thus should be thoroughly documented.

jeehoonkang:
I'm merging this, because I need this in my work on Epoch + HP: https://github.com/jeehoonkang/crossbeam/tree/pebr

bors r+

PR #378, opened 2019-05-17
add BSD license to crossbeam-queue metadata

pacman82:
The BSD License is mentioned in the code, but not visible in the metadata. I want to use this crate at work and fixing the metadata would help.

### DIFF
--- a/crossbeam-queue/Cargo.toml
+++ b/crossbeam-queue/Cargo.toml
@@ -6,7 +6,7 @@ name = "crossbeam-queue"
 # - Create "crossbeam-queue-X.Y.Z" git tag
 version = "0.1.2"
 authors = ["The Crossbeam Project Developers"]
-license = "MIT OR Apache-2.0 AND BSD-2-Clause"
+license = "MIT/Apache-2.0 AND BSD-2-Clause"
 readme = "README.md"
 repository = "https://github.com/crossbeam-rs/crossbeam"
 homepage = "https://github.com/crossbeam-rs/crossbeam/tree/master/crossbeam-utils"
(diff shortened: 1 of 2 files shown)


crossbeam-queue/Cargo.toml
```diff
@@ -6,7 +6,7 @@ name = "crossbeam-queue"
 # - Create "crossbeam-queue-X.Y.Z" git tag
 version = "0.1.2"
 authors = ["The Crossbeam Project Developers"]
-license = "MIT/Apache-2.0"
+license = "MIT OR Apache-2.0 AND BSD-2-Clause"
```
stjepang: Is this the right way of specifying the license?
I wonder if it could be misunderstood as `MIT OR (Apache-2.0 AND BSD-2-Clause)`. :)

We should probably also add the same license to `crossbeam-channel`.

pacman82: Good catch. I have to agree. Taken operator priority into account it probably is going to be understood this way. Yet `cargo` does not support brackets yet. So, I do not know, what todo about that. It still seems an improvement to me to at least mention all three licenses, but I am far from being a lawyer.

pacman82: hm, how do you feel about `MIT/Apache-2.0 AND BSD-2-Clause`. `/` is actually ambiguous, but I still think people will get the right idea looking at the expression.

stjepang: That looks good to me! Could you also add the same to `crossbeam-channel`?

pacman82: Done

stjepang:
Thanks!

bors r+

PR #386, opened 2019-05-28
Conversion of Parker/Unparker into/from a raw pointer

stjepang:
@carllerche Can you review this PR and confirm the API fits your use case?

### DIFF
--- a/crossbeam-utils/src/sync/parker.rs
+++ b/crossbeam-utils/src/sync/parker.rs
@@ -1,8 +1,8 @@
 use std::fmt;
 use std::marker::PhantomData;
-use std::sync::{Arc, Condvar, Mutex};
 use std::sync::atomic::AtomicUsize;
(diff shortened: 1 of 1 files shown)


carllerche:
@stjepang I believe this does, but won't get back to this until later this week. Feel free to ship before then, or you can wait until I can test it.

carllerche:
I used the impl [here](https://github.com/tokio-rs/tokio/blob/std-future/tokio-executor/src/park.rs#L229-L268) (untested as of yet). wdyt?

dekellum:
Has it been sufficiently tested by now, on tokio master since June 24?   I'm asking only because the patch for this branch there (master branch tokio-executor, for crossbeam-utils only) is a bit of a liability for future updates to crossbeam.

jeehoonkang:
I don't know whether it's still relevant to Tokio, but I think the PR itself is okay.  If no one is objecting, I'll merge this 3 days later.

PR #391, opened 2019-06-12
Fix a bug in AtomicCell with garbage padding

### LINKED ISSUE #388: Issue with AtomicCell containing struct with padding bytes
Edit: root cause here: https://github.com/crossbeam-rs/crossbeam/issues/388#issuecomment-499308739

----

The following code appears to deadlock on Windows only:

```rust
fn main() {
    let (tx, rx) = crossbeam_channel::unbounded();
    std::thread::spawn(move || {
        loop {
            tx.send(()).unwrap();
            std::thread::sleep(std::time::Duration::from_millis(50));
        }
    });

    let update_tick = crossbeam_channel::tick(std::time::Duration::from_secs(10));
    let mut data_i = 0;
    loop {
        crossbeam_channel::select! {
            recv(update_tick) -> tick => {
                eprintln!("tick: {:?}", tick);
            },

            recv(rx) -> data => match data {
                Ok(data) => {
                    data_i += 1;
                    if data_i % 50== 0 {
                        dbg!(data_i);
                    }
                    std::thread::sleep(std::time::Duration::from_millis(1));
                }
                Err(_) => break,
            },
        }
    }
}
```

It should print something like:

```
[src/main.rs:22] data_i = 50
[src/main.rs:22] data_i = 100
[src/main.rs:22] data_i = 150
tick: Ok(Instant { tv_sec: 1023, tv_nsec: 433738600 })
[src/main.rs:22] data_i = 200
[src/main.rs:22] data_i = 250
[src/main.rs:22] data_i = 300
```
But I see no further output after `[src/main.rs:22] data_i = 150` when running on Windows. It works fine on Linux.

Tested with: `stable-x86_64-pc-windows-msvc` (`rustc 1.35.0 (3c235d560 2019-05-20)`).

stjepang:
Closes #388 

@mchesser Could you perhaps confirm that this resolves the bug for you?

### DIFF
--- a/crossbeam-utils/src/atomic/atomic_cell.rs
+++ b/crossbeam-utils/src/atomic/atomic_cell.rs
@@ -2,7 +2,6 @@ use core::cell::UnsafeCell;
 use core::fmt;
 use core::mem;
 use core::ptr;
-use core::slice;
 use core::sync::atomic::{self, AtomicBool, AtomicUsize, Ordering};
 
 use Backoff;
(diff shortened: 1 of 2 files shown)


mchesser:
Works for me.

PR #392, opened 2019-06-12
Switch to edition 2018

### LINKED ISSUE #389: CI fails: `dyn` required in `Box<FnMut() + Send + 'env>`
The current nightly requires `dyn` before `FnMut()` in `Box<FnMut() + Send + 'env>`: https://travis-ci.org/crossbeam-rs/crossbeam/jobs/540078416#L494

But Rust 1.26.0 (which we support) doesn't accept `dyn`.

We can avoid this problem by either (1) duplicating the code, or (2) bump the minimum supported rust version.

stjepang:
Closes #389

### DIFF
--- a/crossbeam-channel/benchmarks/chan.rs
+++ b/crossbeam-channel/benchmarks/chan.rs
@@ -1,6 +1,4 @@
-#[macro_use]
-extern crate chan;
-use crossbeam;
+use chan::chan_select;
 
 mod message;
 
--- a/crossbeam-channel/benchmarks/chan.rs
+++ b/crossbeam-channel/benchmarks/chan.rs
@@ -1,6 +1,6 @@
 #[macro_use]
 extern crate chan;
-extern crate crossbeam;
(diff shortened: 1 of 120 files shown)


crossbeam-channel/benchmarks/chan.rs
```diff
@@ -9,7 +7,7 @@ const THREADS: usize = 4;
 
 fn new<T>(cap: Option<usize>) -> (chan::Sender<T>, chan::Receiver<T>) {
     match cap {
-        None => chan::async(),
+        None => chan::r#async(),
```
Vtec234: Long term, should this name be changed not to clash with a keyword?

stjepang: The `chan` crate is deprecated so it probably won't be changed.

stjepang:
@Vtec234 Yeah, I'm considering not switching `crossbeam-deque` to edition 2018 for that reason...

PR #396, opened 2019-06-24
feat: impl {Ref,}UnwindSafe for AtomicCell

Marwes:
This should be correct since (I assume) `AtomicCell` never panics in a method and
will therefore always be correct (under any unrelated panics)

### DIFF
--- a/crossbeam-utils/src/atomic/atomic_cell.rs
+++ b/crossbeam-utils/src/atomic/atomic_cell.rs
@@ -4,6 +4,9 @@ use core::mem;
 use core::ptr;
 use core::sync::atomic::{self, AtomicBool, AtomicUsize, Ordering};
 
+#[cfg(feature = "std")]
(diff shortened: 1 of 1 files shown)


stjepang:
>  (I assume) `AtomicCell` never panics in a method

Unfortunately, that is not true -- the comparator can fail in methods `compare_and_swap` and `compare_exchange` :(

Marwes:
Right. It could be an issue where a panic invalidates a lock if one is held when a comparison panics. Lock-free cells should be ok since the cell can always be used after a panic(?)

stjepang:
Ah, you're absolutely right -- I didn't think deeply enough about this issue. A panic cannot occur while a lock is held and therefore can't invalidate the state of `AtomicCell`.

So adding `UnwindSafe` and `RefUnwindSafe` implementations is perfectly fine. :)

bors r+

Issue #201, opened 2018-11-05
Easy issue: port tests from Go to Rust


stjepang:
In the following file we have already ported some tests from Go to Rust:
https://github.com/crossbeam-rs/crossbeam-channel/blob/master/tests/golang.rs

The idea is to blindly port all the tests for Go channels line by line, which will give us additional confidence in the correctness of crossbeam-channel.

In the test file, you can find stubs with links to the original tests written in Go and todos that need to be filled out.

Porting is a relatively straightforward task, so if you're looking for an approachable issue to tackle, this is the one! :) Also, you don't have to port everything at once - we'll do it incrementally in small pieces. Port as much code as you feel like!

byusti:
Hi, if this is an issue that still needs attention, I would like to help out.

stjepang:
@byusti That'd be great, thank you :) You can look at the following file: https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-channel/tests/golang.rs

There's a bunch of TODOs there with links to tests written in Go that should be ported to Rust.

Horki:
@stjepang Hi, I have a init version for 'select3', could you please review it? https://github.com/crossbeam-rs/crossbeam/pull/575 
- also would like help with 'test_panic'

PR #410, opened 2019-08-15
Add more golang channel tests (closedchan)

fyrchik:
This addresses #201 , specifically tests from here https://github.com/golang/go/blob/master/test/closedchan.go

I am yet new to Rust.
Currently there are 3 issues:
1) `send` implementation of `Chan` can panic while holding the `Mutex`. This is bad if we want to test for panic to occur. Do you have any suggestions on where should I look?
My implementation in comments works in my case, but some other tests are started to take more than 60 seconds.
2) `send` in `select!` for closed channels blocks, because it returns sender for `bounded(0)`.
In `Go` panic is expected. I can implement `SenderFlavour` which always panics, but it will impose runtime overhead for other code, so it doesn't worth it.
Can it be done other way?
3) problem with packing closures in generic vector (I will deal with it later)

Please let me know your opinion.

### DIFF
--- a/crossbeam-channel/tests/golang.rs
+++ b/crossbeam-channel/tests/golang.rs
@@ -53,7 +53,7 @@ impl<T> Chan<T> {
             .expect("sending into closed channel")
             .clone();
         let _ = s.send(msg);
-        
+
         // FIXME
         // let guard = self
         //     .inner
@@ -72,15 +72,10 @@ impl<T> Chan<T> {
     }
 
     fn try_send(&self, msg: T) -> bool {
-        let guard = self
-            .inner
-            .lock()
-            .unwrap();
+        let guard = self.inner.lock().unwrap();
 
         match guard.s.as_ref() {
-            Some(ss) => {
-                ss.clone().try_send(msg).is_ok()
-            },
+            Some(ss) => ss.clone().try_send(msg).is_ok(),
             None => {
                 std::mem::drop(guard);
                 panic!("sending into closed channel")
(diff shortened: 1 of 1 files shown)


crossbeam-channel/tests/golang.rs
```diff
@@ -1424,7 +1452,274 @@ mod chan_test {
...
+    impl GoChan for SChan {
+        fn send(&self, n: i32) {
+            select! {
+                send(self.inner.tx(), n) -> _ => {}
+            }
+        }
+        fn nbsend(&self, n: i32) -> bool {
+            select! {
+                default => { false },
+                send(self.inner.tx(), n) -> _ => { true },
```
stjepang:
```suggestion
                send(self.inner.tx(), n) -> res => { res.unwrap(); true },
```

crossbeam-channel/tests/golang.rs
```diff
@@ -1424,7 +1452,274 @@ mod chan_test {
...
+    }
+
+    struct SChan {
+        inner: Chan<i32>,
+    }
+
+    impl GoChan for SChan {
+        fn send(&self, n: i32) {
+            select! {
+                send(self.inner.tx(), n) -> _ => {}
```
stjepang:
```suggestion
                send(self.inner.tx(), n) -> res => res.unwrap()
```

ghost (commented):
Hi @fyrchik! Thank you for submitting this PR and sorry for the late review!

I think what you were missing here is to unwrap results in `send` and `nbsend` so that we panic when sending fails due to the channel being disconnected. Could you try the changes suggested below?

PR #411, opened 2019-08-27
queue/SegQueue: simplify len calculation

BusyJay:


BusyJay:
It seems a new version of cfg-if breaks the CI instead of the patch.

Vtec234:
This is a nice optimisation and it seems correct in that assuming `tail >= head`, Z3 tells me that the PR'd implementation is exactly equivalent to the [current one](https://github.com/crossbeam-rs/crossbeam/blob/e97836c17f06f4ceae8b948a60cc0ba96bd73350/crossbeam-queue/src/seg_queue.rs#L402) on both 32-bit and 64-bit systems, *except* for when `tail == 2**32 - 2` or `tail == 2**64 - 2`, respectively. I have no idea how the queue works - is this something to worry about? CC @stjepang @taiki-e 

P.S. Sorry for taking a year to review..

Issue #412, opened 2019-08-27
CI fails with `unresolved import `self::std::hint`


jeehoonkang:
https://travis-ci.org/crossbeam-rs/crossbeam/jobs/577310627#L270

I checked in my local machine that it happens in the current master for Rust 1.26.0.

Vtec234:
Doe we still support 1.26.0?

PR #413, opened 2019-09-01
Prevent drift in tick delivery (fix #406)

### LINKED ISSUE #406: channel::tick() behavior questions
Hello !
When using `channel::tick::recv()` I expected it to compute its time to sleep with some kind of `Instant` saved at the time of the `Channel` instantiation.
Instead the `Instant` used is the one taken from the last message received, which has two effects :
- it guarantees that the specified duration at least elapses between two messages received
- it leads to a small shift from the initial `Channel` instantiation Instant, mostly due to the job scheduler not waking up the thread necessarily on time

If this was intended originally, then some documentation about the second effect is necessary.
If this is a bug, I would be happy to send a PR. However we must keep in mind that some users might rely on the first effect above.

_Also, I think the `channel::tick::recv` should not have to loop around a `thread::sleep` as the latter already guarantees to sleep **at least** the provided duration, which leaves room for simplifying this file a lot._

drrlvn:
Calculate the next delivery time based on the previous one instead of the time the thread awoke, preventing a small but noticeable drift from accumulating.

### DIFF
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -272,9 +272,9 @@ pub fn never<T>() -> Receiver<T> {
 /// assert!(eq(r.recv().unwrap(), start + ms(200)));
 /// assert!(eq(Instant::now(), start + ms(600)));
 ///
-/// // This message was sent 700 ms from the start and received 700 ms from the start.
-/// assert!(eq(r.recv().unwrap(), start + ms(700)));
-/// assert!(eq(Instant::now(), start + ms(700)));
+/// // This message was sent 300 ms from the start and received 600 ms from the start.
+/// assert!(eq(r.recv().unwrap(), start + ms(300)));
+/// assert!(eq(Instant::now(), start + ms(600)));
 /// ```
 pub fn tick(duration: Duration) -> Receiver<Instant> {
     Receiver {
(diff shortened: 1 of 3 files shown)


drrlvn:
I fixed the tests so they now pass. I don't know why the CI fails on 1.26/1.28, maybe it's a dependency issue. This is the error:
```
error[E0432]: unresolved import `self::std::hint`

  --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/lazy_static-1.4.0/src/inline_lazy.rs:13:16

   |

13 | use self::std::hint::unreachable_unchecked;

   |                ^^^^ Could not find `hint` in `std`

error: aborting due to previous error
```

drrlvn:
> I don't know why the CI fails on 1.26/1.28, maybe it's a dependency issue. 

Happens on `master` as well, see #412.

stjepang:
Hey, thank you for submitting this PR and apologies for coming to it so late! :( I'm assuming you already forgot about this code so I submitted a new PR: #456

I think `try_recv()` should not change - the drift there is intended and this is exactly how tickers work in Go, too. However, we do need changes in `recv()` - instead of sleeping and then attempting to receive a message, we should receive it first and *then* sleep.

drrlvn:
@stjepang OK, great, I'll close this PR then. Thanks!

PR #414, opened 2019-09-07
Remove ArrayVec dependency

Shnatsel:
`ArrayVec` is used in just one place, and even there crossbeam uses a small part of its API surface. This PR replaces ArrayVec with 100% safe code using a plain old array and ditches the ArrayVec dependency altogether.

This removes 170 unsafe expressions from the dependency tree according to [cargo-geiger](https://github.com/anderejd/cargo-geiger). With less unsafe code to worry about there's less chance of memory safety bugs, and safety/security audits will be cheaper and easier to perform.

An unfortunate casualty of this change is `#[derive(Debug)]` on internal structures because there is no Debug implementation for `[T; 64]` - these impls are only generated up to length of 32. It will be possible to re-enable it once const generics are stabilized. For now they're implemented manually.

`cargo bench` results are unaffected - or might be within experimental noise of 2% or so, hard to tell without statistical analysis on benchmark results.

### DIFF
--- a/crossbeam-epoch/src/internal.rs
+++ b/crossbeam-epoch/src/internal.rs
@@ -42,7 +42,6 @@ use core::ptr;
 use core::sync::atomic;
 use core::sync::atomic::Ordering;
 
-use arrayvec::ArrayVec;
 use crossbeam_utils::CachePadded;
 
 use atomic::{Shared, Owned};
@@ -60,10 +59,11 @@ const MAX_OBJECTS: usize = 64;
 const MAX_OBJECTS: usize = 4;
 
 /// A bag of deferred functions.
-#[derive(Default, Debug)]
+// can't #[derive(Debug)] because Debug is not implemented for arrays 64 items long
 pub struct Bag {
     /// Stashed objects.
-    deferreds: ArrayVec<[Deferred; MAX_OBJECTS]>,
+    deferreds: [Deferred; MAX_OBJECTS],
+    len: usize
 }
 
 /// `Bag::try_push()` requires that it is safe for another thread to execute the given functions.
@@ -77,7 +77,7 @@ impl Bag {
 
     /// Returns `true` if the bag is empty.
     pub fn is_empty(&self) -> bool {
-        self.deferreds.is_empty()
+        self.len == 0
     }
 
     /// Attempts to insert a deferred function into the bag.
@@ -89,7 +89,13 @@ impl Bag {
     ///
     /// It should be safe for another thread to execute the given function.
     pub unsafe fn try_push(&mut self, deferred: Deferred) -> Result<(), Deferred> {
-        self.deferreds.try_push(deferred).map_err(|e| e.element())
+        if self.len < MAX_OBJECTS {
+            self.deferreds[self.len] = deferred;
+            self.len += 1;
+            Ok(())
+        } else {
+            Err(deferred)
+        }
     }
 
     /// Seals the bag with the given epoch.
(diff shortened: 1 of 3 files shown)


Shnatsel:
CI failed with "error: Could not compile `lazy_static`." - apparently 1.26 is too old to compile `lazy_static!`. Nightly has passed.

jeehoonkang:
I'm personally preferring to use `ArrayVec`, because it's specifically designed for such a scenario as `Bag`.  The number of `unsafe` expressions of Crossbeam is not a big concern, I think, as (1) `ArrayVec` is an established project, and (2) its safety is as much scrutinized as Crossbeam's.

Shnatsel:
@Lokathor I've implemented Debug manually

Shnatsel:
@jeehoonkang the selling point of Rust is that you can trust the compiler to verify safety for you instead of trusting humans to get it right.

ArrayVec might be sound now (although I haven't checked, and you wouldn't believe what I've found in some other popular Rust crates recently), but having an unsafe dependency is not just about now, it's a continuous burden. Not only you have to audit all unsafe blocks in the the current version, you also have to commit to auditing all unsafe blocks in _all future versions._

Between taking on the burden of manually auditing all the unsafe in all future versions of ArrayVec and rewriting this small bit as safe code without even incurring a performance penalty, the choice is pretty obvious.

crossbeam-epoch/src/internal.rs
```diff
@@ -98,15 +103,53 @@ impl Bag {
...
+             Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func),
+             Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func),
+             Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func),
+             Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func),
+             Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func),
+             Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func),
+             Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func),
+             Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func),
+             Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func), Deferred::new(no_op_func)]
```
jadbox: I'm curious on two things: why are exactly this many Deferred objects created here and is there a static way to populate the list than to hardcode each new obj construction (to make it DRY'er)?

bluss: You could definitely do this with a macro, maybe just not in the most natural way. For example you could hardcode 0-10 fold repetition and write 64-fold repetition as repeating by 6 * 10 + 4 :slightly_smiling_face:

crossbeam-epoch/src/internal.rs
```diff
@@ -98,15 +103,53 @@ impl Bag {
...
+
 impl Drop for Bag {
     fn drop(&mut self) {
         // Call all deferred functions.
-        for deferred in self.deferreds.drain(..) {
+        for i in 0..self.len {
+            let no_op = Deferred::new(no_op_func);
+            let deferred = mem::replace(&mut self.deferreds[i], no_op);
             deferred.call();
         }
```
bluss: This can be an iterator over `&mut self.deferreds[..]` I think

PR #416, opened 2019-09-09
Require three epoch advancements for deallocation 

### LINKED ISSUE #238: Use after free in Michael-Scott queue?
Queues that suffer from the problem:

* `MsQueue`
* `SegQueue`
* `channel::unbounded()` (see #237, thanks to @Pslydhh)
* `Queue` in `crossbeam-epoch`

---------

Consider the following scenario in `MsQueue`...

Suppose the current epoch is `E`. The queue is empty, meaning `head` and `tail` are pointing to the sentinel node (let's call it `node0`).
In other words, `head = node0`, `tail = node0`, `node0.next = null`.

The following happens next...

Thread `T1` gets pinned in epoch `E` and pushes a new value into the queue by allocating a new node `node1` and then it does `node0.next.cas(null, node1)`. The next operation it needs to do is `tail.cas(node0, node1)`, but the thread gets yielded first so `tail` is still `node0`.

The queue is now: `head = node0`, `tail = node0`, `node0.next = node1`, `node1.next = null`.

Thread `T2` gets pinned in epoch `E` and pops a value from the queue with `head.cas(node0, node0.next)`. The CAS succeeds and `head` becomes `node1`. Then we call `defer_destroy(node0)`, which means `node0` becomes garbage marked with epoch `E`. The thread then exits.

The queue is now: `head = node1`, `tail = node0`, `node0.next = node1`, `node1.next = null`.

The global epoch gets advanced to `E+1`.

Thread `T3` gets pinned in epoch `E+1` and attempts to push a new value into the queue. It allocates a new node called `node2`, loads the `tail` pointer and gets `node0`. Then, it would attempt to do `node0.next.cas(null, node2)`, but first it yields...

Thread `T1` resumes, sets `tail` to `node1`, and exits.

The queue is now: `head = node1`, `tail = node1`, `node1.next = null`.

The global epoch is advanced to `E+2`. Garbage from epoch `E` is collected and so `node0` gets deallocated.

Thread `T3` resumes and attempts to continue with `node0.next.cas(null, node2)`, but `node0` has been destroyed. This is use-after-free!

-----------

To solve this problem, I propose we just relax the invariant used by `defer`.

Currently, the invariant is: We may use `defer` to destroy an object only if it will become inaccessible when all active guards get dropped.

I propose we change it to: We may use `defer` to destroy an object only if it will become inaccessible when all active guards and future guards overlapping with them get dropped.

In other words, instead of saying "call `defer` if the object is inaccessible" we'd say "call `defer` if it is inaccessible or if it will become inaccessible before all active guards get dropped".

The PR fixing this would simply change the `global_epoch - garbage_epoch >= 2` invariant to `global_epoch - garbage_epoch >= 3`.

cc @Pslydhh @jeehoonkang

tomtomjhj:
This patch implements the fix discussed in https://github.com/crossbeam-rs/crossbeam/issues/238#issuecomment-525795655. 

* tag the Bag with the local epoch E
* free at E+3

### DIFF
--- a/crossbeam-epoch/src/internal.rs
+++ b/crossbeam-epoch/src/internal.rs
@@ -227,9 +227,9 @@ unsafe impl Sync for SealedBag {}
 impl SealedBag {
     /// Checks if it is safe to drop the bag w.r.t. the given global epoch.
     fn is_expired(&self, global_epoch: Epoch) -> bool {
-        // A pinned participant can witness at most one epoch advancement. Therefore, any bag that
-        // is within one epoch of the current one cannot be destroyed yet.
-        global_epoch.wrapping_sub(self.epoch) >= 2
+        // A pinned participant can witness at most two epoch advancement. Therefore, any bag that
+        // is within two epoch of the current one cannot be destroyed yet.
+        global_epoch.wrapping_sub(self.epoch) >= 3
     }
 }
 
@@ -260,12 +260,8 @@ impl Global {
     }
 
     /// Pushes the bag into the global queue and replaces the bag with a new empty bag.
-    pub fn push_bag(&self, bag: &mut Bag, guard: &Guard) {
+    pub fn push_bag(&self, bag: &mut Bag, epoch: Epoch, guard: &Guard) {
         let bag = mem::replace(bag, Bag::new());
-
-        atomic::fence(Ordering::SeqCst);
-
-        let epoch = self.epoch.load(Ordering::Relaxed);
         self.queue.push(bag.seal(epoch), guard);
     }
 
@@ -429,7 +425,8 @@ impl Local {
         let bag = &mut *self.bag.get();
 
         while let Err(d) = bag.try_push(deferred) {
-            self.global().push_bag(bag, guard);
+            let epoch = self.epoch.load(Ordering::Relaxed).unpinned();
+            self.global().push_bag(bag, epoch, guard);
             deferred = d;
         }
     }
@@ -438,7 +435,8 @@ impl Local {
         let bag = unsafe { &mut *self.bag.get() };
 
         if !bag.is_empty() {
-            self.global().push_bag(bag, guard);
+            let epoch = self.epoch.load(Ordering::Relaxed).unpinned();
+            self.global().push_bag(bag, epoch, guard);
         }
 
         self.global().collect(guard);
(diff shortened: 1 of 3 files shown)


jeehoonkang:
@tomtomjhj would you please rebase it on the current master?  We recently fixed CI failures.

crossbeam-epoch/src/internal.rs
```diff
@@ -389,38 +387,47 @@ impl Local {
...
+                    // We add a compiler fence to make it less likely for LLVM to do something wrong
+                    // here.  Formally, this is not enough to get rid of data races; practically, it
+                    // should go a long way.
+                    atomic::compiler_fence(Ordering::SeqCst);
+                } else {
+                    self.epoch.store(new, Ordering::Relaxed);
+                    atomic::fence(Ordering::SeqCst);
+                }
+
+                // Now we validate that the value we read from the global epoch is not stale.
```
stjepang: What is the motivation for validation phase? I'm afraid that this loop might slow things down a bit.

Could you run `cargo bench` and see if this affects performance?

tomtomjhj: It's for helping the global epoch advancement in general. It doesn't seem to have a meaningful effect on the performance.

* master:
```
test multi_alloc_defer_free  ... bench:   4,135,110 ns/iter (+/- 566,246)
test multi_defer             ... bench:   2,154,671 ns/iter (+/- 212,984)
test single_alloc_defer_free ... bench:          43 ns/iter (+/- 0)
test single_defer            ... bench:          16 ns/iter (+/- 0)

test multi_flush  ... bench:  22,928,416 ns/iter (+/- 1,546,218)
test single_flush ... bench:         136 ns/iter (+/- 1)

test multi_pin  ... bench:   4,179,789 ns/iter (+/- 167,156)
test single_pin ... bench:           5 ns/iter (+/- 0)
```

* free-after-3
```
test multi_alloc_defer_free  ... bench:   4,176,331 ns/iter (+/- 513,890)
test multi_defer             ... bench:   2,179,475 ns/iter (+/- 197,665)
test single_alloc_defer_free ... bench:          43 ns/iter (+/- 1)
test single_defer            ... bench:          16 ns/iter (+/- 0)

test multi_flush  ... bench:  21,997,370 ns/iter (+/- 3,973,497)
test single_flush ... bench:         135 ns/iter (+/- 0)

test multi_pin  ... bench:   4,173,010 ns/iter (+/- 389,623)
test single_pin ... bench:           5 ns/iter (+/- 0)

```
* free-after-3 without validation
```
test multi_alloc_defer_free  ... bench:   4,140,581 ns/iter (+/- 426,444)
test multi_defer             ... bench:   2,160,064 ns/iter (+/- 196,742)
test single_alloc_defer_free ... bench:          43 ns/iter (+/- 0)
test single_defer            ... bench:          16 ns/iter (+/- 0)

test multi_flush  ... bench:  22,800,624 ns/iter (+/- 2,119,408)
test single_flush ... bench:         136 ns/iter (+/- 1)

test multi_pin  ... bench:   4,228,682 ns/iter (+/- 380,561)
test single_pin ... bench:           5 ns/iter (+/- 0)
```

jeehoonkang:
bors r-

@tomtomjhj and I figured out that actually #238 is not fixed by this commit...  More discussion will come by @tomtomjhj soon.

jeehoonkang:
Btw, I believe it's a good patch, though.  I'm disapproving this PR to notify that it doesn't fix #238.

tomtomjhj:
https://github.com/crossbeam-rs/crossbeam/issues/238#issuecomment-556058428

xacrimon:
Status on this?

tomtomjhj:
@xacrimon I made a new PR that fixes the queue itself #466.

stjepang:
@jeehoonkang Do you think this needs to be merged? Let's make a decision.

jeehoonkang:
I'd like to merge it because this PR streamlines the correctness argument of crossbeam-epoch.

bors r+

jeehoonkang:
CI test fails with an orthogonal error: https://github.com/crossbeam-rs/crossbeam/pull/416/checks?check_run_id=690982289#step:6:15

stjepang:
Let's merge manually.

PR #393, opened 2019-06-13
Allow warnings on nightly

stjepang:
Closes #389

PR #428, opened 2019-10-15
Fix breakage with nightly feature

taiki-e:
`target_has_atomic = "cas"` was removed in https://github.com/rust-lang/rust/pull/65214.

Fixes https://github.com/rust-lang/rust/issues/65424

### DIFF
--- a/crossbeam-epoch/src/lib.rs
+++ b/crossbeam-epoch/src/lib.rs
@@ -73,10 +73,7 @@ cfg_if! {
     }
 }
 
-#[cfg_attr(
-    feature = "nightly",
-    cfg(all(target_has_atomic = "cas", target_has_atomic = "ptr"))
-)]
+#[cfg_attr(feature = "nightly", cfg(target_has_atomic = "ptr"))]
 cfg_if! {
     if #[cfg(any(feature = "alloc", feature = "std"))] {
         extern crate crossbeam_utils;
--- a/crossbeam-skiplist/src/lib.rs
+++ b/crossbeam-skiplist/src/lib.rs
@@ -18,10 +18,7 @@ cfg_if! {
     }
 }
 
-#[cfg_attr(
-    feature = "nightly",
-    cfg(all(target_has_atomic = "cas", target_has_atomic = "ptr"))
-)]
+#[cfg_attr(feature = "nightly", cfg(target_has_atomic = "ptr"))]
 cfg_if! {
     if #[cfg(any(feature = "alloc", feature = "std"))] {
         extern crate crossbeam_epoch as epoch;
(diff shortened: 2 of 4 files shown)


jeehoonkang:
Thanks!

bors r+

taiki-e:
@jeehoonkang  Since the build on Rust 1.26 is broken, I think we need to merge this and #429 at the same time.

jeehoonkang:
Thanks for updating the dependency.

bors r+

jeehoonkang:
CI fails with permission error: https://travis-ci.org/crossbeam-rs/crossbeam/jobs/598198019#L203

jeehoonkang:
Thinking again, downgrading only for CI looks suspicious.  Then CI can only tell a modified Crossbeam (with different dependency) compiles for Rust 1.26, but not Crossbeam as-is.  I think it may not be what we want...  Would you introduce how Rayon deal with such a situation?

taiki-e:
> Thinking again, downgrading only for CI looks suspicious. Then CI can only tell a modified Crossbeam (with different dependency) compiles for Rust 1.26, but not Crossbeam as-is. I think it may not be what we want... Do you know how Rayon deals with such a situation?

I think the following comments accurately explain the current situation: 

https://github.com/rayon-rs/rayon/issues/586#issuecomment-464032529

> The state of the ecosystem is that, for better or worse, any less-than-latest Rust version is not supported in general, and in practice absolutely unusable without an old `Cargo.lock`. Debian's Rust is just totally useless.

Also, they recently bumped the minimum version to 1.28, so I think we can drop 1.26 and 1.27's support.

taiki-e:
@jeehoonkang CI passed.

jeehoonkang:
I'm fine with this change, I'd like to wait for another approval because it's changing the CI policy.  @Vtec234 @stjepang Any thoughts?

stjepang:
Let's drop support for 1.26 and 1.27.

taiki-e:
Added commit to drop support for 1.26 and 1.27. Also, since the minimum version is 1.28, we can revert #393 and fix `bare_trait_objects` warnings.

taiki-e:
Oh... `deprecated` warnings... 
```text
error: use of deprecated item 'std::mem::uninitialized': use `mem::MaybeUninit` instead
  --> crossbeam-epoch/src/deferred.rs:39:38
   |
39 |                 let mut data: Data = mem::uninitialized();
   |                                      ^^^^^^^^^^^^^^^^^^
   |
   = note: `-D deprecated` implied by `-D warnings`
```

I will use `#[allow(deprecated)]` because the alternative requires 1.36.

jeehoonkang:
Thanks!

bors retry

jeehoonkang:
bors retry

PR #429, opened 2019-10-15
Limit lazy_static version

### LINKED ISSUE #412: CI fails with `unresolved import `self::std::hint`

jeehoonkang:
Closes #412 

It's necessary to accept #428, which is fixing a breakage of Rust ecosystems...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ std = [
 alloc = ["crossbeam-epoch/alloc", "crossbeam-utils/alloc"]
 
 [dependencies]
-cfg-if = "0.1.2"
+cfg-if = "~0.1.9"
 
 [dependencies.crossbeam-channel]
 version = "0.3.9"
--- a/crossbeam-epoch/Cargo.toml
+++ b/crossbeam-epoch/Cargo.toml
@@ -23,7 +23,7 @@ alloc = ["crossbeam-utils/alloc"]
 sanitize = [] # Makes it more likely to trigger any potential data races.
 
 [dependencies]
-cfg-if = "0.1.2"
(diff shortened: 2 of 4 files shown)


jeehoonkang:
Due to its relative urgency, I'm self-approving it.

bors r+

jeehoonkang:
bors retry

Cargo.toml
```diff
@@ -32,7 +32,7 @@ std = [
 alloc = ["crossbeam-epoch/alloc", "crossbeam-utils/alloc"]
 
 [dependencies]
-cfg-if = "0.1.2"
+cfg-if = "~0.1.9"
```
taiki-e: Personally, I prefer to downgrade the dependency version with ci instead of limiting the version. ([rayon did it before.](https://github.com/rayon-rs/rayon/commit/b21969416c36be0ad033452cebfd9a66db3da7f7#diff-354f30a63fb0907d4ad57269548329e3L20-L22))

jeehoonkang:
Closing.  Dependency version will be handled in #428.

Cargo.toml
stjepang: @taiki-e Would you be willing to submit a PR taking a similar strategy as in Rayon?

taiki-e: The second commit of #428 is doing that: https://github.com/crossbeam-rs/crossbeam/pull/428/commits/b2e6cf83366512d1d0e369b757b855d95887e91a

PR #430, opened 2019-10-16
Cheat around the unsafe_code lint in the select macro

### LINKED ISSUE #405: Unable to use select macro in #![forbid(unsafe_code)] environments
## Description

I am using `#![forbid(unsafe_code)]` in my crate and for a while I only had a single `recv()` inside of a select macro and everything was compiling fine.

When I added a second `recv()` arm it failed to compile.

I switched to using the `Select` struct for now and it compiles and works properly.

I expected the `select!` macro to only use the safe API of `Select`, not emitting unsafe code inside the macro.

## Example code

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=efee4be440204c5c160827d5fa1de5fb

stjepang:
Fixes #405.

### DIFF
--- a/crossbeam-channel/src/select_macro.rs
+++ b/crossbeam-channel/src/select_macro.rs
@@ -829,8 +829,7 @@ macro_rules! crossbeam_channel_internal {
             let _oper = $crate::internal::select(&mut $sel);
 
             // Erase the lifetime so that `sel` can be dropped early even without NLL.
-            #[allow(unsafe_code)]
-            unsafe { ::std::mem::transmute(_oper) }
+            $crate::internal::unsafe_transmute(_oper)
         };
 
         crossbeam_channel_internal! {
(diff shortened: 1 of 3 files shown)


taiki-e:
As `#[forbid(unsafe_code)]` does not forbid unsafe code generated by macros, I think it is sufficient to just remove `#[allow(unsafe_code)]`.

crossbeam-channel/src/select_macro.rs
```diff
@@ -1212,3 +1207,14 @@ macro_rules! select {
         )
     };
 }
+
+/// Same as `std::mem::transmute`, except the function is not marked with `unsafe`.
```
jeehoonkang: Pedantically speaking, undefined behavior may be invoked because of the absence of `unsafe`.  Not that I'm opposing to this change, but how about saying it's a `HACK`?

jeehoonkang:
Oh, I'm sorry not to notice that #431 tries to replace it.  #431 looks better if it's correct, but I'm not an expert to decide whether it's correct...

taiki-e:
The way used by #431 doesn't seem to work on Rust 1.28. The current behavior is probably because of https://github.com/rust-lang/rust/pull/52467 and seems to be available after 1.29. (fyi: https://github.com/rust-lang/rust/issues/53975)

stjepang:
Closing in favor of #431

PR #431, opened 2019-10-16
Remove #[allow(unsafe_code)] from select! macro

### LINKED ISSUE #405: Unable to use select macro in #![forbid(unsafe_code)] environments

taiki-e:
Replaces #430 

Closes #405

### DIFF
--- a/crossbeam-channel/src/select_macro.rs
+++ b/crossbeam-channel/src/select_macro.rs
@@ -829,7 +829,6 @@ macro_rules! crossbeam_channel_internal {
             let _oper = $crate::internal::select(&mut $sel);
 
             // Erase the lifetime so that `sel` can be dropped early even without NLL.
-            #[allow(unsafe_code)]
             unsafe { ::std::mem::transmute(_oper) }
         };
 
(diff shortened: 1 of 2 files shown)


stjepang:
Hmmm, why is this failing on Travis CI? Does this require a newer version of the compiler?

taiki-e:
The behavior on which this PR depends is introduced in rust-lang/rust#52467 and has been available since 1.29.

taiki-e:
(Personally, I think it's fine to postpone the fix of #405 until the minimum supported version is 1.29+.)

extrawurst:
when is that 

> (Personally, I think it's fine to postpone the fix of #405 until the minimum supported version is 1.29+.)

@taiki-e  when is that gonna be ?

stjepang:
@taiki-e Feel free to merge this PR if you think it's ready!

PR #442, opened 2019-10-27
Do not perform superfluous bounds checks on dropping crossbeam-epoch::Bag

Shnatsel:
Addresses [this comment](https://github.com/crossbeam-rs/crossbeam/pull/414#discussion_r328208426) that I didn't get a chance to act upon before merging #414. 

This way bounds checks should be performed only once. However, this does not measurably impact benchmarks.


jeehoonkang:
Thanks!

bors r+

PR #444, opened 2019-11-04
Prepare for the next release

### LINKED ISSUE #435: crossbeam-channel v0.3.9 fails to compile
After bumping the rustc-ap-* crates in rustfmt, I get
```
error[E0432]: unresolved import `crossbeam_utils::atomic`
 --> /home/r/.cargo/registry/src/github.com-1ecc6299db9ec823/crossbeam-channel-0.3.9/src/flavors/tick.rs:8:22
  |
8 | use crossbeam_utils::atomic::AtomicCell;
  |                      ^^^^^^ could not find `atomic` in `crossbeam_utils`
```
The dependency tree looks as follows:
```
crossbeam-channel v0.3.9
└── crossbeam-utils v0.6.6
    ├── cfg-if v0.1.10
    └── lazy_static v1.4.0
```

### LINKED ISSUE #443: Crossbeam-channel release including clippy fixes
Would it be possible to do a release that includes the below commit? Looks like `0.3.9` just missed it...

https://github.com/crossbeam-rs/crossbeam/commit/c725403c9b4c28e1f5e9a35849c236f1b34f67cc

taiki-e:
This updates version numbers and changelogs.

- crossbeam 0.7.3
- crossbeam-channel 0.4.0
- crossbeam-deque 0.7.2
- crossbeam-epoch 0.8.0
- crossbeam-queue 0.2.0
- crossbeam-utils 0.7.0

cc #435 
cc #443 
cc #446

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ name = "crossbeam"
 # - Update CHANGELOG.md
 # - Update README.md
 # - Create "crossbeam-X.Y.Z" git tag
-version = "0.7.2"
+version = "0.7.3"
 authors = ["The Crossbeam Project Developers"]
 license = "MIT/Apache-2.0"
 readme = "README.md"
@@ -35,7 +35,7 @@ alloc = ["crossbeam-epoch/alloc", "crossbeam-utils/alloc"]
 cfg-if = "0.1.2"
 
 [dependencies.crossbeam-channel]
-version = "0.3.9"
+version = "0.4"
 path = "./crossbeam-channel"
 optional = true
 
@@ -45,17 +45,17 @@ path = "./crossbeam-deque"
 optional = true
 
 [dependencies.crossbeam-epoch]
-version = "0.7.2"
+version = "0.8"
 path = "./crossbeam-epoch"
 default-features = false
 
 [dependencies.crossbeam-queue]
-version = "0.1.0"
+version = "0.2"
 path = "./crossbeam-queue"
 optional = true
 
 [dependencies.crossbeam-utils]
-version = "0.6.6"
+version = "0.7"
 path = "./crossbeam-utils"
 default-features = false
 
--- a/README.md
+++ b/README.md
@@ -96,6 +96,10 @@ Next, add this to your crate:
 extern crate crossbeam;
 ```
 
+## Compatibility
+
+The minimum supported Rust version is 1.28. Any change to this is considered a breaking change.
+
 ## Contributing
 
 Crossbeam welcomes contribution from everyone in the form of suggestions, bug reports,
--- a/crossbeam-epoch/CHANGELOG.md
+++ b/crossbeam-epoch/CHANGELOG.md
@@ -1,3 +1,10 @@
+# Version 0.8.0
+
+- Bump the minimum required version to 1.28.
+- Fix breakage with nightly feature due to rust-lang/rust#65214.
+- Make `Atomic::null()` const function at 1.31+.
+- Bump `crossbeam-utils` to `0.7`.
+
 # Version 0.7.2
 
 - Add `Atomic::into_owned()`.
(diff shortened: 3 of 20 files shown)


Cargo.toml
```diff
@@ -55,7 +55,7 @@ path = "./crossbeam-queue"
 optional = true
 
 [dependencies.crossbeam-utils]
-version = "0.6.6"
+version = "0.6.7"
```
taiki-e: It needs to bump epoch and utils for -Zminimal-versions.

taiki-e:
(I've opened this, but I don't actually have the rights to publish)

crossbeam-epoch/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+# Version 0.7.3
+
+- Bump the minimum required version to 1.28.
```
jeehoonkang: Isn't it considered a breaking change?  (I'm asking this because I don't know the convention of Rust libraries...)

Vtec234: I would say it is a breaking change. In #412 we had this exact issue, where `lazy_static` dropped support for older rustc in a minor version bump, which caused a little trouble.

taiki-e: There are many crates that do this in patch version (lazy_static, [pkg-config](https://github.com/rust-lang/pkg-config-rs/blob/master/CHANGELOG.md#0315---2019-07-25), [serde_derive](https://github.com/serde-rs/serde/releases/tag/v1.0.99), [failure](https://github.com/rust-lang-nursery/failure/pull/319#issuecomment-531891937), etc.), but I think it makes sense to consider it a breaking change.

taiki-e:
Updated version numbers and added explanations about MSRV to readme.

crossbeam-epoch/CHANGELOG.md
jeehoonkang: Since it's a breaking change, I would like to bump the minor version...

Having said that, I also hesitantly agree not to bump the version following the other projects' precedence.  I' marking this issue resolved...

jeehoonkang:
bors r+

As soon as it's merged, I'll release new versions.  Thank you!

taiki-e:
wait, I've updated this PR to bump the minor version in
https://github.com/crossbeam-rs/crossbeam/commit/c76f1451820009bb81bbc0ccf782eb976637667f
.


2019年11月6日(水) 12:13 Jeehoon Kang <notifications@github.com>:

stjepang:
New versions have been published :)

README.md
```diff
@@ -96,6 +96,10 @@ Next, add this to your crate:
 extern crate crossbeam;
 ```
 
+## Compatibility
+
+The minimum supported Rust version is 1.28. Any change to this is considered a breaking change.
```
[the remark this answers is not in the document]
stjepang: Thank you for writing this up! I think we *really* should bring Crossbeam to 1.0 soon. I will follow up with an issue on what changes we need to do to get there.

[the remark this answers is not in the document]
matklad: One additional thing I've noticed: the umbrella crate `crossbeam` did not made a breaking change, it just bumped major versions of it's public deps, which is, dejure, a  breaking change (which doesn't mean it creates problems in practice, only bug reports will tell :) ).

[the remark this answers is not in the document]
stjepang: I will admit it: this was a very rushed and belated release.

Part of the reason is the fact that I *dread* making releases because they're so time-consuming and finicky (update readmes, changelogs, and `Cargo.toml`s for every crate, fix up dependencies, publish to crates.io, publish git tags) and I *always* make mistakes while releasing new versions. I don't think there's been a single release that went smoothly without mistakes.

Grouping all (or most) crossbeam crates into the main crate (kind of like what `tokio` is doing these days) and publishing 1.0 would alleviate this pain for everyone.

PR #447, opened 2019-11-07
Channel list flavor: block thread if backoff expires in read

### LINKED ISSUE #366: Yielding in crossbeam-channel
Hey @papertigers, I just saw your message on IRC, which I am repeating here:

> Have you done any testing with sending values at a decently high throughput?

> I noticed that sending values relatively quickly through a crossbeam channel results in a lot yields and increased CPU usage compared to std::sync::mpsc 

> https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=3677d9c50b85252170dc410137d489dd

> is a small example that shows what I am talking about

> https://gist.github.com/papertigers/0bd33bc0f1241463dcee35763337c579

Thanks for the report! I think it'd be easier to chat either over on GitHub or on our [discord channel](https://discordapp.com/invite/JXYwgWZ) since we have moved off IRC.

My guess what is going on here is that in `std::sync::mpsc`, `Sender` in `std::sync::mpsc` is slower than its corresponding `Receiver` so the messages build up and nobody ever has to block.

On the other hand, it may be that in `crossbeam-channel`, `Sender` is slower than `Receiver` so sometimes the `Receiver` has to block, which is where those yields come from.

Could you perhaps verify if my hypothesis is true?

gterzian:
Maybe FIX #366

### DIFF
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -1376,7 +1376,7 @@ pub unsafe fn write<T>(s: &Sender<T>, token: &mut Token, msg: T) -> Result<(), T
 pub unsafe fn read<T>(r: &Receiver<T>, token: &mut Token) -> Result<T, ()> {
     match &r.flavor {
         ReceiverFlavor::Array(chan) => chan.read(token),
-        ReceiverFlavor::List(chan) => chan.read(token),
+        ReceiverFlavor::List(chan) => chan.read(token, false),
         ReceiverFlavor::Zero(chan) => chan.read(token),
         ReceiverFlavor::After(chan) => {
             mem::transmute_copy::<Result<Instant, ()>, Result<T, ()>>(&chan.read(token))
--- a/crossbeam-channel/src/flavors/list.rs
+++ b/crossbeam-channel/src/flavors/list.rs
@@ -371,7 +371,7 @@ impl<T> Channel<T> {
     }
 
     /// Reads a message from the channel.
-    pub unsafe fn read(&self, token: &mut Token) -> Result<T, ()> {
+    pub unsafe fn read(&self, token: &mut Token, blocking: bool) -> Result<T, ()> {
         if token.list.block.is_null() {
             // The channel is disconnected.
             return Err(());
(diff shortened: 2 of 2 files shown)


gterzian:
So the idea is to remove calls to `yield`, that could occur as part of `read` if the sender is slow to write into the block, by removing the use of a backoff in `Slot`, and instead trying to "reuse" the backoff used in the main loop of `recv`. 

However, then I saw that `start_recv` also uses it's own backoff, so I removed that as well. That is when I think I broke everything, however test are still passing. 

Essentially, trying to use a single `Backoff`, that blocks the thread if completed, as part of `recv`. 
On the other hand, I guess that not spinning/snoozing in `start_recv`, and also `read` when it's used inside `try_recv`, might have other consequences which perhaps do not out-weight the potential benefits of yielding less.

gterzian:
EDIT: added a backoff in `try_recv`. 

I personally think it's better to use a "top-level" backoff in both `recv` and `try_recv`, versus adding backoffs inside the `start_recv` and `read` calls, or other nested calls, since it's clearer to me what is going on if the entire algorithm uses a single loop and backoff, rather then having them being nested. 

However, there might be some good performance reasons for nested those backoffs and introducing some nested spinning.

gterzian:
Ok so there are indeed some tests failing, and for example the `linearizable` one would seem to point out that the `try_recv` returns empty "too early".

gterzian:
Maybe the problem is that this now introduces new blocking of threads,  since previous spinning is now potentially blocking when the backoff completes, and those blocks are not matched by wake-up operations.

I'm thinking of the `drops` test, now failing, for example.

gterzian:
Ok so change of plans, I went for a smaller change only affecting the use of backoff in `read`. 

So instead of a `while cond {snooze}`, it now blocks the thread if the backoff is completed. 

If I understand it correctly, this blocking is ok even in `try_recv`, since the `read` call is nested within a `start_recv`, which guarantees something will be written? 

It would seem better, even in a `try_recv`, to block, versus spinning indefinitely?

gterzian:
Closing dormant PR. I think it's probably not a big problem enough to try to merge this...

PR #449, opened 2019-11-17
fix broken doc link

ericseppanen:
The docs tried to refer to a top-level 'defer' function; instead link to
Guard::defer.


jeehoonkang:
Thanks!

bors r+

PR #450, opened 2019-11-20
Format with rustfmt

taiki-e:
This fix build failure in #449 and #448

It seems old rustfmt did not properly format the module in the macro.


jeehoonkang:
Thanks!  bors r+

PR #452, opened 2019-11-28
Support dynamically sized types

jeehoonkang:
It supersedes #209 for supporting dynamically sized types (DST) in Crossbeam.  It's much cleaner and less intrusive than the previous attempts in that it doesn't introduce another `Atomic` type.  `Atomic<T>` is still there without significant changes.

The key idea is (1) instead of requiring the condition on `T: Sized`, (2) requiring `T: Pointable` that means an object of type `T` can be pointed to by a single word.  For instance, `Atomic` becomes: `pub struct Atomic<T: ?Sized + Pointable> { ... }`.

It is breaking the backward compatibility in that it ~~(1) increases the minimum required Rust version to 1.36 (for `MaybeUninit`) and (2)~~ now `const fn Atomic::null()` is Nightly only.

### DIFF
--- a/crossbeam-epoch/src/atomic.rs
+++ b/crossbeam-epoch/src/atomic.rs
@@ -8,6 +8,7 @@ use core::slice;
 use core::sync::atomic::{AtomicUsize, Ordering};
 
 use crate::alloc::alloc;
+use crate::alloc::boxed::Box;
 use crate::guard::Guard;
 use crossbeam_utils::atomic::AtomicConsume;
 
--- a/crossbeam-epoch/src/atomic.rs
+++ b/crossbeam-epoch/src/atomic.rs
@@ -130,7 +130,7 @@ fn decompose_tag<T: ?Sized + Pointable>(data: usize) -> (usize, usize) {
 ///
 /// ```
 /// use std::mem::MaybeUninit;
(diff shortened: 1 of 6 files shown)


jeehoonkang:
I think it's ready to be merged.  If everyone's happy with it, I'll merge it 3 days later.

crossbeam-epoch/src/atomic.rs
```diff
@@ -87,52 +88,211 @@ impl CompareAndSetOrdering for (Ordering, Ordering) {
...
+    fn init(init: Self::Init) -> usize;
+
+    /// Dereferences the given pointer.
+    unsafe fn deref<'a>(ptr: usize) -> &'a Self;
+
+    /// Mutably dereferences the given pointer.
+    unsafe fn deref_mut<'a>(ptr: usize) -> &'a mut Self;
+
+    /// Drops the object pointed to by the given pointer.
+    unsafe fn drop(ptr: usize);
```
taiki-e, line 158: It would be nice these unsafe public APIs to have safety documents.

jeehoonkang: Thanks!  I just wrote down the safety conditions.

PR #454, opened 2019-12-18
Make AtomicCell lockfree for u8, u16, u32, u64 sized values.

### LINKED ISSUE #423: AtomicCell without lock for non-usize values in stable.
Now that `std::atomic::AtomicU{8,16,32,64}` has been stabilized, please allow `AtomicCell` to store values with non-`usize` widths without a lock. 

This would require increasing the min rust version. Let me know if you like me to send a pull request.

Thanks.

manuthambi:
Fixes #423

### DIFF
--- a/crossbeam-utils/build.rs
+++ b/crossbeam-utils/build.rs
@@ -3,6 +3,12 @@ extern crate autocfg;
 fn main() {
     let cfg = autocfg::new();
     if cfg.probe_rustc_version(1, 31) {
-        println!("cargo:rustc-cfg=has_min_const_fn");
+        autocfg::emit("has_min_const_fn");
     }
+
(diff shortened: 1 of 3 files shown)


jeehoonkang:
I think we need to discuss the minimum supported Rust version (MSRV).  This PR is bumping it from 1.28 to 1.34, which is quite okay IMO.  Any thoughts?

(Actually, I'd like to bump it to 1.36, which is required for #452.)

manuthambi:
I reworked the patch to use autocfg crate. The code is cleaner and we don't need the MSRV bump anymore. 

But bumping MSRV is a great idea regardless.

jeehoonkang:
I think it's great!  Thank you.

bors r+

PR #455, opened 2019-12-24
Add no_std support to crossbeam-queue

stjepang:


phil-opp:
Shouldn't we update the `Cargo.toml` of the main `crossbeam` crate? Right now the crossbeam-queue crate is only included with the `std` feature.

PR #458, opened 2020-01-01
Fix unsoundness issues (mem::zeroed / ManuallyDrop -> MaybeUninit)

cynecx:
Several libs in crossbeam contain unsound code (eg. `Block::new`). This PR fixes them and requires a MSRV upgrade to 1.36.

### DIFF
--- a/crossbeam-channel/src/flavors/array.rs
+++ b/crossbeam-channel/src/flavors/array.rs
@@ -15,7 +15,7 @@
 
 use std::cell::UnsafeCell;
 use std::marker::PhantomData;
-use std::mem;
+use std::mem::{self, MaybeUninit};
 use std::ptr;
 use std::sync::atomic::{self, AtomicUsize, Ordering};
 use std::time::Instant;
@@ -33,7 +33,7 @@ struct Slot<T> {
     stamp: AtomicUsize,
 
     /// The message in this slot.
-    msg: UnsafeCell<T>,
+    msg: UnsafeCell<MaybeUninit<T>>,
 }
 
 /// The token type for the array flavor.
@@ -233,7 +233,7 @@ impl<T> Channel<T> {
         let slot: &Slot<T> = &*(token.array.slot as *const Slot<T>);
 
         // Write the message into the slot and update the stamp.
-        slot.msg.get().write(msg);
+        slot.msg.get().write(MaybeUninit::new(msg));
         slot.stamp.store(token.array.stamp, Ordering::Release);
 
         // Wake a sleeping receiver.
@@ -323,7 +323,7 @@ impl<T> Channel<T> {
         let slot: &Slot<T> = &*(token.array.slot as *const Slot<T>);
 
         // Read the message from the slot and update the stamp.
-        let msg = slot.msg.get().read();
+        let msg = slot.msg.get().read().assume_init();
         slot.stamp.store(token.array.stamp, Ordering::Release);
 
         // Wake a sleeping sender.
--- a/crossbeam-channel/src/flavors/array.rs
+++ b/crossbeam-channel/src/flavors/array.rs
@@ -542,7 +542,17 @@ impl<T> Drop for Channel<T> {
             };
 
             unsafe {
-                self.buffer.add(index).drop_in_place();
+                let ptr = self.buffer.add(index);
+                {
+                    // This requires an extra scope because when we drop the Slot,
+                    // reference to it should not exist.
+                    let slot = &mut *ptr;
+                    let msg = &mut *slot.msg.get();
+                    // Drop the message (MaybeUninit<T>).
+                    msg.as_mut_ptr().drop_in_place();
+                }
+                // Drop slot (This should be a no-op).
+                ptr.drop_in_place();
             }
         }
 
--- a/crossbeam-channel/src/flavors/array.rs
+++ b/crossbeam-channel/src/flavors/array.rs
@@ -542,17 +542,12 @@ impl<T> Drop for Channel<T> {
             };
 
             unsafe {
-                let ptr = self.buffer.add(index);
-                {
-                    // This requires an extra scope because when we drop the Slot,
-                    // reference to it should not exist.
-                    let slot = &mut *ptr;
+                let p = {
+                    let slot = &mut *self.buffer.add(index);
                     let msg = &mut *slot.msg.get();
-                    // Drop the message (MaybeUninit<T>).
-                    msg.as_mut_ptr().drop_in_place();
-                }
-                // Drop slot (This should be a no-op).
-                ptr.drop_in_place();
+                    msg.as_mut_ptr()
+                };
+                p.drop_in_place();
             }
         }
 
--- a/crossbeam-channel/src/flavors/array.rs
+++ b/crossbeam-channel/src/flavors/array.rs
@@ -15,12 +15,14 @@
 
 use std::cell::UnsafeCell;
 use std::marker::PhantomData;
-use std::mem::{self, MaybeUninit};
+use std::mem;
 use std::ptr;
 use std::sync::atomic::{self, AtomicUsize, Ordering};
 use std::time::Instant;
 
 use crossbeam_utils::{Backoff, CachePadded};
+
+use maybe_uninit::MaybeUninit;
 
 use context::Context;
 use err::{RecvTimeoutError, SendTimeoutError, TryRecvError, TrySendError};
--- a/crossbeam-epoch/src/sync/queue.rs
+++ b/crossbeam-epoch/src/sync/queue.rs
@@ -8,8 +8,7 @@
 //! Simon Doherty, Lindsay Groves, Victor Luchangco, and Mark Moir. 2004b. Formal Verification of a
 //! Practical Lock-Free Queue Algorithm. https://doi.org/10.1007/978-3-540-30232-2_7
 
-use core::mem::{self, ManuallyDrop};
-use core::ptr;
+use core::mem::MaybeUninit;
 use core::sync::atomic::Ordering::{Acquire, Relaxed, Release};
 
 use crossbeam_utils::CachePadded;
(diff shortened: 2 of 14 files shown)


stjepang:
Thank you for the PR! Is it still in progress? Looks good to merge to me :)

We also have some instances of uninitialized memory in bounded queues (files `array.rs` and `array_queue.rs`) where it would be good to use `MaybeUninit`.

cynecx:
@stjepang Well, the issue is the MSRV bump to 1.36 (MaybeUninit), that's why I've opened this as a draft. So you are saying that we should go ahead with the MSRV bump?

> We also have some instances of uninitialized memory in bounded queues (files array.rs and array_queue.rs) where it would be good to use MaybeUninit.

Yes, I will update them shortly and mark this PR ready for review then :).

cynecx:
@stjepang There is a failure in crossbeam's array flavoured channel: [travis](https://travis-ci.org/crossbeam-rs/crossbeam/jobs/633442774?utm_medium=notification&utm_source=github_status).

```rust
---- drops stdout ----
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `9336`,
 right: `9338`', crossbeam-channel/tests/array.rs:533:9
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

crossbeam-channel/src/flavors/array.rs
```diff
@@ -542,7 +542,17 @@ impl<T> Drop for Channel<T> {
...
+                {
+                    // This requires an extra scope because when we drop the Slot,
+                    // reference to it should not exist.
+                    let slot = &mut *ptr;
+                    let msg = &mut *slot.msg.get();
+                    // Drop the message (MaybeUninit<T>).
+                    msg.as_mut_ptr().drop_in_place();
+                }
+                // Drop slot (This should be a no-op).
+                ptr.drop_in_place();
```
cynecx: Should we leave this in or remove it, since it's clearly a no-op?

crossbeam-queue/src/array_queue.rs
```diff
@@ -415,7 +413,17 @@ impl<T> Drop for ArrayQueue<T> {
...
+                {
+                    // This requires an extra scope because when we drop the Slot,
+                    // reference to it should not exist.
+                    let slot = &mut *ptr;
+                    let value = &mut *slot.value.get();
+                    // Drop the message (MaybeUninit<T>).
+                    value.as_mut_ptr().drop_in_place();
+                }
+                // Drop slot (This should be a no-op).
+                ptr.drop_in_place();
```
cynecx: Same as above, should be remove this?

cynecx:
@stjepang It's ready for review now.

crossbeam-channel/src/flavors/array.rs
stjepang: I'm fine either way, but we could clean this up quite a bit by removing `ptr.drop_in_place()` because then we can also remove the extra scope, right?

cynecx: Sounds good, will push the remaining changes very soon.

cynecx:
This PR is ready to merge.

README.md
```diff
@@ -98,7 +98,7 @@ extern crate crossbeam;
 
 ## Compatibility
 
-The minimum supported Rust version is 1.28. Any change to this is considered a breaking change.
+The minimum supported Rust version is 1.36. Any change to this is considered a breaking change.
```
cuviper: > Any change to this is considered a breaking change.

... then it should be accompanied by a semver bump of all affected crates.

cuviper: (or at least update the text if the policy has changed)

jeehoonkang: Yeah, we decided bumping MSRV is a breaking change: https://github.com/crossbeam-rs/crossbeam/commit/d953f49ecc7eb5fe6c271b018a4a753ba435ee58  We need to write a changelog...

cynecx: Done.

cynecx:
Friendly neighborhood ping.

taiki-e:
fyi: by using https://crates.io/crates/maybe-uninit crate, we can avoid the MSRV bump ([like smallvec 0.6 did](https://github.com/servo/rust-smallvec/pull/180))

spacejam:
@taiki-e that seems really appealing to me. `maybe-uninit` seems like a pretty lightweight crate, 0 dependencies and over 1 million downloads since its fairly recent release in November 2019. Some pretty large projects rely on this crate and it's nice to avoid the transitive MSRV bump decision if they would otherwise be willing to update. This then allows a larger number of Rust users of unsafe code to make more progress when running miri.

stjepang:
I think we should just bump the minimum Rust version and consider publishing crossbeam 1.0 soon. A lot of unsafe code in crossbeam is really dealing with uninitialized memory so it's important we start using `MaybeUninit` ASAP.

spacejam:
one final consideration, @stjepang mentioned to me yesterday that [rayon lags by a year](https://github.com/rayon-rs/rfcs/blob/master/accepted/rfc0003-minimum-rustc.md). bumping MSRV to 1.36 means they need to wait until July 4 2020 before they can reduce their reliance on the uninitialized usage here. By using the `maybe-uninit` crate as a stopgap measure until july 4, it's possible for crossbeam to do a non-breaking patch update that silently reduces risk for more users. but, I'm not the one maintaining this, so I'll duck out at this point. although, if using that crate is a desired solution, I'm happy to implement it real quick since I'm an unemployed hobo.

cynecx:
I am fine either way, however @spacejam’s reasoning is really convincing.
ping @stjepang.

jonhoo:
I would love to see this land so that miri will work on downstream crates like https://github.com/jonhoo/flurry too!

jonhoo:
@jeehoonkang @stjepang what are next steps for this PR? I think we need your decision for who to best move forward. It'd probably be good to land this PR in one form or another sooner rather than later.

jeehoonkang:
I prefer `maybe-uninit` to bumping MSRV.  It looks like a fine temporary measure we can take.

jonhoo:
Thanks! @cynecx, do you have the time to update the PR to use `maybe-uninit`?

cynecx:
@jonhoo Sure :) The changes should be up in a few hours!

cynecx:
A question though, instead of adding `maybe-uninit` to each crate, should I just put it in `crossbeam-utils` and export a type-alias (`MaybeUninit`)? This would have a benefit, that future updates would simply have to change the type-alias in order to change the underlying `MaybeUninit` implementation. Also most of the affected crates are already having `crossbeam-utils` as dependency so that we just have to add the `maybe-uninit` dependency in `crossbeam-utils`. Thoughts?

jonhoo:
I would keep it separate, since (I think) it's weird for `crossbeam-utils` crate to publicly expose that type alias. If it could expose it _just_ to other `crossbeam` crates, then sure, but I don't think that's really an option.

cuviper:
If you reexport it, it becomes part of the public API, including the fact that it's identical to `maybe-uninit`'s type. You'd have to wrap it in a newtype to avoid that equality, to let you freely change the impl.

jonhoo:
@jeehoonkang I think this is now ready for review :tada:

crossbeam-epoch/src/sync/queue.rs
```diff
@@ -137,17 +134,20 @@ impl<T> Queue<T> {
...
-                self.head
-                    .compare_and_set(head, next, Release, guard)
-                    .map(|_| {
-                        guard.defer_destroy(head);
-                        Some(ManuallyDrop::into_inner(ptr::read(&n.data)))
-                    })
-                    .map_err(|_| ())
-            },
-            None | Some(_) => Ok(None),
+        unsafe {
```
jeehoonkang: It seems the scope of `unsafe` is enlarged---why?

crossbeam-epoch/src/sync/queue.rs
```diff
@@ -137,17 +134,20 @@ impl<T> Queue<T> {
...
+            match next.as_ref() {
+                Some(n) if condition(&*n.data.as_ptr()) => {
+                    self.head
+                        .compare_and_set(head, next, Release, guard)
+                        .map(|_| {
+                            guard.defer_destroy(head);
+                            // TODO: Replace with MaybeUninit::read when api is stable
```
jeehoonkang: "when the api is stabilized"?

crossbeam-queue/src/array_queue.rs
```diff
@@ -186,9 +188,7 @@ impl<T> ArrayQueue<T> {
                 ) {
                     Ok(_) => {
                         // Write the value into the slot and update the stamp.
-                        unsafe {
-                            slot.value.get().write(value);
-                        }
+                        unsafe { slot.value.get().write(MaybeUninit::new(value)) }
```
jeehoonkang: I think it's clearer to add `;`, because it's a statement.

crossbeam-epoch/src/sync/queue.rs
cynecx: Because the if guard of the match requires an unsafe dereference, so that would require an unsafe scope, however it would rather seem "weird" applying unsafe for just some fragments for this match block, although that would be more precise. I decided against precision and opted for a more cosmetic approach, since really most of the things in this match block is unsafe: `as_ref`,  dereferencing raw pointer, the matching `match` block.

crossbeam-epoch/src/sync/queue.rs
cynecx: Oh, that can be removed actually, it's a stale comment. Will push a cleanup in a few minutes.

crossbeam-queue/src/array_queue.rs
cynecx: Will do.

cynecx:
I hope if it's okay if I force-push now with the merge-conflicts? Because that will mess-up the review for sure xD.

crossbeam-epoch/src/sync/queue.rs
jeehoonkang: I prefer the original version, because it is accurately describing the fact that (1) converting a raw pointer to reference is potentially unsafe, and (2) we manually inspected (at least, in theory...) the safety of the conversion.  The conversion is safe because we have a sufficient permission to (de)reference the node of our interest after acquire load.

I believe this discussion can be made into another PR.  How about merging this PR without this change for now?

cynecx: I personally don't really mind, so pushed a new commit which reverts this :)

jeehoonkang:
Thanks!

bors r+

jonhoo:
@jeehoonkang Nice, thanks! What's the plan for next release between this and #466 ? Would be good to land something soon to get those out.

PR #465, opened 2020-02-02
Implement OS extension traits for ScopedJoinHandle

### LINKED ISSUE #197: Implement OS-specific methods on ScopedJoinHandle
See impls of `JoinHandleExt`, `AsRawHandle`, and `IntoRawHandle` here:
https://doc.rust-lang.org/nightly/std/thread/struct.JoinHandle.html

exrook:
Resolves #197

I implemented `into_pthread_t()`/`into_raw_handle()` methods by calling the `as` variants, as I'm not sure it would be sound to consume the join handle, since there would then be no guarantee the thread is joined before the end of the scope.

### DIFF
--- a/crossbeam-utils/src/thread.rs
+++ b/crossbeam-utils/src/thread.rs
@@ -522,6 +522,42 @@ impl<'scope, T> ScopedJoinHandle<'scope, T> {
     }
 }
 
+cfg_if! {
+    if #[cfg(unix)] {
+        use std::os::unix::thread::{JoinHandleExt, RawPthread};
+
+        impl<'scope, T> JoinHandleExt for ScopedJoinHandle<'scope, T> {
+            fn as_pthread_t(&self) -> RawPthread {
+                // Borrow the handle. The handle will surely be available because the root scope waits
+                // for nested scopes before joining remaining threads.
+                let handle = self.handle.lock().unwrap();
+                handle.as_ref().unwrap().as_pthread_t()
+            }
+            fn into_pthread_t(self) -> RawPthread {
+                self.as_pthread_t()
+            }
+        }
+    } else if #[cfg(windows)] {
+        use std::os::windows::io::{AsRawHandle, IntoRawHandle, RawHandle};
+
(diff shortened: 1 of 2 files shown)


crossbeam-utils/src/thread.rs
```diff
@@ -522,6 +522,35 @@ impl<'scope, T> ScopedJoinHandle<'scope, T> {
     }
 }
 
+#[cfg(unix)]
+use std::os::unix::thread::{JoinHandleExt, RawPthread};
+#[cfg(unix)]
+impl<'scope, T> JoinHandleExt for ScopedJoinHandle<'scope, T> {
+    fn as_pthread_t(&self) -> RawPthread {
+        self.handle.lock().unwrap().as_ref().unwrap().as_pthread_t()
```
jeehoonkang: How about splitting this into two lines (as done for Windows)?

crossbeam-utils/src/thread.rs
```diff
@@ -522,6 +522,35 @@ impl<'scope, T> ScopedJoinHandle<'scope, T> {
     }
 }
 
+#[cfg(unix)]
```
jeehoonkang: Is it possible to use cfg-if?  IMO, it would look better.

crossbeam-utils/src/thread.rs
jeehoonkang: I'm concerned with the safety here.  Is it okay to return a resource protected by a lock?  If so, why?

exrook: I believe this is safe since the resource being returned (`RawPthread`) is an opaque handle to the underlying OS thread that is copied and returned to the caller. I don't see how this operation could affect the soundness of any other code.

crossbeam-utils/src/thread.rs
exrook: done

crossbeam-utils/src/thread.rs
exrook: I've split the lines

PR #466, opened 2020-02-03
fix use-after-free in crossbeam-epoch/sync/queue

### LINKED ISSUE #238: Use after free in Michael-Scott queue?

tomtomjhj:
`pop()` must completely unlink the popped node from the shared memory before it calls `defer_destroy()` to prevent use-after-free. This implementation is based on the variation by Doherty et al. where the `head == tail` check is done after a successful CAS, which can be slightly more efficient than the original MSQueue.

closes #238

### DIFF
--- a/crossbeam-epoch/src/sync/queue.rs
+++ b/crossbeam-epoch/src/sync/queue.rs
@@ -4,6 +4,9 @@
 //!
 //! Michael and Scott.  Simple, Fast, and Practical Non-Blocking and Blocking Concurrent Queue
 //! Algorithms.  PODC 1996.  http://dl.acm.org/citation.cfm?id=248106
+//!
+//! Simon Doherty, Lindsay Groves, Victor Luchangco, and Mark Moir. 2004b. Formal Verification of a
+//! Practical Lock-Free Queue Algorithm. https://doi.org/10.1007/978-3-540-30232-2_7
 
 use core::mem::{self, ManuallyDrop};
 use core::ptr;
@@ -117,6 +120,11 @@ impl<T> Queue<T> {
                 self.head
                     .compare_and_set(head, next, Release, guard)
                     .map(|_| {
+                        let tail = self.tail.load(Relaxed, guard);
+                        // Advance the tail so that we don't retire a pointer to a reachable node.
+                        if head == tail {
+                            let _ = self.tail.compare_and_set(tail, next, Release, guard);
+                        }
                         guard.defer_destroy(head);
                         Some(ManuallyDrop::into_inner(ptr::read(&n.data)))
                     })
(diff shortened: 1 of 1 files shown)


jeehoonkang:
This patch looks good, but I'm not completely sure whether it closes #238.  Is it the case that all the use-after-free bugs in queue resolved?

tomtomjhj:
Other queues use their own reclamation methods.

jeehoonkang:
I'm merging it, as I think it's better to merge bugfix fast...

bors r+

PR #469, opened 2020-02-09
Added get_or_insert_with function fn to SkipList for lazy values.

snow01:
Added get_or_insert_with function fn to SkipList for lazy values.

### DIFF
--- a/crossbeam-skiplist/src/base.rs
+++ b/crossbeam-skiplist/src/base.rs
@@ -472,6 +472,11 @@ where
 
     /// Finds an entry with the specified key, or inserts a new `key`-`value` pair if none exist.
     pub fn get_or_insert(&self, key: K, value: V, guard: &Guard) -> RefEntry<K, V> {
+        self.insert_internal(key, || value, false, guard)
+    }
+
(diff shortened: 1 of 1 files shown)


crossbeam-skiplist/src/base.rs
```diff
@@ -881,7 +886,7 @@ where
 
                 // Write the key and the value into the node.
                 ptr::write(&mut (*n).key, key);
-                ptr::write(&mut (*n).value, value);
+                ptr::write(&mut (*n).value, value());
```
taiki-e, line 889: I would prefer to create the value before creating a node, as closure may panics.

https://github.com/crossbeam-rs/crossbeam/blob/220e690d1d5d0643cfc006aa10455bd7b28297ef/crossbeam-skiplist/src/base.rs#L878-L879

taiki-e (changes requested):
Thanks for the PR and sorry for the late review. This looks good to me aside from one nit.

crossbeam-skiplist/src/base.rs
snow01: I have created new PR request on latest codebase. Please review that at the earliest.

Issue #347, opened 2019-03-22
epoch: possible reads of uninitialized data


ghost:
Reported in https://github.com/servo/servo/issues/23077

```
==130127== Thread 43 WRRenderBackend#:
==130127== Conditional jump or move depends on uninitialised value(s)
==130127==    at 0x95D09AA: crossbeam_epoch::internal::Global::collect (internal.rs:189)
==130127==    by 0x957E156: crossbeam_epoch::internal::Local::pin (internal.rs:379)
==130127==    by 0x9551625: crossbeam_epoch::collector::LocalHandle::pin (collector.rs:72)
==130127==    by 0x95550E7: crossbeam_epoch::default::pin::{{closure}} (default.rs:23)
==130127==    by 0x9555097: crossbeam_epoch::default::with_handle::{{closure}} (default.rs:42)
==130127==    by 0x955668D: <std::thread::local::LocalKey<T>>::try_with (local.rs:300)
==130127==    by 0x9554F72: crossbeam_epoch::default::with_handle (default.rs:42)
==130127==    by 0x63530B9: crossbeam_epoch::default::pin (default.rs:23)
==130127==    by 0x634717E: <crossbeam_channel::flavors::list::Channel<T>>::start_send (list.rs:161)
==130127==    by 0x6347DB8: <crossbeam_channel::flavors::list::Channel<T>>::send (list.rs:381)
==130127==    by 0x634E381: <crossbeam_channel::channel::Sender<T>>::send (channel.rs:416)
==130127==    by 0x634F347: compositing::compositor_thread::CompositorProxy::send (compositor_thread.rs:31)
==130127== 
==130127== Conditional jump or move depends on uninitialised value(s)
==130127==    at 0x95D4A2F: core::ptr::real_drop_in_place (ptr.rs:193)
==130127==    by 0x95D09D0: crossbeam_epoch::internal::Global::collect (internal.rs:192)
==130127==    by 0x957E156: crossbeam_epoch::internal::Local::pin (internal.rs:379)
==130127==    by 0x9551625: crossbeam_epoch::collector::LocalHandle::pin (collector.rs:72)
==130127==    by 0x95550E7: crossbeam_epoch::default::pin::{{closure}} (default.rs:23)
==130127==    by 0x9555097: crossbeam_epoch::default::with_handle::{{closure}} (default.rs:42)
==130127==    by 0x955668D: <std::thread::local::LocalKey<T>>::try_with (local.rs:300)
[...]

PR #474, opened 2020-02-17
Patch release for all crossbeam crates.

### LINKED ISSUE #409: Use of deprecated mem::uninitialized
https://github.com/crossbeam-rs/crossbeam/blob/staging/crossbeam-epoch/src/deferred.rs#L39
https://github.com/crossbeam-rs/crossbeam/blob/staging/crossbeam-epoch/src/deferred.rs#L54
https://github.com/crossbeam-rs/crossbeam/blob/staging/crossbeam-epoch/src/sync/queue.rs#L50

### LINKED ISSUE #468: [bug] [undefined behaviour] Use of `mem::uninitialized()`
Seems to occur in `crossbeam-epoch`.

Full MIRI output is here: https://github.com/xacrimon/dashmap/runs/433047062.

### LINKED ISSUE #472: Multiple versions for dependency `autocfg`
#### Summary
Latest release of crossbeam-utils (0.7.0) depends on an old version of autocfg, which triggers the following Clippy warning when used together with other packages and slows down compilation:
```
warning: multiple versions for dependency `autocfg`: 1.0.0, 0.1.7
  |
  = note: `-W clippy::multiple-crate-versions` implied by `-W clippy::cargo`
  = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#multiple_crate_versions
```
The version from git already depends on `autocfg 1`, so it seems it could be fixed by just releasing a new version of crossbeam-utils - the last one is 3 months old.

#### Steps to reproduce
Add these dependencies to your project:
dirs 2.0.2, which depends on dirs-sys -> redox_users -> rust-argon2 -> **crossbeam-utils -> autocfg 0.1.7**
rust-ini 0.15.1, which depends on ordered-multimap -> **hashbrown -> autocfg 1**

Run: `cargo clippy`

### LINKED ISSUE #473: Publish new releases of crossbeam crates
We currently have a couple of open issues that are resolved on master (like #472 and #468), and some important fixes that have landed (like #466 and #458). It'd be great to push out a new release of the various crossbeam crates to get those to users. What currently stands in the way of doing so? Can the next batch of releases all be point releases, or is a breaking release necessary?

What are the next steps, and is there any way in which I can help?
/cc @stjepang @jeehoonkang

jonhoo:
This bumps the patch version for all sub-crates and the top-level crate and updates the changelogs with references to PRs where applicable. As far as I can tell, there are no breaking changes. The top-level crate has not change at all, though I bumped `rand` to `0.7` in all the crates, which I guess technically requires a patch bump if we do another release.

Fixes #473.
Closes #472.
Closes #468.
Closes #409.
Not sure if this also solves #347 by virtue of #458?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,3 @@
-# Version 0.7.4
-
-- Bump `rand` dev-dependency to version 0.7.
-
 # Version 0.7.3
 
 - Fix breakage with nightly feature due to rust-lang/rust#65214.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Version 0.7.4
+
+- Bump `rand` dev-dependency to version 0.7.
+
 # Version 0.7.3
 
 - Fix breakage with nightly feature due to rust-lang/rust#65214.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ name = "crossbeam"
 # - Update CHANGELOG.md
 # - Update README.md
 # - Create "crossbeam-X.Y.Z" git tag
-version = "0.7.3"
+version = "0.7.4"
 authors = ["The Crossbeam Project Developers"]
 license = "MIT/Apache-2.0"
 readme = "README.md"
@@ -60,7 +60,7 @@ path = "./crossbeam-utils"
 default-features = false
 
 [dev-dependencies]
-rand = "0.6"
+rand = "0.7"
 
 [workspace]
 members = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ name = "crossbeam"
 # - Update CHANGELOG.md
 # - Update README.md
 # - Create "crossbeam-X.Y.Z" git tag
-version = "0.7.4"
+version = "0.7.3"
 authors = ["The Crossbeam Project Developers"]
 license = "MIT/Apache-2.0"
 readme = "README.md"
@@ -60,7 +60,7 @@ path = "./crossbeam-utils"
 default-features = false
 
 [dev-dependencies]
-rand = "0.7"
+rand = "0.6"
 
 [workspace]
 members = [
--- a/crossbeam-channel/CHANGELOG.md
+++ b/crossbeam-channel/CHANGELOG.md
@@ -1,6 +1,5 @@
 # Version 0.4.1
 
-- Bump `rand` dev-dependency to version 0.7.
 - Avoid time drift in `channel::tick`. (#456)
 - Fix unsoundness issues by adopting `MaybeUninit`. (#458)
 
--- a/crossbeam-channel/CHANGELOG.md
+++ b/crossbeam-channel/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Version 0.4.1
+
+- Bump `rand` dev-dependency to version 0.7.
+- Avoid time drift in `channel::tick`. (#456)
+- Fix unsoundness issues by adopting `MaybeUninit`. (#458)
+
 # Version 0.4.0
 
 - Bump the minimum required version to 1.28.
--- a/crossbeam-channel/Cargo.toml
+++ b/crossbeam-channel/Cargo.toml
@@ -4,7 +4,7 @@ name = "crossbeam-channel"
 # - Update CHANGELOG.md
 # - Update README.md
 # - Create "crossbeam-channel-X.Y.Z" git tag
-version = "0.4.0"
+version = "0.4.1"
 authors = ["The Crossbeam Project Developers"]
 license = "MIT/Apache-2.0 AND BSD-2-Clause"
 readme = "README.md"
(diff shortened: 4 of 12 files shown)


jeehoonkang:
It looks good to me.  Thanks!

bors r+

I'll publish the crates soon.

jeehoonkang:
CI fails with e.g. https://travis-ci.org/crossbeam-rs/crossbeam/jobs/651554494#L233  @jonhoo any ideas how to fix?

jonhoo:
Bah, looks like `rand 0.7` requires support for the 2018 edition, which I believe is 1.32.
That said, it is a `dev-dependency`, and generally MSRV applies only to the library itself, not the library's tests. Would you consider changing the CI to only run `cargo check --lib --examples` on MSRV?

jeehoonkang:
Do we have an option to leave `rand` as is?  If `rand 0.6` doesn't have a critical bug, I would rather wait for a few more months before bumping it.  What do you think?

jonhoo:
Reverted the `rand` bump

jeehoonkang:
uh-oh, seems like a real bug... https://travis-ci.org/crossbeam-rs/crossbeam/jobs/651567571#L538

jonhoo:
Ouch... As far as I can tell from [the history](https://github.com/crossbeam-rs/crossbeam/commits/master/crossbeam-channel), this basically has to be https://github.com/crossbeam-rs/crossbeam/pull/458 if the bug wasn't already present in the previous release. There is, I suppose, also a slim chance that it is related to https://github.com/crossbeam-rs/crossbeam/pull/454 in `crossbeam-utils`? /cc @manuthambi @cynecx.

jonhoo:
It _could_ also be that the test itself is just racy?

jonhoo:
It certainly looks like that might be the case, because it passed just fine here: https://travis-ci.org/crossbeam-rs/crossbeam/builds/651566762

jeehoonkang:
Hmm... I feel we should ping @stjepang for this matter.  I'm not familiar with `crossbeam-channel`.  Stjepan, do you think we can publish new crates even in the presence of the bug that popped up in the CI?

cuviper:
> uh-oh, seems like a real bug... https://travis-ci.org/crossbeam-rs/crossbeam/jobs/651567571#L538

I was just looking at Fedora failed builds in that same test, on the current crossbeam-channel 0.4.0. In other words, I think this isn't new, at least.

```
$ for (( i=1 ;; i+=1 )); do echo "*** test $i ***"; cargo test --test ready fairness || break; done
```

This loops until failure -- usually I make it 20-30 times successfully. It seems to fails even more frequently with `--release`, so maybe the test just needs more allowance?

jonhoo:
Yeah, that'd be my guess. My inclination here would be to merge this and then do a release. The test is probably racy, and isn't caused by this release, so it's not worth holding up the release we wait for Stjepan to have a chance to look at the test to see if the tolerance should be increased.

jonhoo:
@jeehoonkang What do you think?

jeehoonkang:
Okay, I'm trying again to merge

bors retry

jeehoonkang:
@jonhoo I'll release crates now.  Would you please create an issue on the CI failure?

jeehoonkang:
- `crossbeam-deque` is not published due to the following error:

  ```
  error[E0599]: no method named `into_handle` found for type `epoch::Owned<Buffer<T>>` in the current scope
     --> src/lib.rs:230:33
      |
  230 |             buffer.into_owned().into_handle().dealloc();
      |                                 ^^^^^^^^^^^ method not found in `epoch::Owned<Buffer<T>>`
  ```

- `crossbeam-queue` is not published because I don't have a permission to do so.  @stjepang would you please give me permission?

jonhoo:
@jeehoonkang 
Racu test filed as https://github.com/crossbeam-rs/crossbeam/issues/477.
As for the compile error, `cargo publish --dry-run` runs just fine for me on `crossbeam-deque`:
```
$ cargo publish --dry-run
    Updating crates.io index
   Packaging crossbeam-deque v0.7.3 (/home/jon/dev/others/crossbeam/crossbeam-deque)
   Verifying crossbeam-deque v0.7.3 (/home/jon/dev/others/crossbeam/crossbeam-deque)
   Compiling crossbeam-utils v0.7.2
   Compiling crossbeam-epoch v0.8.2
   Compiling crossbeam-deque v0.7.3 (/home/jon/.cargo-target/package/crossbeam-deque-0.7.3)
    Finished dev [unoptimized + debuginfo] target(s) in 1.35s
   Uploading crossbeam-deque v0.7.3 (/home/jon/dev/others/crossbeam/crossbeam-deque)
```
I think maybe what happened is that `crossbeam-deque` relies on the new version of `crossbeam-epoch` or `crossbeam-epoch`, and that new version wasn't published yet when you tried to run `cargo publish`. `cargo publish` will never use `path` dependencies, and will instead just use whatever the latest version on crates.io is. This makes me think that we need to bump the minimum version requirement for one of the internal dependencies of `crossbeam-deque` before uploading it to crates.io. Specifically, we need to set it to the version that introduced `into_handle` for `epoch::Owned<Buffer<T>>`.

jeehoonkang:
@jonhoo sorry, it was my bad.  It was resolved in #476.  I just published a new version of deque: https://crates.io/crates/crossbeam-deque

jeehoonkang:
`crossbeam-queue` is not published because I don't have a permission to do so. @stjepang would you please give me permission?

Issue #451, opened 2019-11-21
Possible atomicity violation in crossbeam-channel SyncWaker notify


BurtonQin:
https://github.com/crossbeam-rs/crossbeam/blob/424f926673d652f306f266d4309ed8717e931144/crossbeam-channel/src/waker.rs#L221-L231

is_empty is an AtomicBool. 
In function SyncWaker::notify()
1. load is_empty to check
2. lock and notify
3. store new value to is_empty

The struct implements Sync and notify() takes an immutable self, so it can be called in two threads.
To me, it seems that the function intends to notify only once when is_empty is not true.
However, the following sequence may happen:
Th1: 1->W->W->W->2 
Th2: W->1->2->3
Thread 1 load is_empty and check, then Thread 2 notify and change the is_empty. Thread 2 executes notify again. Now notify() is called twice.
The fix is to lift the lock before the is_empty check, or check is_empty again after the lock.

stjepang:
The reason why I didn't bother with the additional `is_empty` check after locking is because I thought spurious notifications are not much of a problem (they might sometimes hurt performance, but not correctness).

I'd accept a PR adding an additional `is_empty` check after the lock :)

taiki-e:
Fixed in #478

PR #478, opened 2020-02-27
channel: call notify only from one thread

snsmac:
Fix for #451 

This does not seem to have a noticeable influence on the benchmarks


snsmac:
I have no clue what is causing the ci failure

PR #479, opened 2020-03-16
relax memory ordering of treiber stack

laizy:
I think it is correct to use relaxed ordering, since:
1. pop function has no data to publish to other thread.
2. all write operations on `TreiberStack.head` is atomic read-modify-write, so will be part of the `release sequence` headed by the release operation in `push` function, according to the [cpp ref](https://en.cppreference.com/w/cpp/atomic/memory_order).


jeehoonkang:
It's true.  Thanks!

bors r+

PR #480, opened 2020-03-20
Make the `queue` module available on no_std + alloc

phil-opp:
The crossbeam_queue crate is no_std compatible since #455.

### DIFF
--- a/crossbeam-queue/src/lib.rs
+++ b/crossbeam-queue/src/lib.rs
@@ -11,6 +11,7 @@
 #![warn(missing_docs)]
 #![warn(missing_debug_implementations)]
 #![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(feature = "nightly", feature(cfg_target_has_atomic))]
 
 #[macro_use]
 extern crate cfg_if;
@@ -29,6 +30,7 @@ cfg_if! {
 
 extern crate crossbeam_utils;
 
+#[cfg_attr(feature = "nightly", cfg(target_has_atomic = "ptr"))]
 cfg_if! {
     if #[cfg(any(feature = "alloc", feature = "std"))] {
         mod array_queue;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,6 +77,12 @@ pub mod utils {
     pub use crossbeam_utils::CachePadded;
 }
 
+mod _queue {
+    pub extern crate crossbeam_queue;
+}
+#[doc(inline)]
+pub use _queue::crossbeam_queue as queue;
+
 cfg_if! {
     if #[cfg(feature = "std")] {
         mod _deque {
@@ -95,12 +101,6 @@ cfg_if! {
         // HACK(stjepang): This is the only way to reexport `select!` in Rust older than 1.30.0
         #[doc(hidden)]
         pub use _channel::*;
-
-        mod _queue {
-            pub extern crate crossbeam_queue;
-        }
-        #[doc(inline)]
-        pub use _queue::crossbeam_queue as queue;
 
         pub use crossbeam_utils::sync;
         pub use crossbeam_utils::thread;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,6 +77,7 @@ pub mod utils {
     pub use crossbeam_utils::CachePadded;
 }
 
+#[cfg_attr(feature = "nightly", cfg(target_has_atomic = "ptr"))]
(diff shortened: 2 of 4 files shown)


jeehoonkang:
Thanks!

bors r+

P.S. I'm really enjoying your blog on OS.  Thank you for great posts.

phil-opp:
> P.S. I'm really enjoying your blog on OS. Thank you for great posts.

Thanks a lot, great to hear that! This change is related to the upcoming post about async/await: I plan to use the `SeqQueue`/`ArrayQueue` types for my executor implementation. So thank you for your work on this crate!

phil-opp:
It seems like the travis build with `--no-default-features` failed. I will look into it.

phil-opp:
The problem was that the `crossbeam_queue` dependency is only enabled with either the `std` or `alloc` feature, so that the build without both features failed. There are two ways to fix this: Either make `crossbeam_queue` non-optional or only make the `queue` module available if at least the `alloc` feature is enabled. I tried both variants and decided for the latter because it leads to cleaner code and the `queue` module would be empty anyway without the `alloc`/`std` features.

Sorry about the churn!

phil-opp:
One more fix was needed because [travis still failed](https://travis-ci.org/github/crossbeam-rs/crossbeam/jobs/664821742?utm_medium=notification&utm_source=github_status): The dependency was missing `default-features = false` so that the `std` feature was still enabled for `no_std`.

phil-opp:
Oh wow, this change is more difficult than I thought :sweat_smile:. Travis fails again, this time because atomics are not available on `thumbv6m-none-eabi`. I will push a fix in a moment, but I recommend that we wait until travis succeeds before giving it to bors, in case there are other platform-specific requirements that I did not think of.

phil-opp:
It seems like there is no good solution for this at the moment. The problem is that the atomic compare-and-swap operations are not available on all devices. The `thumbv6m-none-eabi` architecture is an example for a device without this feature. (Note that this issue was already present in the `crossbeam-queue` sub-crate before my PR, it just wasn't caught by the CI.)

Unfortunately, I can't find a way to disable the queue module for such architectures. The [`#[cfg(target_has_atomic = ...)]`](https://github.com/rust-lang/rust/issues/32976) attribute would be the best approach, but it is not stable yet and thus only works on nightly. There also seems to be no way to disable the module explicitly for the `thumbv6m-none-eabi` as a workaround.

So it is not clear to me how we should proceed with this. What do you think?

src/lib.rs
```diff
@@ -77,6 +77,16 @@ pub mod utils {
     pub use crossbeam_utils::CachePadded;
 }
 
+cfg_if! {
```
taiki-e:
```suggestion
#[cfg_attr(feature = "nightly", cfg(target_has_atomic = "ptr"))]
cfg_if! {
```

phil-opp: Thanks, seems like a good solution!

phil-opp:
I needed to move the cfg to the crossbeam-queue crate because it is compiled as a dependency before the cfg in crossbeam itself applies.

phil-opp:
Ok, travis is finally green, so this is ready for review again!

crossbeam-queue/src/lib.rs
```diff
@@ -29,6 +30,7 @@ cfg_if! {
 
 extern crate crossbeam_utils;
 
+#[cfg_attr(feature = "nightly", cfg(target_has_atomic = "ptr"))]
```
jeehoonkang: In stable+std+thumbv6m-none-eabi, what happens?  I guess the compilation fails, and wonder if it's the desired behavior.

phil-opp: I don't think that the standard library is available for `thumbv6m-none-eabi`, so the compilation would fail either way. As far as I know, it is also not possible to create a `no_std` binary that uses the `alloc` library on stable yet, because the required `#[alloc_error_handler]` is not stable yet.

phil-opp:
Thanks for merging this! Any chance we can get a crates.io release for this soon or do you prefer to bundle more changes into the next version? I'm asking because I'm currently working on a blog post that uses the `ArrayQueue` type and am currently using the `crossbeam-queue` subcrate for this, but would like to use the main `crossbeam` crate if possible.

jeehoonkang:
@stjepang would you please give me an access to crates.io for the `crossbeam-queue` crate?  I happen to have access to the other crates, but not for this one.

PR #481, opened 2020-03-22
Added latest flume to benchmarks

zesterer:
As a point of interest, here are the results for my Linux Ryzen 7 4-core/8-thread machine.

image

I plan to enable more of the benchmarks soon since Flume now supports more features.


zesterer:
This isn't quite ready to merge yet, not all benchmarks that can be enabled have been enabled.

PR #485, opened 2020-04-06
Fix `features` typo in ArrayQueue's error types (crossbeam-queue)

mexus:
Currenly, `PushError` and `PopError` do not implement `std::error::Error` even when `std` feature is activated :(

### DIFF
--- a/crossbeam-queue/src/err.rs
+++ b/crossbeam-queue/src/err.rs
@@ -16,7 +16,7 @@ impl fmt::Display for PopError {
     }
 }
 
-#[cfg(features = "std")]
+#[cfg(feature = "std")]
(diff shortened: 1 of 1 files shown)


mexus:
rebased :ok_hand:

jeehoonkang:
It seems CI fails: https://github.com/crossbeam-rs/crossbeam/pull/485/checks?check_run_id=595070829

taiki-e:
Probably it needs to add leading-colon to the path. (`std::error::Error` -> `::std::error::Error`)

mexus:
Yep, seems like it did the trick!

Sorry I haven't tested it locally at the first place :(

PR #487, opened 2020-04-08
Add preliminary support for loom

### LINKED ISSUE #486: Add loom support to crossbeam-epoch
It'd be really cool if `crossbeam-epoch` had support for [`loom`](https://github.com/tokio-rs/loom). That would allow us to write more reliable tests where every execution interleaving and atomic visibility case actually gets exercised, without resorting to some of the "spin up a bunch of threads" tests we have today. It would also allow downstream crates to write similar tests on top of `crossbeam-epoch`, such as in `flurry` where we need loom to exercise some tricky [contention cases](https://github.com/jonhoo/flurry/pull/72#issuecomment-610618109) that are hard to hit reliably otherwise.

The loom README has a decent amount of info on how to go about adding loom support, but it _basically_ boils down to using types from `loom` instead of from `std` (or `core`) when `cfg(loom)`. Then, for tests, you just surround the test with `loom::model(|| {})`, and loom will automatically run the closure under different concurrent execution schedules.

/cc @jeehoonkang @carllerche @domenicquirl

jonhoo:
This patch only adds support to parts of `utils` and to `epoch`. Some
parts of `utils` had to be left out, since they rely on
`AtomicUsize::new` being `const` (which it is not in `loom`). Other
parts had to be left out due to the lack of `thread::Thread` in `loom`.
All the parts needed for `epoch` were successfully moved to loom.

The one loom test currently passes! I've added a loom pass on
crossbeam-epoch to CI.

Also, note that the uses of `UnsafeCell` in `utils` have not been moved
to `loom::cell::UnsafeCell`, as loom's `UnsafeCell` does not support `T:
?Sized`, which `AtomicCell` depends on.

Fixes #486.


crossbeam-epoch/src/lib.rs
```diff
@@ -66,6 +66,68 @@ extern crate core;
...
+    pub(crate) mod sync {
+        pub(crate) mod atomic {
+            pub(crate) use loom::sync::atomic::{AtomicUsize, Ordering};
+            pub(crate) fn fence(ord: Ordering) {
+                if let Ordering::Acquire = ord {
+                } else {
+                    // loom only supports acquire fences at the moment
+                    // let's not panic...
+                }
+                loom::sync::atomic::fence(Ordering::Acquire)
```
cynecx, line 86: So this would weaken `atomic::fence(Ordering::SeqCst)`s which crossbeam-epoch uses in some places. I was wondering what kind of effects that would have, however I am not sure how hard it is to trigger it with loom?

cynecx:
I think it would be nice if we could run some variant of the examples (`treiber-stack`) through loom too?

crossbeam-epoch/src/lib.rs
jonhoo: Yes, it would indeed, but I found no other way to get around it at the moment. I filed https://github.com/tokio-rs/loom/issues/117 for things like this that are currently short-comings of this PR.

jonhoo:
Should be quite possible! The easiest way would probably be to factor out the core of `treiber-stack` into some kind of function that we could easily call from a test. Either than, or we copy-paste the treiber stack into a test and wrap it in `loom::model`.

jonhoo:
I added one! And interestingly enough, it _fails_ with
> Causality violation: Concurrent load and mut accesses.

jonhoo:
@carllerche When I get the error above, how do I then find out _where_ the violation happened?

jonhoo:
I chatted to Carl, and it looks like the cause for this is probably a bug in loom: https://github.com/tokio-rs/loom/issues/124. Let's see when it is fixed whether it then passes!

jonhoo:
Oooh, with https://github.com/tokio-rs/loom/pull/125, loom actually runs for the treiber stack example! Letting it go through the iterations now to see if it finds any problems. None so far :D

jonhoo:
And with https://github.com/tokio-rs/loom/pull/128, it runs to completion without errors! :open_mouth:

jonhoo:
CI doesn't like the patch in 08b37abff58ff2b92525585e8a91c5b2e2eee0c5 :sweat_smile: It will go away once loom gets another release.

jonhoo:
Hmm, how can I see what the errors were @taiki-e? All of them just say "This check failed" as far as I can tell?

taiki-e:
@jonhoo  I restarted CI -- looks like it fixed the issue: https://github.com/crossbeam-rs/crossbeam/pull/487/checks

jonhoo:
The minimal version bump to 1.30 is a little awkward, but is needed to allow us to use paths for macro imports. While digging into it, I also noticed that _technically_ 1.28 isn't the msrv at all for these crates. For example, the `alloc` crate didn't stabilize [until 1.36](https://caniuse.rs/features/alloc), yet we use it if someone does a `no_std` build with `--feature alloc`..

jonhoo:
@jeehoonkang I believe this is now ready for review :)

crossbeam-epoch/src/lib.rs
```diff
@@ -66,6 +66,87 @@ extern crate core;
...
+    pub(crate) use loom::thread_local;
+}
+#[cfg(not(loom))]
+#[allow(unused_imports, dead_code)]
+pub(crate) mod concurrency {
+    #[cfg(any(feature = "alloc", feature = "std"))]
+    pub(crate) mod cell {
+        #[derive(Debug)]
+        #[repr(transparent)]
+        pub(crate) struct UnsafeCell<T>(::core::cell::UnsafeCell<T>);
```
taiki-e, line 113: Why this is a new type, not just reexport?

crossbeam-utils/Cargo.toml
```diff
@@ -25,6 +25,9 @@ alloc = []
 cfg-if = "0.1.2"
 lazy_static = { version = "1.1.0", optional = true }
 
+[target.'cfg(loom)'.dependencies]
```
taiki-e, line 28: I'm surprised that this works. (I felt it was a bug, but actually, [only some things like `cfg(feature)` don't seem to work.](https://github.com/rust-lang/cargo/pull/7660/files#diff-2ab1ff723e1d553d9d5382510b10cfe6R75-R97))

taiki-e:
> the `alloc` crate didn't stabilize [until 1.36](https://caniuse.rs/features/alloc), yet we use it if someone does a `no_std` build with `--feature alloc`..

I think this is fine given that `alloc` feature is not enabled by default and its name is the same as the name of the crate on which it depends (other features should also be fine). but I agree with it is preferable to mention features that have a different msrv from the default feature.

crossbeam-epoch/src/lib.rs
jonhoo: I'm not sure I follow — I need to add new methods to it (`with` and `with_mut`), which requires a newtype.

crossbeam-utils/Cargo.toml
jonhoo: Yeah, I was surprised too! I think the thinking here is that you can only use `cfg()` in this spot for `cfg` that are available from the very early stages of compilation (like `RUSTFLAGS`). Or something. Things that are only available later (like after figuring out what features the crate will be compiled by) can't be used.

crossbeam-epoch/src/lib.rs
taiki-e: Oops, sorry! I somehow missed that UnsafeCell does not have those methods.

jonhoo: Yeah, loom's `UnsafeCell` has a slightly different API basically because it wants to track how long you use the mutable reference for. And so we need to shim that in the non-loom case.

cynecx:
Ping.

jonhoo:
@cynecx I believe this is ready to merge once reviewed. I've pushed a rebased version now to make sure there aren't any conflicts.

jonhoo:
The two failures appear unrelated to this PR. I think CI is simply broken on `master`. Once that's fixed I'd be happy to rebase again.

jeehoonkang:
Crossbeam has evolved quite a lot recently, so I wonder if the unrelated failures are gone.

jonhoo:
@jeehoonkang rebased!

jonhoo:
Hey, it works again! :tada:

taiki-e (changes requested):
Sorry for the delay. Could you do the following?
- Resolve conflicts.
- Rename `cfg(loom)` to something like `cfg(loom_crossbeam)` to avoid conflict with other crates that using `cfg(loom)`. (I don't know what the `loom`'s recommendation for this is. If there is a recommendation, it is better to use it.)
- Bump `loom` to 0.4.
- Document that `cfg(loom_)` is not stable API (probably on the line above `cfg(loom).dependencies`, or on the `concurrency` module).
   As for the content of the docs, I think it's okay to with something similar to the [`nightly`](https://github.com/crossbeam-rs/crossbeam/blob/47110fd42d9b5d67548071e32714f39f59af642b/crossbeam-epoch/Cargo.toml#L30-L34) or [`sanitize`](https://github.com/crossbeam-rs/crossbeam/pull/591/files#diff-01093eefa5ed07f78fe117f0a03c931349d2ab37aae2a0483e470f25c1bfe5e5R27-R29) feature.

Ideally, I'd like @jeehoonkang to review the change to epoch, but this patch doesn't change the behavior of epoch when `cfg(loom)` is disabled, so I'll merge this once my review is addressed.

crossbeam-epoch/src/atomic.rs
```diff
@@ -487,7 +506,14 @@ impl<T> Atomic<T> {
     /// }
     /// ```
     pub unsafe fn into_owned(self) -> Owned<T> {
-        Owned::from_usize(self.data.into_inner())
+        #[cfg(loom)]
+        {
+            Owned::from_usize(self.data.unsync_load())
```
jeehoonkang, line 511: I'd like loom to support into_inner.  I believe matching the APIs with std is one of the design goals of loom, right?

crossbeam-epoch/src/atomic.rs
```diff
@@ -149,6 +150,24 @@ impl<T> Atomic<T> {
     ///
     /// let a = Atomic::<i32>::null();
     /// ```
+    #[cfg(loom)]
+    pub fn null() -> Atomic<T> {
```
jeehoonkang, line 154: The reason is loom doesn't support const fn, right?  Does it now (at version 0.4)?

crossbeam-epoch/src/collector.rs
```diff
@@ -109,7 +109,7 @@ impl fmt::Debug for LocalHandle {
     }
 }
 
-#[cfg(test)]
+#[cfg(all(test, not(loom)))]
```
jeehoonkang, line 112: Tests are excluded from loom.  Is it intended?

crossbeam-epoch/src/internal.rs
```diff
@@ -35,10 +35,11 @@
 //! Ideally each instance of concurrent data structure may have its own queue that gets fully
 //! destroyed as soon as the data structure gets dropped.
 
-use core::cell::{Cell, UnsafeCell};
+use crate::concurrency::cell::UnsafeCell;
```
jeehoonkang, line 38: UnsafeCell is not technically a concurrent thing.  Is it logically correct to have it under `crate::concurrency`?

crossbeam-epoch/src/lib.rs
```diff
@@ -60,6 +60,85 @@
...
+
+        impl<T> UnsafeCell<T> {
+            #[inline]
+            pub(crate) fn new(data: T) -> UnsafeCell<T> {
+                UnsafeCell(::core::cell::UnsafeCell::new(data))
+            }
+
+            #[inline]
+            pub(crate) fn with<R>(&self, f: impl FnOnce(*const T) -> R) -> R {
```
jeehoonkang, line 114: It seems loom's unsafecell has these methods but std's doesn't.  I'd like to ask if you could write down a few comments on it.

crossbeam-epoch/tests/loom.rs
```diff
@@ -0,0 +1,145 @@
...
+
+        jh.join().unwrap();
+
+        drop(collector);
+    })
+}
+
+#[test]
+fn treiber_stack() {
+    // this is mostly a copy-paste from the example
```
jeehoonkang, line 49: Maybe we can refactor them to have only one copy of stack-

jeehoonkang (changes requested):
Thank you for this PR!  I left a few comments.

crossbeam-epoch/src/atomic.rs
jonhoo: Well, sort of. Loom's `AtomicUsize::new` is simply not `const` because it has to interact with the loom "runtime". Unlike `AtomicUsize::new` from `std`, which is really just a constant literal, loom's version must register the `AtomicUsize` with the model so that it's execution pattern can be tracked.

crossbeam-epoch/src/collector.rs
jonhoo: Yes indeed. When we're running under loom, we _only_ want to run the loom tests that have the loom "runtime" (in the form of `loom::model`).

crossbeam-epoch/src/internal.rs
jonhoo: I agree, but `concurrency` is the module that holds all the stuff that differs between the loom and non-loom contexts, and it felt weird to break out a new, dedicated top-level module just for `UnsafeCell`. We could rename `concurrency` to something else, but I can't immediately think of something better? And to give a (weak) defense for the current scheme, `UnsafeCell` does _usually_ come up in the context of concurrency where you want to unsafely assert that you can access a given type that is technically shared.

crossbeam-epoch/src/atomic.rs
jonhoo: Unfortunately there are still some holes, and `into_inner` is one of them: https://github.com/tokio-rs/loom/issues/117

crossbeam-epoch/tests/loom.rs
jonhoo: Hmm, I don't know of a way to grab code from an example into another file. We could move the code from the example file into the library, but then it'd no longer be an example. Maybe we could do some hacking with `include!`, but I suspect that would be far uglier. If you have ideas I'm all ears!

jonhoo:
@taiki-e Done! Loom tests still pass (in ~11m):
```
Completed in 9690926 iterations
test treiber_stack ... ok
```

jonhoo:
@jeehoonkang I think I addressed your comments, but will leave it to you as to whether you're satisfied with my answers :)

crossbeam-epoch/Cargo.toml
```diff
@@ -41,6 +41,12 @@ cfg-if = "1"
 const_fn = "0.4"
 memoffset = "0.6"
 
+# An internal configuration option used when running loom.
+# This is not part of the public API and crossbeam may make breaking changes
```
taiki-e, line 45: > This is not part of the public API

I thought you were trying to make this an unstable public API used by the downstream crate (https://github.com/crossbeam-rs/crossbeam/issues/486#issuecomment-610977956).
If so,  "An internal configuration" and "not part of the public API"  are not a proper description, and perhaps something like "An unstable configuration" and "not part of the stable public API" preferred.

crossbeam-epoch/src/atomic.rs
```diff
@@ -327,6 +347,7 @@ impl<T: ?Sized + Pointable> Atomic<T> {
     /// let a = Atomic::<i32>::null();
     /// ```
     ///
+    #[cfg(not(loom_crossbeam))]
     #[const_fn(feature = "nightly")]
```
taiki-e, line 351:
```suggestion
    #[const_fn(all(feature = "nightly", not(loom_crossbeam)))]
```

taiki-e: >     #[const_fn(all(feature = "nightly", not(loom_crossbeam)))]

Oh, I haven't implemented https://github.com/taiki-e/const_fn/issues/28 yet. Updated suggestion.

crossbeam-epoch/tests/loom.rs
jeehoonkang: How about just move the treiber's stack to this file?  I'm fine with removing the examples altogether.  @taiki-e ?

jeehoonkang (changes requested):

crossbeam-epoch/tests/loom.rs
taiki-e: I'm also fine with removing the example.

crossbeam-epoch/src/internal.rs
jonhoo: Hmm, yeah, it's really anything that ends up being a synchronization point. We could use `synchronization`?

jonhoo:
I've addressed all comments beyond the naming :)

crossbeam-epoch/src/atomic.rs
```diff
@@ -637,7 +638,17 @@ impl<T: ?Sized + Pointable> Atomic<T> {
     /// }
     /// ```
     pub unsafe fn into_owned(self) -> Owned<T> {
-        Owned::from_usize(self.data.into_inner())
+        #[cfg(loom_crossbeam)]
+        {
+            // loom does not yet support into_inner, so we use unsync_load for now,
```
taiki-e, line 643:
```suggestion
            // FIXME: loom does not yet support into_inner, so we use unsync_load for now,
```

crossbeam-utils/Cargo.toml
```diff
@@ -33,6 +33,12 @@ nightly = []
 cfg-if = "1"
 lazy_static = { version = "1.4.0", optional = true }
 
+# An internal configuration option used when running loom.
+# This is not part of the public API and crossbeam may make breaking changes
+# to the behavior of this configuration flag at any time.
```
taiki-e, line 38:
```suggestion
# Enable the use of loom for concurrency testing.
#
# This configuration option is outside of the normal semver guarantees: minor
# versions of crossbeam may make breaking changes to it at any time.
```

ci/crossbeam-epoch-loom.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+cd "$(dirname "$0")"/../crossbeam-epoch
+set -ex
+
+export RUSTFLAGS="-D warnings --cfg=loom_crossbeam"
+
+env LOOM_MAX_PREEMPTIONS=2 cargo test --test loom --features sanitize --release -- --nocapture
```
jeehoonkang, line 8: I'd like to ask if you could write here why LOOM_MAX_PREEMPTIONS=2 is good enough.  (Or =3 is too costly.)

jeehoonkang (commented):
Again, I sincerely thank you for this effort.  Now I'm mostly fine with this PR.

- I left a few minor comments that can be easily addressed.

- Regarding the name of `crate::concurrency`: I think the module is not fundamentally about concurrency but about the core platform functionalities crossbeam depends on.  So I suggest renaming it to `crate::core`, `crate::basis`, or something that represents that idea.

jonhoo:
I went with `primitive`, though I think the word `overloaded` could also work. I went with `primitive` over `overloaded` since the types are only _sometimes_ overloaded, but including overloaded in the name might make a reader think it is _always_ overloaded.

taiki-e (approved):
Looks great to me! Thanks so much for working on this!

taiki-e:
Hmm, seems there are conflicts:
crossbeam-487
@jonhoo Could you rebase?

jonhoo:
Err, rebasing this will probably be pretty painful. Do you mean squash? I merged with master just now, and that went off without a hitch, so it should just be a matter of using GitHub's "squash and merge" button I think.

taiki-e:
Oh, I didn't notice about "squash and merge" worked. I'll use that. Thanks for pointing that out.

taiki-e:
@jeehoonkang: Do you want another chance at reviewing this before I merge it?

taiki-e:
> * Rename `cfg(loom)` to something like `cfg(loom_crossbeam)` to avoid conflict with other crates that using `cfg(loom)`. (I don't know what the `loom`'s recommendation for this is. If there is a recommendation, it is better to use it.)

Oh, I noticed that loom has a PR that tries to recommend the use of cfg like `${project_name}_loom`, not `loom_${project_name}`. 
@jonhoo That PR hasn't been merged yet, but should we probably follow it?

jonhoo:
I don't feel strongly whether the name goes first or last, but I agree that crossbeam should probably follow whatever convention ends up being documented.

taiki-e:
I've filed #658 to rename cfg(loom_crossbeam) to cfg(crossbeam_loom). Also, I've approved tokio-rs/loom#159.

matklad:
Is `[target.'cfg(crossbeam_loom)'.dependencies]` in Cargo.toml the right pattern here? With this, loom (and its deps) go into rev-dep Cargo.locks unconditionally. If this were a Cargo feature (`--private--loom`), then it would be invisible to consumers. 

Context: was wondering about 

```
  Downloaded generator v0.6.24
  Downloaded loom v0.4.0
  Downloaded 2 crates (81.8 KB) in 1.08s
```

output when building rust-analyzer.

taiki-e:
@matklad 
> Is `[target.'cfg(crossbeam_loom)'.dependencies]` in Cargo.toml the right pattern here?

Yes. see Alex's [this comment](https://github.com/rust-lang/cargo/pull/7660#issuecomment-561733620).

> With this, loom (and its deps) go into rev-dep Cargo.locks unconditionally. 

As far as I know, it will be downloaded but not actually compiled. See also https://github.com/tokio-rs/bytes/issues/400 and https://github.com/tokio-rs/bytes/issues/411.
That said, I agree that the cargo messages and crates.io UI are confusing...

> If this were a Cargo feature (`--private--loom`), then it would be invisible to consumers.

Unfortunately, as far as I know, (at least in the v1 resolver) the cargo feature doesn't provide enough functionality to support loom. See also https://github.com/crossbeam-rs/crossbeam/pull/638

matklad:
>As far as I know, it will be downloaded but not actually compiled.

compilation is not the only cost: I read Cargo.lock from time to time, and I read .lock diffs regularly. I care a lot about keeping those as small as possible. 

>Unfortunately, as far as I know, (at least in the v1 resolver) the cargo feature doesn't provide enough functionality to support loom.

Can `loom` then be gated by *both* feature and cfg? The following seems to work:

```
[target.'cfg(loom)'.dependencies]
loom = {version="*", optional=true}
```

taiki-e:
> Can `loom` then be gated by _both_ feature and cfg? The following seems to work:
> 
> ```
> [target.'cfg(loom)'.dependencies]
> loom = {version="*", optional=true}
> ```

Oh, I didn't know it works. I would like to switch to it if it works.

taiki-e:
will be fixed by #666

PR #490, opened 2020-04-10
epoch: Move away from pointers as usize

jonhoo:
Pointers stored as `usize` tend to cause miri to lose pointer provenance
tracking, which means we can't take advantage of its checking! See also
the discussion at
https://github.com/rust-lang/miri/issues/940#issuecomment-612097677.

This does not yet compile since `AtomicPtr` does not have `fetch_*`
methods. They were added and then removed from the standard library back
in the day (https://github.com/rust-lang/rust/pull/10154), but I think
the reason they were removed has now been remedied, so they will
hopefully come back again!

cc @RalfJung

### DIFF
--- a/crossbeam-epoch/src/atomic.rs
+++ b/crossbeam-epoch/src/atomic.rs
@@ -6,7 +6,7 @@ use core::marker::PhantomData;
 use core::mem;
 use core::ops::{Deref, DerefMut};
 use core::ptr;
-use core::sync::atomic::{AtomicUsize, Ordering};
+use core::sync::atomic::{AtomicPtr, Ordering};
 
 use crossbeam_utils::atomic::AtomicConsume;
 use guard::Guard;
@@ -103,15 +103,20 @@ fn low_bits<T>() -> usize {
 ///
 /// `tag` is truncated to fit into the unused bits of the pointer to `T`.
 #[inline]
-fn data_with_tag<T>(data: usize, tag: usize) -> usize {
-    (data & !low_bits::<T>()) | (tag & low_bits::<T>())
+fn data_with_tag<T>(data: *mut T, tag: usize) -> *mut T {
+    // transmute preserves pointer provenance
+    let data: usize = core::mem::transmute(data);
+    core::mem::transmute((data & !low_bits::<T>()) | (tag & low_bits::<T>()))
 }
 
 /// Decomposes a tagged pointer `data` into the pointer and the tag.
 #[inline]
-fn decompose_data<T>(data: usize) -> (*mut T, usize) {
-    let raw = (data & !low_bits::<T>()) as *mut T;
-    let tag = data & low_bits::<T>();
+fn decompose_data<T>(data: *mut T) -> (*mut T, usize) {
+    // transmute preserves pointer provenance
+    let data: usize = core::mem::transmute(data);
+    let raw = core::mem::transmute(data & !low_bits::<T>());
+    // explicitly erase provenance of tag
+    let tag = data as usize & low_bits::<T>();
     (raw, tag)
 }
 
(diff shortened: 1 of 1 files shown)


crossbeam-epoch/src/atomic.rs
```diff
@@ -103,15 +103,20 @@ fn low_bits<T>() -> usize {
 ///
 /// `tag` is truncated to fit into the unused bits of the pointer to `T`.
 #[inline]
-fn data_with_tag<T>(data: usize, tag: usize) -> usize {
-    (data & !low_bits::<T>()) | (tag & low_bits::<T>())
+fn data_with_tag<T>(data: *mut T, tag: usize) -> *mut T {
+    // transmute preserves pointer provenance
+    let data: usize = core::mem::transmute(data);
+    core::mem::transmute((data & !low_bits::<T>()) | (tag & low_bits::<T>()))
```
RalfJung, line 109: If the transmute preserved provenance, doing integer operations on the pointer will still kill it unless we do really awful things (things that we used to do but that I very happily removed when integer-ptr-casts got properly supported).

jonhoo:
Note in particular the (backwards-compatible) change to `Pointer`, which is sad, but pretty much necessary.

crossbeam-epoch/src/atomic.rs
jonhoo: Hmm, interesting. Is there any way to support provenance through tagged pointers then?

RalfJung: See `ptr_int_arithmetic` removed [in this commit](https://github.com/rust-lang/miri/commit/82da68c5caca6b534d77c01cbcf9d1ed0a6ef496) for the awful things we used to do.

RalfJung: I opened https://github.com/rust-lang/miri/issues/1318 to track the Miri-side issue (and to avoid having this discussion in 4 threads in parallel^^).

taiki-e:
@jonhoo What is the current status of this PR? It seems both https://github.com/rust-lang/rust/pull/71004 and https://github.com/rust-lang/miri/issues/1318 were closed.

jonhoo:
It's.. complicated. I suspect that any time we do `usize` arithmetic on pointers (i.e., not through the standard pointer operations), we're going to run into issues with miri. @RalfJung is probably a better person to ask to make the judgement call on this.

RalfJung:
Miri (in its default settings) does support ptrs being cast to ints and back. Yes, provenance is lost, but that is a "feature" of int-to-ptr casts and Miri supports this. So you should be able to benefit from Miri's checking regardless.

However, the memory leak checker requires provenance to be intact to "recognize" pointers that it has to follow to memory that is still alive. You can either disable the memory leak checker (`-Zmiri-ignore-leaks`) or call the extern function `miri_static_root` to mark some memory as "not considered leaked even if it still exists when the program is done". If there are more elaborate APIs that you need to get proper memory leak checking, we can also try to offer those.

Furthermore, `-Zmiri-track-raw-pointers` puts Miri into a mode where more aliasing issues are found, but int-to-ptr casts are not supported any more. But there's a lot of code out there that is not compatible with this code, so that seems fine. However, if you are curious, there are also some hacks one can use to "finesse" pointer provenance such that this should still work, at least in many cases.

So... I guess what I am saying is, I am not entirely sure which problem this PR is solving. ;) IIRC it originated from discussions around the leak checker? It probably makes more sense to first get `MIRIFLAGS=-Zmiri-ignore-leaks cargo miri test` to work on this repo, and put that under CI, before trying to appease the leak checker.

taiki-e:
@RalfJung Thanks for the detailed explanation!

> It probably makes more sense to first get `MIRIFLAGS=-Zmiri-ignore-leaks cargo miri test` to work on this repo, and put that under CI, before trying to appease the leak checker.

Agree. ...and I have a draft PR for this: https://github.com/crossbeam-rs/crossbeam/pull/578

PR #491, opened 2020-04-19
Add len() methods to deque

stjepang:
The implementation for `Injector::len()` was simply copied from `SegQueue::len()`.


dignifiedquire:
Yes please.

jeehoonkang:
bors r+

(Not related to this PR, but would you please add me as the owner of the `crossbeam-queue` crate?  I was not able to release a new version.)

stjepang:
@jeehoonkang You already seem to be an owner :) https://crates.io/crates/crossbeam-deque

Perhaps you need to set up your crates.io token again?

PR #492, opened 2020-04-22
Add a ping-pong benchmark to crossbeam-channel

vgatherps:
Ping-pong is an important benchmark since it measure latency
characteristics of the transport as opposed to plain throughput
characteristics. In addition, it also is a decent test of the
empty-channel read cost

### DIFF
--- a/crossbeam-channel/benchmarks/chan.rs
+++ b/crossbeam-channel/benchmarks/chan.rs
@@ -43,6 +43,26 @@ fn spsc(cap: Option<usize>) {
     .unwrap();
 }
 
+fn ping_pong(cap: Option<usize>) {
+    let (txa, rxb) = new(cap);
+    let (txb, rxa) = new(cap);
+
+    crossbeam::scope(|scope| {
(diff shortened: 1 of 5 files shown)


vgatherps:
While I'm sure it's possible, it wasn't actually clear how to do a clean ping-pong benchmark with the futures channels so I haven't added one yet.

I also didn't update the benchmark png since I have a different machine, and don't necessarily want to push a totally new set of baseline numbers. The generated one is here: https://i.imgur.com/O9b9CLx.png

SUPERCILEX:
@taiki-e any reason this wasn't merged?

bsilver8192:
This is a more complete version of what I wrote for demonstrating #821. I think having it as an official benchmark to measure performance of is valuable.

taiki-e:
Sorry for the late response. I've read the first half of the https://github.com/crossbeam-rs/crossbeam/pull/492#issuecomment-617969071 and assumed this was a WIP state. Otherwise, I think adding such a benchmark is fine.

PR #495, opened 2020-05-03
Relax racy fairness test

### LINKED ISSUE #477: channel: ready::fairness2 test is racy
The test occasionally [fails in CI](https://travis-ci.org/crossbeam-rs/crossbeam/jobs/651567571#L538) and [locally](https://github.com/crossbeam-rs/crossbeam/pull/474#issuecomment-589250957).

Reproduce by running:
```console
$ for (( i=1 ;; i+=1 )); do echo "*** test $i ***"; cargo test --test ready fairness || break; done
```

To quote @cuviper:

> It seems to fail even more frequently with `--release`, so maybe the test just needs more allowance?

jonhoo:
Fixes #477

### DIFF
--- a/crossbeam-channel/tests/ready.rs
+++ b/crossbeam-channel/tests/ready.rs
@@ -772,7 +772,7 @@ fn fairness1() {
 
 #[test]
 fn fairness2() {
-    const COUNT: usize = 10_000;
+    const COUNT: usize = 100_000;
 
     let (s1, r1) = unbounded::<()>();
     let (s2, r2) = bounded::<()>(1);
@@ -831,7 +831,7 @@ fn fairness2() {
                 }
             }
         }
-        assert!(hits.iter().all(|x| x.get() >= COUNT / hits.len() / 10));
+        assert!(hits.iter().all(|x| x.get() > 0));
     })
     .unwrap();
 }


stjepang:
This looks good to me, but seems like CI is broken. I'm inclined to merge anyway and take care of that later...

taiki-e:
> seems like CI is broken

It seems due to signal-hook's MSRV was increased to 1.31: https://github.com/vorner/signal-hook/commit/e63edcf55f94ba33f8f884948b610b3cad74c86d

taiki-e:
> It seems due to signal-hook's MSRV was increased to 1.31: [vorner/signal-hook@e63edcf](https://github.com/vorner/signal-hook/commit/e63edcf55f94ba33f8f884948b610b3cad74c86d)

So we need to either:
* Downgrade signal-hook's patch version in MSRV CI (add `cargo update -p signal-hook --precise 0.1.13`):
  https://github.com/crossbeam-rs/crossbeam/blob/485971eae072592d07845e387bd5a0df87facbd7/.github/workflows/ci.yml#L40-L44

* Bump crossbeam's MSRV to 1.31

jonhoo:
One upside of the latter is that it would also mean https://github.com/crossbeam-rs/crossbeam/pull/487 no longer needs to bump the MSRV. But that's definitely minor.

taiki-e:
I have tried the first way (downgrade deps) locally, but it seems that the MSRV of so many dependencies is actually increasing, so this seems to be difficult...

jonhoo:
One issue here is that the docs say:
> The minimum supported Rust version is 1.28. Any change to this is considered a breaking change.

That is a _very_ strict policy. The policy should (probably) instead be "six months" or something? But as it stands, we'd need to issue new major releases of all the crossbeam crates to do that... It also definitely gets weird since people will have to run through all sorts of hoops to _actually_ get it running with 1.28 — so are we even _really_ supporting MSRV 1.28?

taiki-e:
I agree with the current policy needs to be changed as suggested in https://github.com/crossbeam-rs/crossbeam/pull/444#discussion_r343737057.


> It also definitely gets weird since people will have to run through all sorts of hoops to _actually_ get it running with 1.28 — so are we even _really_ supporting MSRV 1.28?

I mentioned this in https://github.com/crossbeam-rs/crossbeam/pull/428#issuecomment-542269227, but I think it is ok to set MSRV based on the compilable version (without downgrading the dependencies).

stjepang:
Let's merge and fix CI in a follow-up PR.

PR #496, opened 2020-05-10
Add Receiver::new_sender and Sender::new_receiver

Thomasdezeeuw:
I've added some basic tests, let me know if it need more extensive tests. I'm also not sure the changes I've made are sufficient or if they also need changes inside the various flavours.

### DIFF
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -1007,6 +1007,23 @@ impl<T> Receiver<T> {
             _ => false,
         }
     }
+
+    /// Returns a new sender that sends to this channel.
+    ///
+    /// This returns `None` for channels that don't have a `Sender` (the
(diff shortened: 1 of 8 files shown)


Thomasdezeeuw:
The CI failure seems to be unrelated.

BurntSushi:
Could you say more about what the use case for these methods is?

Thomasdezeeuw:
I have a use case where I'm holding a `Receiver` and a `Sender` in the same structure, where the `Sender` is only used to be cloned and shared. Since both are a pointer to the same structure it just wastes memory. I'm also using thousands of these structures so even small reductions add up.

Thomasdezeeuw:
Are there are reviewers available for this? I really need this feature.

taiki-e:
Seems this needs to clarify what happens when called after the channel is disconnected.

Thomasdezeeuw:
This has taken too long and I've moved on to using a different crate.

Issue #256, opened 2018-12-05
Breaking changes before releasing crossbeam-channel 1.0


stjepang:
In early 2019, I'd like to publish version 1.0 of `crossbeam-channel`.

This issue tracks the potential breaking changes for the final release. Please upvote/downvote changes if you like/dislike them.

stjepang:
Change the signature of `Select::recv()` and `Select::send()` so that we use a custom type `Token` (similar to [`Token`](https://docs.rs/mio/0.6.16/mio/struct.Token.html) in `mio`) instead of indices of type `usize`. The `Token` type would wrap the index so that we get some type safety.

The interface of `Select::recv()` would take one of these forms:

1. `fn recv<T>(&mut self, r: &Receiver<T>) -> Token`
2. `fn recv<T>(&mut self, r: &Receiver<T>, token: Token)`

stjepang:
Use the word "closed" instead of "disconnected". That means we'd rename variants like `TryRecvError::Disconnected` to `TryRecvError::Closed`.

Prior art:

* [`std::sync::mpsc`](https://doc.rust-lang.org/std/sync/mpsc/index.html) uses "disconnected".
* [`chan`](https://docs.rs/chan/0.1.23/chan/) uses "closed".
* [`tokio-channel`](https://docs.rs/tokio-channel/0.1.0/tokio_channel/) uses "closed".
* [`futures::sync::mpsc`](https://docs.rs/futures/0.1.25/futures/sync/mpsc/index.html) uses "closed".
* Golang uses "closed".

See also: https://github.com/rust-lang-nursery/futures-rs/issues/1346

stjepang:
Not a breaking change, but a new feature.

Add methods `Sender::disconnect()` and `Receiver::disconnect()` to explicitly close channels.

See also: https://github.com/crossbeam-rs/crossbeam/issues/236

stjepang:
Not a breaking change, but a new feature.

Add methods `Sender::is_disconnected()` and `Receiver::is_disconnected()` that check whether the channel has been disconnected.

Thomasdezeeuw:
@stjepang I'm sure not if you want to keep this as a meta issue or also discuss in it, but here are some of my toughs.

Regarding the use of a token/index in `Select::recv()`, can we do better then the generic name `Token`? For mio I proposed `EventId`/`event::Id` (https://github.com/carllerche/mio/issues/905).

[...]

PR #497, opened 2020-05-10
Add {Sender, Receiver}::is_connected

Thomasdezeeuw:
Checks if the other side of the channel is connected.

Updates https://github.com/crossbeam-rs/crossbeam/issues/256#issuecomment-444620303

### DIFF
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -440,6 +440,16 @@ impl<T> Sender<T> {
         }
     }
 
+    /// Returns `true` if one or more receivers are connected, `false`
+    /// otherwise.
+    pub fn is_connected(&self) -> bool {
+        match &self.flavor {
+            SenderFlavor::Array(chan) => chan.is_connected(),
+            SenderFlavor::List(chan) => chan.is_connected(),
+            SenderFlavor::Zero(chan) => chan.is_connected(),
+        }
+    }
+
     /// Returns `true` if the channel is empty.
     ///
     /// Note: Zero-capacity channels are always empty.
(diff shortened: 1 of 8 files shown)


Thomasdezeeuw:
The CI failure seems unrelated.

BurntSushi:
What is the use case for methods like this? They seem inherently problematic to use. _At the very least_, the docs on these methods should discuss this.

Thomasdezeeuw:
The use case is I want to know before creating a message if the receiving end is still connected, if not I can avoid doing any work for a message that is never going to be received.

> They seem inherently problematic to use. 

What do you see as problematic? Of course the method is inherently race-y as the moment the method returns `true` it could already be `false`, but its still an optimisation I consider useful.

BurntSushi:
Its raciness is exactly why I consider it problematic. I don't maintain this library, but I personally wouldn't accept a new API item like this without a reasonable benchmark. If we must have them, the docs should do their best to de-emphasize using this routine and explain their niche use case.

Thomasdezeeuw:
What do you want to benchmark? If creating a message is expansive, then its always going to be faster to not create a message than creating it and the dropping it (after failing to send it).

BurntSushi:
That sounds like a micro-benchmark to me. I'd prefer seeing something more holistic.

Thomasdezeeuw:
Sorry @BurntSushi but I have no idea what you want from me... I asked what you want me to benchmark but you haven't really given an answer to that.

BurntSushi:
@Thomasdezeeuw I'm not the maintainer of this library. I'm just saying what I personally would like to see because I personally find these APIs to be of somewhat dubious value. Which in turn, to me, means they should be _really well motivated_ before adding them. Namely, I'd like to see a holistic example of a real use case where these sorts of APIs would help. I would probably expect some kind of latency/throughput figures with/without these APIs for example. It may be hard to provide this.

Thomasdezeeuw:
@stjepang could you take a look at this? In https://github.com/crossbeam-rs/crossbeam/issues/256#issuecomment-444620303 you mentioned this feature.

taiki-e:
I don't have a very strong opinion on whether to add this API at this time, but I think it's better to be `is_disconnected` instead of `is_connected`. (It matches existing APIs such as error type.)

Thomasdezeeuw:
@taiki-e I'm ok with changing it, but would that mean this would be considered to be merged?

taiki-e:
@Thomasdezeeuw I think documentation about what users should be notes of when using this API is needed to merge this PR, not only renaming. FYI: https://github.com/golang/go/issues/20680 and https://github.com/golang/go/issues/30906 have some specific explanations about the problems of similar features.

Thomasdezeeuw:
This has taken too long and I've moved on to using a different crate.

Issue #503, opened 2020-05-19
Publish version 1.0


stjepang:
Crossbeam hasn't changed much in a long time. It may be time to publish 1.0 :)

My opinion is that all crates except for crossbeam-epoch have "peaked" and reached maturity. The thing about crossbeam-epoch is that its API is still not perfect and needs some improvement, mainly around custom collectors. But, even considering that, the API hasn't changed in a very long time, so why not just go for 1.0? If we decide to actually change it in the future, we can bump to 2.0.

I proposed we do the following now:

- Bump minimum required Rust version to 1.36.0 (the one with `MaybeUninit`).
- Fix CI.
- Decide what to do with #416 
- Merge #431
- Bump all crate versions to 1.0 and publish.

What does everyone think?

BurntSushi:
Exciting! SGTM. :)

jonhoo:
Oooh, very excited for this!

The one point I think warrants some discussion is the MSRV policy, which I know is a topic many people have opinions on. As we've seen several times with CI breaking, other crates do not consider MSRV a part of their stability guarantee, so saying just "1.36" is inevitably going to cause us headache down the line. If that line is in the README, we need to commit to it for _all_ of 1.0. Of course, if we _don't_ commit to this, then we're a part of the problem. There isn't a good solution to this at the moment I think, but the least painful path probably lies in not committing to a _particular_ version for the entire time of 1.0.

This has also been discussed in the past in https://github.com/crossbeam-rs/crossbeam/pull/428#issuecomment-542269227, https://github.com/crossbeam-rs/crossbeam/pull/458#issuecomment-576079793, and https://github.com/crossbeam-rs/crossbeam/pull/495#issuecomment-623162147.

My suggestion here is to say something like

> Crossbeam supports stable Rust releases going back at least six months, and every time the minimum supported Rust version is increased, a new minor version is released. Currently, the oldest supported version is 1.36.

This is definitely a weakly held opinion, and is mostly motivated by the several issues we have seen over the past months (and years) where CI breaks, or where PRs need to make somewhat awkward workarounds to avoid bumping MSRV (`maybe-uninit` being an example). Not to mention all the ways in which we don't _actually_ have the stated MSRV when different features are used (like `alloc`).

stjepang:
@jonhoo Good point, I agree with that. We should stop considering bumping MSRV a breaking change.

BurntSushi:
[...]

PR #504, opened 2020-05-19
Bump MSRV to 1.36

taiki-e:
Maybe this can fix the CI.

cc #503


stjepang:
Almost there - looks like this just needs a `cargo fmt` :)

taiki-e:
CI passed.

taiki-e:
Ah, wait, we can remove some build scripts and cfgs: https://github.com/crossbeam-rs/crossbeam/blob/c239d8f318c022b2c1bce68b8d20e468ea4f1197/crossbeam-epoch/build.rs#L5-L7

taiki-e:
I added the folloing changes:
* Remove has_min_const_fn cfg and remove a build script of crossbeam-epoch.
* Use std::mem::MaybeUninit directly and remove maybe-uninit dep.

README.md
```diff
@@ -98,7 +98,7 @@ extern crate crossbeam;
 
 ## Compatibility
 
-The minimum supported Rust version is 1.28. Any change to this is considered a breaking change.
+The minimum supported Rust version is 1.36. Any change to this is considered a breaking change.
```
taiki-e, line 101: > Any change to this is considered a breaking change.
 
We should change this.

taiki-e:
Updated MSRV policy based on @jonhoo's [suggestion](https://github.com/crossbeam-rs/crossbeam/issues/503#issuecomment-631106295).
> Crossbeam supports stable Rust releases going back at least six months, and every time the minimum supported Rust version is increased, a new minor version is released. Currently, the oldest supported version is 1.36.

jonhoo:
Do we want to give explicit MSRV numbers for the different features too? At least where they differ. Or maybe 1.36 covers all features? Do we want CI to check that MSRV holds for _all_ features?

crossbeam-skiplist/README.md
```diff
@@ -35,7 +35,7 @@ extern crate crossbeam_skiplist;
 
 ## Compatibility
 
-The minimum supported Rust version is 1.28. Any change to this is considered a breaking change.
+Crossbeam Skiplist supports stable Rust releases going back at least six months, and every time the minimum supported Rust version is increased, a new minor version is released. Currently, the oldest supported version is 1.36.
```
jeehoonkang, line 38: I prefer to use "minimum supported Rust version" instead of "oldest supported version", as the former is a more widely used term.

taiki-e: Done

stjepang:
@jonhoo I think MSRV policy should cover all features (except for the `nightly` feature, of course).

jonhoo:
@jeehoonkang Ah, sorry, what I meant was that we'll want to make sure that the MSRV covers _all_ features, or if it doesn't, document which features have which MSRV.

@stjepang If that's the case, that's awesome! Do we check this on CI at the moment? If not, might be a good thing to add so we don't accidentally break that later on. I think @taiki-e's [`cargo-hack`](https://github.com/taiki-e/cargo-hack) may be able to help with that.

taiki-e:
> I think @taiki-e's [`cargo-hack`](https://github.com/taiki-e/cargo-hack) may be able to help with that.

Yup, by using cargo-hack, we can check this with the following one command, and will not miss it if features are added in the future.

```shell
cargo +1.36.0 hack check --all --each-feature --no-dev-deps --ignore-private --skip nightly
```

jonhoo:
That seems like a _fantastic_ thing to have in CI!

PR #508, opened 2020-05-20
Make std feature to depend on alloc feature

taiki-e:
`alloc` crate is available on 1.36, see https://github.com/crossbeam-rs/crossbeam/pull/368#discussion_r283091668 for why it was previously separated.

This is a breaking change proposed in #503.


stjepang:
There's an unused `#[macro_use]`.

taiki-e:
CI passed.

crossbeam-epoch/src/lib.rs
```diff
@@ -61,20 +61,14 @@
...
-        extern crate alloc;
-    } else if #[cfg(feature = "std")] {
-        extern crate std as alloc;
-    }
-}
-
 #[cfg_attr(feature = "nightly", cfg(target_has_atomic = "ptr"))]
 cfg_if! {
-    if #[cfg(any(feature = "alloc", feature = "std"))] {
+    if #[cfg(feature = "alloc")] {
```
jeehoonkang, line 71: - If `alloc` feature is on and `std` is off, then the compilation will fail, right?  I'd like to avoid such cases..
- I wonder if we can just remove the `alloc` feature and always require `alloc`.  Is it possible?

taiki-e: > * If `alloc` feature is on and `std` is off, then the compilation will fail, right?  

No. Items in this `if` block are not require `std`. items that require `std` is in `if #[cfg(feature = "std")]` block: https://github.com/crossbeam-rs/crossbeam/blob/e782e91db8916b934eedb9714ef57ad08a619a0a/crossbeam-epoch/src/lib.rs#L92-L100

> I wonder if we can just remove the `alloc` feature and always require `alloc`. Is it possible?

Yes, the current crossbeam-epoch can do this, but I'd prefer to leave `alloc` feature for forward compatibility. (As long as `alloc` or `std` feature is not enabled, crossbeam-epoch with no-default-feature does not depend on `alloc` crate.)

jeehoonkang: Oh sorry, I meant to ask what if `alloc` is off and `std` is on.  Then `mod default` will result in a compilation error, I guess.

Further, I don't understand the relationship between `alloc` and forward compatibility.  May I ask if you could further explain this point?

taiki-e: @jeehoonkang 

> I meant to ask what if `alloc` is off and `std` is on. Then `mod default` will result in a compilation error, I guess.

Even if the user does not have `alloc` feature enabled, `alloc` feature is always enabled when `std` feature is enabled because [`std` feature depends on `alloc` feature](https://github.com/crossbeam-rs/crossbeam/pull/508/files#diff-a3fcd04172682b8bc7a8532346160690R21).

> I don't understand the relationship between `alloc` and forward compatibility. May I ask if you could further explain this point?

If we add a public API that does not depend on `alloc` to crossbeam-epoch, in order to use it in no-std environment where `alloc` crate cannot be used, it is necessary to separate APIs that require `alloc` crate by a feature gate.

However, this is a breaking change because adding this feature will prevent crate, which currently uses crossbeam-epoch with no-default-feature, from accessing APIs that require `alloc` crate.

To prevent this breaking change, from the beginning, APIs that require `alloc` crate should be under of a feature flags.

taiki-e:
> If we add a public API that does not depend on `alloc` to crossbeam-epoch, in order to use it in no-std environment where `alloc` crate cannot be used, it is necessary to separate APIs that require `alloc` crate by a feature gate.
> 
> However, this is a breaking change because adding this feature will prevent crate, which currently uses crossbeam-epoch with `default-feature = false`, from accessing APIs that require `alloc` crate.

As a similar issue, adding no-std support to crossbeam-channel and crossbeam-deque, which currently do not have feature flags and depend on std, is actually a breaking change for crates using these crate with `default-feature = false`. (Actually, this is why [futures-io has std feature](https://github.com/rust-lang/futures-rs/blob/95d5801c0d95e9c9c07b96b74ec3ecb3a269585f/futures-io/Cargo.toml#L14-L16)).

So, unless it's absolutely impossible for these crates to add support no-std, I tend to prefer to add `std` feature to these crates...
@stjepang @jeehoonkang Any thoughts about this?

crossbeam-epoch/src/lib.rs
jeehoonkang: Thank you for your kind explanation.  Now I can quite well understand your points.

One suggestion I'd like to make is that maybe we can write down a comment saying that `std` automatically imports `alloc`?  By doing so we can prevent someone like me from having difficulties why this should compile well always.

jeehoonkang:
- Maybe in the future we want `crossbeam-deque` to support `no-std` with a custom allocator.  So I'd say it's plausible to support `no-std` in `crossbeam-deque`, e.g.

- I think it's relatively easy to break backward compatibility for the time being, as we didn't reach 1.0 yet.  I prefer to have a cleaner feature configurations regarding `std`.  @taiki-e what do you think is the best practice?  Is the `std` feature the clean and widely used measure for this?

stjepang:
> So, unless it's absolutely impossible for these crates to add support no-std, I tend to prefer to add `std` feature to these crates...

Sounds good to me, let's do it.

taiki-e:
@jeehoonkang 

> what do you think is the best practice? 

There is a discussion on this in  https://github.com/rust-lang/api-guidelines/issues/189, but as far as I know, it's the only way.

Also, it may be nice to [prevent users from accidental compilation without `std`](https://github.com/rust-lang/regex/blob/f98eea5d9eb97e4fc9efcf31963364eb54f98610/src/lib.rs#L620-L621):

```rust
#[cfg(not(feature = "std"))]
compile_error!("`std` feature is currently required to build this crate");
```

> Is the `std` feature the clean and widely used measure for this?

For example [regex is one of the 1.0+ crates that do this](https://github.com/rust-lang/regex/blob/f98eea5d9eb97e4fc9efcf31963364eb54f98610/Cargo.toml#L38-L43), but I'm not sure how many crates actually use this way.

crossbeam-epoch/src/lib.rs
taiki-e: > One suggestion I'd like to make is that maybe we can write down a comment saying that `std` automatically imports `alloc`? By doing so we can prevent someone like me from having difficulties why this should compile well always.

Sounds good, I'll add descriptions of the feature flag.

taiki-e:
> Also, it may be nice to [prevent users from accidental compilation without `std`](https://github.com/rust-lang/regex/blob/f98eea5d9eb97e4fc9efcf31963364eb54f98610/src/lib.rs#L620-L621):

Ah, this approach doesn't work with crossbeam, as we depend on these crates via `nightly` feature.

https://github.com/crossbeam-rs/crossbeam/blob/24ab0fd5cf0c596aa6cf7ffa71851c2dce987c90/Cargo.toml#L24

taiki-e:
Also, I noticed that crossbeam-utils' alloc feature isn't really doing anything, so I removed it.

taiki-e:
Added document to feature flags (https://github.com/crossbeam-rs/crossbeam/pull/508/commits/86a6aa6fb8a13c3924f9a4422dda347067802ad8):

```toml
# Enable to use APIs that require `std`.
# This is enabled by default.
std = ["alloc"]

# Enable to use APIs that require `alloc`.
# This is enabled by default and also enabled if the `std` feature is enabled.
alloc = []

# Enable to use of unstable functionality.
# This is disabled by default and requires recent nightly compiler.
# Note that this is outside of the normal semver guarantees and minor versions
# of crossbeam may make breaking changes to them at any time.
nightly = []
```

crossbeam-deque/src/deque.rs
```diff
@@ -0,0 +1,1987 @@
+use std::cell::{Cell, UnsafeCell};
```
taiki-e, line 1: This file is increasing the diff, but it really just moved the all implementation from lib.rs.

jeehoonkang:
I'm happy with this PR.  @stjepang do you have more to discuss?  If not, I'll merge it 2 days later.

Thanks!

PR #510, opened 2020-05-22
Clippy

jeehoonkang:
I ran `cargo clippy` and fixed a few warnings.  In the long run, I'd like to clippy in the CI test.

Breaking change: constant `Seqlock::INIT` is changed to const fn ~~`Seqlock::init()`~~ `Seqlock::new()`.


jeehoonkang:
I fixed CI fails except for: https://github.com/crossbeam-rs/crossbeam/pull/510/checks?check_run_id=700024796#step:4:438  It seems a transient proc-macro bug.

crossbeam-epoch/src/lib.rs
```diff
@@ -58,6 +58,9 @@
 #![warn(missing_debug_implementations)]
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(feature = "nightly", feature(cfg_target_has_atomic))]
+#![cfg_attr(feature = "nightly", feature(const_fn))]
```
taiki-e, line 61: Which function needs this feature?

crossbeam-utils/Cargo.toml
```diff
@@ -24,6 +24,7 @@ alloc = []
 [dependencies]
 cfg-if = "0.1.2"
 lazy_static = { version = "1.1.0", optional = true }
+arr_macro = "0.1.3"
```
taiki-e, line 27: As this is a proc-macro, it may greatly increase the compile time. cc @stjepang

taiki-e:
I re-run CI, but looks like there is another test fail: https://github.com/crossbeam-rs/crossbeam/pull/510/checks?check_run_id=700081685

crossbeam-epoch/src/lib.rs
jeehoonkang: I just removed it.  Thanks!

crossbeam-utils/Cargo.toml
jeehoonkang: I removed the dependency and manually expanded `arr![SeqLock::init(), 97]`.

PR #511, opened 2020-05-22
Migrate to Rust 2018

taiki-e:
This migrates all crates to Rust 2018.

The first commit contains the minimal changes needed for the migration. This was mainly done by `cargo fix --editon`.
The second commit migrates the code to be the Rust 2018 idiomatic style. This mainly based on suggestions from `rust_2018_idioms` lint.

### DIFF
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -7,11 +7,11 @@ use std::panic::{RefUnwindSafe, UnwindSafe};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
-use context::Context;
-use counter;
-use err::{RecvError, RecvTimeoutError, SendError, SendTimeoutError, TryRecvError, TrySendError};
-use flavors;
-use select::{Operation, SelectHandle, Token};
+use crate::context::Context;
+use crate::counter;
+use crate::err::{RecvError, RecvTimeoutError, SendError, SendTimeoutError, TryRecvError, TrySendError};
+use crate::flavors;
+use crate::select::{Operation, SelectHandle, Token};
 
 /// Creates a channel of unbounded capacity.
 ///
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -9,7 +9,9 @@ use std::time::{Duration, Instant};
 
 use crate::context::Context;
 use crate::counter;
-use crate::err::{RecvError, RecvTimeoutError, SendError, SendTimeoutError, TryRecvError, TrySendError};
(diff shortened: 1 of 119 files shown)


crossbeam-channel/src/channel.rs
```diff
@@ -1092,22 +1090,22 @@ impl<T> IntoIterator for Receiver<T> {
 ///
 /// assert_eq!(v, [1, 2, 3]);
 /// ```
-pub struct Iter<'a, T: 'a> {
+pub struct Iter<'a, T> {
```
jeehoonkang, line 1093: I wonder if it changes the semantics.

stjepang: It doesn't change semantics. Since Rust 2018, the compiler automatically infers `T: 'a` even if you don't write it explicitly.

taiki-e: @jeehoonkang 

This is a feature named infer-outlives-requirements ([RFC2093](https://github.com/rust-lang/rfcs/blob/master/text/2093-infer-outlives.md
)) stabilized in Rust 1.31, and not changing the semantics ([This is also available in Rust 2015](https://play.rust-lang.org/?version=stable&mode=debug&edition=2015&gist=34db2bf0e238890eb296be1d4c96d73e)):

> Remove the need for explicit T: 'x annotations on structs. We will infer their presence based on the fields of the struct. In short, if the struct contains a reference, directly or indirectly, to T with lifetime 'x, then we will infer that T: 'x is a requirement:
> 
> ```rust
> struct Foo<'x, T> {
>   // inferred: `T: 'x`
>   field: &'x T
> }  
>```

For example:
* In a field `&'a T`, `T` must have a longer lifetime than the reference `&'a`, so `T: 'a` can be elided in struct definitions.
* In a field `&'a Type<'b, T>`, `Type` and both a type `T` and a lifetime `'b` included in `Type` must have a longer lifetime than the reference `&'a`, so `'b: 'a` and `T: 'a` can be elided in struct definitions.

Then, `explicit_outlives_requirements` included in `rust_2018_idioms` warn if lifetime bounds can be inferred based on this rule (i.e, if elided it does not change the semantics).

For example: [no warning is issued if lifetimes and generics are not associated.](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=00af57c2dc97f01fefbc887ce47404ce)

Of course, it can compile even if manually remove those bounds and change the semantics, but the elisions of lifetime bounds I did this time are all based on `explicit_outlives_requirements`'s warning.

jeehoonkang:
Thanks!

bors r+

PR #515, opened 2020-05-24
Deny warnings for doc tests

taiki-e:
There are some unused imports/`mut`s.

Refs: rust-lang/futures-rs#1672, tokio-rs/tokio#1539

### DIFF
--- a/crossbeam-channel/src/lib.rs
+++ b/crossbeam-channel/src/lib.rs
@@ -122,8 +122,6 @@
 //! It's also possible to share senders and receivers by reference:
 //!
 //! ```
-//! # fn main() {
-//! use std::thread;
 //! use crossbeam_channel::bounded;
 //! use crossbeam_utils::thread::scope;
 //!
@@ -140,7 +138,6 @@
 //!     s.send(1).unwrap();
 //!     r.recv().unwrap();
 //! }).unwrap();
-//! # }
 //! ```
 //!
 //! # Disconnection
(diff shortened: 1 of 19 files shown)


crossbeam-channel/src/lib.rs
```diff
@@ -338,6 +331,13 @@
 //! [`Sender`]: struct.Sender.html
 //! [`Receiver`]: struct.Receiver.html
 
+#![doc(test(
+    no_crate_inject,
+    attr(
+        deny(warnings, rust_2018_idioms),
```
jeehoonkang, line 337: `deny(warnings)` may not be forward compatible, as future versions of Rust may produce more warnings than before.  How about just leaving warnings as warnings and turning them into errors in CI?

taiki-e: > `deny(warnings)` may not be forward compatible, as future versions of Rust may produce more warnings than before.

This `deny` only applies to the examples in the documentation and not to the published code, so there is no forward compatibility issue. If the compiler adds a new warning, it just breaks the CI.

> How about just leaving warnings as warnings and turning them into errors in CI?

We've looked into this before, but unfortunately couldn't find a way to apply deny to CI only: https://github.com/rust-lang/futures-rs/pull/1672#issuecomment-502018036

jeehoonkang:
I'm not 100% sure about the forward compatibility story, but I'd just like to follow the precedent of https://github.com/rust-lang/futures-rs. Thanks!

bors r+

PR #516, opened 2020-05-24
Update scopeguard to 1, rand to 0.7

taiki-e:
Updates major version of 2 private (dev-)dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,7 @@ path = "./crossbeam-utils"
 default-features = false
 
 [dev-dependencies]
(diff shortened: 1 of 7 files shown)


taiki-e:
FYI: The other older dependencies are `flume` and `futures` in `benchmarks` (`flume` will be addressed by #481).

https://github.com/crossbeam-rs/crossbeam/blob/9d57966c4abb0bc7ec894fc619579dc0109616f4/crossbeam-channel/benchmarks/Cargo.toml#L14-L15

PR #518, opened 2020-05-25
Enable tests on Windows on CI

taiki-e:
crossbeam has some `cfg(windows)` code, but we don't test on CI. It would be nice to be able to test these with CI.
https://github.com/crossbeam-rs/crossbeam/blob/48790384d49192b928476400b52d81cd008b8869/crossbeam-utils/src/thread.rs#L537
https://github.com/crossbeam-rs/crossbeam/blob/48790384d49192b928476400b52d81cd008b8869/crossbeam-utils/tests/thread.rs#L196

Travis's Windows support was experimental and very slow, but GitHub Actions' Windows support relatively fast. Hopefully we can add this to our CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,11 +7,18 @@ on:
       - staging
       - trying
 
+env:
+  RUSTFLAGS: -Dwarnings
+  RUST_BACKTRACE: 1
+
+defaults:
+  run:
+    shell: bash
+
 jobs:
   # Test crates on their minimum Rust versions and nightly Rust.
   test:
     name: test
-    runs-on: ubuntu-latest
     env:
       RUST_VERSION: ${{ matrix.rust }}
     strategy:
@@ -27,6 +34,10 @@ jobs:
         rust:
           - 1.36.0
           - nightly
+        os:
+         - ubuntu
+         - windows
+    runs-on: ${{ matrix.os }}-latest
     steps:
       - uses: actions/checkout@master
       - name: Install Rust
@@ -60,3 +71,32 @@ jobs:
         run: rustup update stable && rustup default stable
       - name: rustfmt
         run: ./ci/rustfmt.sh
+
+  # These jobs don't actually test anything, but they're used to tell bors the
+  # build completed, as there is no practical way to detect when a workflow is
+  # successful listening to webhooks only.
+  #
+  # ALL THE PREVIOUS JOBS NEEDS TO BE ADDED TO THE `needs` SECTION OF THIS JOB!
+
+  ci-success:
+    name: ci
+    if: github.event_name == 'push' && success()
+    needs:
+      - test
+      - dependencies
+      - rustfmt
+    runs-on: ubuntu-latest
+    steps:
+      - name: Mark the job as a success
+        run: exit 0
+  ci-failure:
+    name: ci
+    if: github.event_name == 'push' && !success()
+    needs:
+      - test
+      - dependencies
+      - rustfmt
+    runs-on: ubuntu-latest
+    steps:
+      - name: Mark the job as a failure
+        run: exit 1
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -60,3 +75,32 @@ jobs:
         run: rustup update stable && rustup default stable
       - name: rustfmt
         run: ./ci/rustfmt.sh
+
+  # These jobs don't actually test anything, but they're used to tell bors the
+  # build completed, as there is no practical way to detect when a workflow is
+  # successful listening to webhooks only.
+  #
+  # ALL THE PREVIOUS JOBS NEEDS TO BE ADDED TO THE `needs` SECTION OF THIS JOB!
```
taiki-e, line 83: This is a realistic way for bors to receive build status from Github Actions, which do a lot of jobs (It's used in [`rust-lang/rust` etc.](https://github.com/rust-lang/rust/blob/62da38d00d8d09dbaaa092c7b5e7ea343fdc2126/src/ci/github-actions/ci.yml#L631-L633), [previously we had to manually manage all job names.](https://github.com/crossbeam-rs/crossbeam/pull/518/files#diff-2f8c9db9253b64f9a332c09f492cd54e))

taiki-e:
(I originally tried to add both macOS and Windows to the CI, but I had a lot of locally non-reproducible errors with macOS, so I opened a windows-only PR for now.)

taiki-e:
Now, this should be ready for review.

ci/crossbeam-epoch.sh
```diff
@@ -13,13 +13,15 @@ if [[ "$RUST_VERSION" == "nightly" ]]; then
...
     cargo test --features nightly
 
-    ASAN_OPTIONS="detect_odr_violation=0 detect_leaks=0" \
-    RUSTFLAGS="-Z sanitizer=address" \
-    cargo run \
-        --release \
-        --target x86_64-unknown-linux-gnu \
-        --features sanitize,nightly \
-        --example sanitize
+    if [[ "$OS" != "windows" ]]; then
```
jeehoonkang, line 16: how about saying "if the os is linux"?  Because we're targetting on "x86_64-unknown-linux-gnu" in this code.

crossbeam-channel/examples/stopwatch.rs
```diff
@@ -52,3 +53,8 @@ fn main() {
         }
     }
 }
+
+#[cfg(windows)]
```
jeehoonkang, line 57: How about moving this to the top?  In the spirit of early return.. :)

.github/workflows/ci.yml
```diff
@@ -27,6 +35,13 @@ jobs:
         rust:
           - 1.36.0
           - nightly
+        os:
+         - ubuntu
+         - windows
+        exclude:
```
jeehoonkang, line 41: May I ask the meaning of this exclusion?

ci/crossbeam-epoch.sh
taiki-e: Sounds good. 

Also, I changed to get `$OSTYPE` directly from bash instead of from the environment variable set in CI settings.

crossbeam-channel/examples/stopwatch.rs
taiki-e: Sounds good to me :)

.github/workflows/ci.yml
taiki-e: I removed that exclusion.
Originally, I excluded it because I had a lot of errors when I first ran it (I got the same error from both nightly and 1.36 respectively and it was a hassle), but, in reality, we probably don't need to exclude it.

taiki-e:
Hm, bors-ng doesn't seem to handle the change from fork to bors.toml correctly? 

Related: https://github.com/bors-ng/bors-ng/issues/806

bors r-

taiki-e:
So, I'll open the PR from this repo's branch with the same commit.

taiki-e:
Closing in favor of #519

PR #524, opened 2020-05-29
Remove obsolete FnMut conversion

e00E:
Currently we convert the FnOnce closure to a FnMut closure. This is not
needed and there is no comment explaining why this is done. I suspect
this might have been required originally because the api or compiler was
different.


e00E:
I found the original PR that added these https://github.com/crossbeam-rs/crossbeam-utils/pull/47 .
Even there it does not look like it's needed and no explanation is given.

PR #525, opened 2020-05-31
add support of a constant initialization routine

stlankes:
helps the compiler to pre-compute some quantities

### DIFF
--- a/crossbeam-utils/src/cache_padded.rs
+++ b/crossbeam-utils/src/cache_padded.rs
@@ -90,21 +90,6 @@ impl<T> CachePadded<T> {
     ///
     /// let padded_value = CachePadded::new(1);
     /// ```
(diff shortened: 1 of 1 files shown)


crossbeam-utils/src/cache_padded.rs
```diff
@@ -90,10 +90,25 @@ impl<T> CachePadded<T> {
...
+    /// Pads and aligns a value to the length of a cache line.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use crossbeam_utils::CachePadded;
+    ///
+    /// let padded_value = CachePadded::new(1);
+    /// ```
+    #[cfg(feature = "nightly")]
```
taiki-e, line 107: This actually seems to work without the `nightly` feature.

PR #528, opened 2020-06-05
Fix clippy warnings

taiki-e:


taiki-e:
bors r+

(this isn't approved by other reviewers, but this is needed to fix broken CI and I believe the changes themselves are non-controversial.)

PR #535, opened 2020-06-26
Add initial docs to crossbeam-skiplist

caelunshun:
As per https://github.com/crossbeam-rs/crossbeam/issues/503#issuecomment-631679442, I'm writing the docs for `crossbeam-skiplist`. 

Checklist:
* [x] Crate-level docs
* [x] Docs for `SkipMap`
* [ ] Docs for `SkipSet`

I also have some thoughts about the API:
* `Entry`'s name is misleading; it's not the same as the standard library's map entry API. A name like `Ref` might be more suitable, to align with `RefCell`. 
* `set::Entry` could implement `Deref<Target=T>` for convenience. 
* I see the base `SkipList` struct is exposed, though I'm not sure if it belongs in the public API.

### DIFF
--- a/crossbeam-skiplist/src/base.rs
+++ b/crossbeam-skiplist/src/base.rs
@@ -356,7 +356,7 @@ impl<K, V> SkipList<K, V> {
     pub fn len(&self) -> usize {
         let len = self.hot_data.len.load(Ordering::Relaxed);
 
-        // Due to the relaxed  memory ordering, the length counter may sometimes
+        // Due to the relaxed memory ordering, the length counter may sometimes
         // underflow and produce a very large value. We treat such values as 0.
         if len > isize::max_value() as usize {
             0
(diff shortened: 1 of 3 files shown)


stjepang:
This is amazing work <3 

`Entry` is a lot like the new linked list `Cursor`: https://doc.rust-lang.org/nightly/alloc/collections/linked_list/struct.Cursor.html

The idea behind `SkipList` is that it is a low-level API for people who want more control than what `SkipSet` and `SkipMap` offer. I believe @Amanieu is using it.

Amanieu:
`SkipList` is exposed because it is usable in `no_std` code. It does not rely on the thread-local `Collector` (since `no_std` doesn't have TLS).

taiki-e:
Hi @caelunshun! Any updates on this? (It seems that there is still one checkbox left.)

If you don't have the bandwidth to work on the remaining things, I'd like to leave them to follow-up PR (by me or someone) and merge this to avoid conflicts.

BTW, this looks great. Thanks for working on this!

taiki-e (approved):
Thanks @caelunshun for working on this!

I added the remaining things to the checklist of the tracking issue (crossbeam-rs/crossbeam#635).

(It's been a month since the previous comment, so I'll merge it.)
bors r+

Issue #523, opened 2020-05-28
SegQueue.len() cannot be called concurrently while changing


frostyplanet:
I'm using crossbeam-0.7.3,  wrapping crossbeam channel with future, I use two SegQueues to pass waker of the other end. while running with mutliple thread tokio runtime, SegQueue.len() panic while concurrent pop and push happend. It will reproduce every time.

(Have not try crossbeam-queue-0.2.2 because it's not released)

```
mpsc::bounded::tests::test_future_both_8_thread_multi_4tx ... thread 'tokio-runtime-worker' panicked at 'attempt to subtract with overflow', /home/plan/.cargo/registry/src/github.com-9030ffeae4fbd61d/crossbeam-queue-0.2.1/src/seg_queue.rs:423:21
stack backtrace:
   0:     0x556a57d1b1d4 - backtrace::backtrace::libunwind::trace::h90669f559fb267f0
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1:     0x556a57d1b1d4 - backtrace::backtrace::trace_unsynchronized::hffde4e353d8f2f9a
                               at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2:     0x556a57d1b1d4 - std::sys_common::backtrace::_print_fmt::heaf44068b7eaaa6a
                               at src/libstd/sys_common/backtrace.rs:77
   3:     0x556a57d1b1d4 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h88671019cf081de2
                               at src/libstd/sys_common/backtrace.rs:59
   4:     0x556a57d3c5bc - core::fmt::write::h4e6a29ee6319c9fd
                               at src/libcore/fmt/mod.rs:1052
   5:     0x556a57d17d07 - std::io::Write::write_fmt::hf06b1c86d898d7d6
                               at src/libstd/io/mod.rs:1426
   6:     0x556a57d1d515 - std::sys_common::backtrace::_print::h404ff5f2b50cae09
                               at src/libstd/sys_common/backtrace.rs:62
   7:     0x556a57d1d515 - std::sys_common::backtrace::print::hcc4377f1f882322e
                               at src/libstd/sys_common/backtrace.rs:49
[...]

PR #537, opened 2020-07-16
Change crossbeam-channel's license to MIT OR Apache-2.0

### LINKED ISSUE #536: AND license in crossbeam_channel
Hi, I see that in #419 issues with AND usage of licenses were at least partly addressed and the master branch now has a license expressed as `MIT AND BSD-2-Clause OR Apache-2.0 AND BSD-2-Clause`.

I'm opening this ticket in the context of adding a dependency to crossbeam channel in [webrender](https://github.com/servo/webrender/), for use Firefox. We have to somewhat meticulous about licenses. I am not personally well versed in the topic, so I will mostly quote the feedback I got at the license approval review:

"MIT AND BSD-2-Clause" is somewhat reasonable, but that's only because there's vanishingly little difference between the MIT and BSD-2 licenses: MIT implies some freedoms BSD-2 is explicit about, and that's just about it. Using OR here would not change the situation here significantly.

"Apache-2.0 AND BSD-2-Clause" is a problematic construct: Apache 2.0 provides explicit patent liability protection and the BSD-2-clause explicitly says you get no such protection.

Quoting from the conclusion of the license approval review: "If we're going to pull this in while staying compliant with the licenses involved we should ask upstream to either pick one or "OR" two, and we definitely don't want to permit "AND" as an acceptable predicate in our licensing logic because that way lies madness."

What particular guarantees do you mean to express through the current license? would it be possible to simplify it into `MIT OR Apache-2.0 OR BSD-2-Clause`?

nical:
Fixes #536 

Was there breaking changes since the last published versions? it looks like it was 5 months ago so it sounds likely. If so I suspect you'll want to wait for other potential changes before making a `0.5` release, would it be possible to create a 0.4 branch and change publish the license update there on top of `0.4.2`'s revision?

That would allow me to use crossbeam-channel in WebRender.


nical:
The CI doesn't provide much information about the failing tests. They had been pending for a long long time so I suppose the backlog was just canceled/timed-out at some point.

stjepang:
Published a new patch version of crossbeam-channel that includes this PR.

nical:
Thanks a lot !

taiki-e:
Hm, the `master` branch was supposed to contain breaking changes... But does the compile seem to succeed?

Either way, I think we need to create the `v0.7` branch and backport this and publish a new version of crossbeam-channel. (0.4.3 has [UB](https://github.com/crossbeam-rs/crossbeam/issues/539) (there is no UB in versions 0.4.2 or less), and [adding default-feature](https://github.com/crossbeam-rs/crossbeam/commit/208bda7b796141b061749ce05dbecedf7908965a) is technically a breaking change)
(I did the same thing in crossbeam-queue before, but I pushed only [the tag](https://github.com/crossbeam-rs/crossbeam/releases/tag/crossbeam-queue-0.2.3) without pushing the branch; https://github.com/crossbeam-rs/crossbeam/issues/523#issuecomment-642347864)

taiki-e:
Published crossbeam-channel v0.4.4, and yanked crossbeam-channel v0.4.3.

PR #546, opened 2020-08-25
Make AtomicCell::is_lock_free const function at 1.46+

### LINKED ISSUE #531: Mark `AtomicCell::is_lock_free()` as const fn?
It would be convenient if I could check if AtomicCells are lock-free at compile time. by marking `AtomicCell<T>::is_lock_free()` as a const fn.

For context, I'm working on an audio program in C++, due to C++ libraries including Qt. In audio programs, taking locks on the audio thread is regarded as unacceptable because it can block and cause the audio to stutter. As a result, I'd like to statically assert that a given atomic type written by the audio thread is lock-free. (I'm mostly looking at Rust for inspiration and comparison, to see whether parts of my code are nicer in C++ or Rust.)

https://github.com/crossbeam-rs/crossbeam/blob/81ab18efc7376f3277a6359fdc20d00b4bcf911c/crossbeam-utils/src/atomic/atomic_cell.rs#L108-L110

calls `atomic_is_lock_free<T>`:

https://github.com/crossbeam-rs/crossbeam/blob/81ab18efc7376f3277a6359fdc20d00b4bcf911c/crossbeam-utils/src/atomic/atomic_cell.rs#L809-L812

expands to the `atomic!` macro:

https://github.com/crossbeam-rs/crossbeam/blob/81ab18efc7376f3277a6359fdc20d00b4bcf911c/crossbeam-utils/src/atomic/atomic_cell.rs#L777-L807

Marking `AtomicCell<T>::is_lock_free()` as `const` doesn't seem to cause issues, but depends on `atomic_is_lock_free`. Trying to mark `atomic_is_lock_free` as `const` fails on https://github.com/rust-lang/rust/issues/49146 "Allow `if` and `match` in constants", and https://github.com/rust-lang/rust/issues/52000 "Allow `loop` in constant evaluation".

From a cursory examination, `atomic!` doesn't truly need a `loop`, but depends on `if`. Note that I don't fully understand `atomic!` nor all call sites.

Are there any plans to mark this as a const fn, to allow users to statically assert it's lock-free? I don't know if `atomic!` needs to be split into two macros or not, with duplication and possible future desynchronization.

taiki-e:
Closes #531

### DIFF
--- a/crossbeam-utils/src/atomic/atomic_cell.rs
+++ b/crossbeam-utils/src/atomic/atomic_cell.rs
@@ -12,6 +12,7 @@ use core::sync::atomic::{self, AtomicBool, Ordering};
 use std::panic::{RefUnwindSafe, UnwindSafe};
 
 use super::seq_lock::SeqLock;
+use const_fn::const_fn;
 
 /// A thread-safe mutable memory location.
 ///
@@ -105,7 +106,8 @@ impl<T> AtomicCell<T> {
     /// // operations on them will have to use global locks for synchronization.
     /// assert_eq!(AtomicCell::<[u8; 1000]>::is_lock_free(), false);
     /// ```
-    pub fn is_lock_free() -> bool {
+    #[const_fn("1.46")]
+    pub const fn is_lock_free() -> bool {
         atomic_is_lock_free::<T>()
     }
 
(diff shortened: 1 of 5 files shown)


crossbeam-utils/src/atomic/atomic_cell.rs
```diff
@@ -105,7 +106,8 @@ impl<T> AtomicCell<T> {
     /// // operations on them will have to use global locks for synchronization.
     /// assert_eq!(AtomicCell::<[u8; 1000]>::is_lock_free(), false);
     /// ```
-    pub fn is_lock_free() -> bool {
+    #[const_fn("1.46")]
+    pub const fn is_lock_free() -> bool {
```
taiki-e, line 110: `#[const_fn]` is proc-macro, but it is very simple and [has no proc-macro related dependencies](https://github.com/taiki-e/const_fn/issues/16), so it has little impact on compile time.

stjepang (approved):
Thank you!

PR #547, opened 2020-08-30
impl From<T> for AtomicCell<T>

taiki-e:
std atomic types also have similar `From` impls: e.g., [`AtomicBool`](https://doc.rust-lang.org/nightly/std/sync/atomic/struct.AtomicBool.html#impl-From%3Cbool%3E), [`AtomicUsize`](https://doc.rust-lang.org/nightly/std/sync/atomic/struct.AtomicUsize.html#impl-From%3Cusize%3E)


stjepang (approved):

taiki-e:
bors r=stjepang

PR #548, opened 2020-08-30
Remove our own error types from crossbeam-queue

taiki-e:
Currently, `crossbeam-queue` has two error types: `PopError`, `PushError`.
* `PopError` is a unit struct with no field, and is returned by `ArrayQueue::pop` and `SegQueue::pop` if the queue is empty.
* `PushError` is a tuple struct with a public field that stores the element that user tried to push, and is returned by `ArrayQueue::push` if the queue is full.

I guess the current API was influenced by channels, but there is only one kind(variant) of error in each, and unlike channels, multiple kinds(variants) of errors do not occur.
And, neither of these provides additional information about the error (fields are all public, so it doesn't seem like they are going to provide them in the future).

So, APIs that return `Option` (on `pop` methods) and `Result` (on `ArrayQueue::push`) without our own errors are simple, and I think it is preferable. 
(Especially for the `pop` methods, it will be more consistent, because all other public `pop` methods in crossbeam return `Option`.)

(Of course, this is a breaking change, but the master already has some breaking changes.)


stjepang (commented):
This seems like an okay change to me. :) I only chose to return `Result` from `pop()` for symmetry between `push()` and `pop()`.

PR #552, opened 2020-09-01
Make Local less than 2048 bytes

### LINKED ISSUE #551: The Local structure is 2104 bytes long and jemalloc rounds this to 4096 bytes.
The Local structure here https://searchfox.org/mozilla-central/source/third_party/rust/crossbeam-epoch/src/internal.rs#287 is 2104 bytes long. When compiled into Firefox and using the jamalloc memory allocator this is rounded up to 4096 bytes (a very common practice).  This wastes just under 12KB for a Firefox process that loads a simple website. It'd be nice if it could allocate 2048 bytes or less saving this memory.

Thanks.

taiki-e:
Set `MAX_OBJECTS` to 62, and make the size of `Local` to 2040 bytes.

Closes #551

### DIFF
--- a/crossbeam-epoch/src/internal.rs
+++ b/crossbeam-epoch/src/internal.rs
@@ -55,7 +55,7 @@ use crate::sync::queue::Queue;
 
 /// Maximum number of objects a bag can contain.
 #[cfg(not(feature = "sanitize"))]
-const MAX_OBJECTS: usize = 64;
+const MAX_OBJECTS: usize = 62;
 #[cfg(feature = "sanitize")]
 const MAX_OBJECTS: usize = 4;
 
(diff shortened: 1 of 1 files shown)


crossbeam-epoch/src/internal.rs
```diff
@@ -55,7 +55,7 @@ use crate::sync::queue::Queue;
 
 /// Maximum number of objects a bag can contain.
 #[cfg(not(feature = "sanitize"))]
-const MAX_OBJECTS: usize = 64;
+const MAX_OBJECTS: usize = 62;
```
taiki-e, line 58: fwiw, if make `Local` to be less than or equal to 1024 bytes, `MAX_OBJECTS` needs to be 30 (1016 bytes)

taiki-e: If I didn't misunderstand @stjepang's [this comment](https://github.com/crossbeam-rs/crossbeam/issues/551#issuecomment-684515013), It seems fine with less than half of what it is now, so it may be okay with this.

stjepang (approved):
Looks good to me!

taiki-e:
bors r=stjepang

PR #554, opened 2020-09-04
Added "deadline" versions of the various "timeout" entry points in crossbeam channel

### LINKED ISSUE #553: Feature request: `recv_deadline`
This is a feature request for an additional `recv` form, `recv_deadline`, for `Recevier` (and an equivalent version for `Sender`). In some use cases, you have an `Instant` to wait for, instead of a `Duration` to wait for, and there's no reason to convert this into a `Duration` when `recv_timeout` will just be converting it *back* into an `Instant`.

Lucretiel:
Added:

- `Receiver::recv_deadline`
- `Sender::send_deadline`
- `channel::at` (the deadline version of `channel::after`)
- `Select::select_deadline`
- `Select::ready_deadline`

Added associated docs & doctests for these, but in general they share implementations with the existing `timeout` versions, so I didn't add additional unit tests.

Didn't add a deadline version of the select macro for now, since it wasn't clear what the syntax would be or if it's really necessary (since deadline versions of things will be more for low/medium-level abstractions built on top of crossbeam, which would want to minimize abstraction).

Didn't add Parker::park_deadline, since that would have involved a heavier refactor of the `Parker` implementation (probably removing the possibility of a spurious wakeup).

Fixes #553

### DIFF
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -172,7 +172,54 @@ pub fn bounded<T>(cap: usize) -> (Sender<T>, Receiver<T>) {
 /// ```
 pub fn after(duration: Duration) -> Receiver<Instant> {
     Receiver {
-        flavor: ReceiverFlavor::After(Arc::new(flavors::after::Channel::new(duration))),
+        flavor: ReceiverFlavor::After(Arc::new(flavors::after::Channel::new_timeout(duration))),
+    }
+}
+
+/// Creates a receiver that delivers a message after at a certain instant in time
+///
+/// The channel is bounded with capacity of 1 and never gets disconnected. Exactly one message will
+/// be sent into the channel at the moment in time `when`. The message is the instant at which it
+/// is sent, which is the same as `when`.
+///
+/// # Examples
+///
+/// Using an `after` channel for timeouts:
+///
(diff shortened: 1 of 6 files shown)


Lucretiel:
Oh, certainly; my rule of thumb is always to not change that stuff unless I have to. I'll update it tomorrow.

Lucretiel:
I think I caught all of them, but please review closely in case I missed any.

crossbeam-channel/src/channel.rs
```diff
@@ -172,7 +172,55 @@ pub fn bounded<T>(cap: usize) -> (Sender<T>, Receiver<T>) {
 /// ```
 pub fn after(duration: Duration) -> Receiver<Instant> {
     Receiver {
-        flavor: ReceiverFlavor::After(Arc::new(flavors::after::Channel::new(duration))),
+        flavor: ReceiverFlavor::At(Arc::new(flavors::at::Channel::new_timeout(duration))),
+    }
+}
+
+/// Creates a receiver that delivers a message at a certain instant in time
```
stjepang, line 179:
```suggestion
/// Creates a receiver that delivers a message at a certain instant in time.
```

stjepang (approved):

PR #555, opened 2020-09-06
Add AtomicCell::get_mut

taiki-e:
std atomic types also have the same method: e.g., [AtomicBool](https://doc.rust-lang.org/nightly/std/sync/atomic/struct.AtomicBool.html#method.get_mut), [AtomicUsize](https://doc.rust-lang.org/nightly/std/sync/atomic/struct.AtomicUsize.html#method.get_mut)

### DIFF
--- a/crossbeam-utils/src/atomic/atomic_cell.rs
+++ b/crossbeam-utils/src/atomic/atomic_cell.rs
@@ -62,6 +62,25 @@ impl<T> AtomicCell<T> {
         }
     }
 
+    /// Returns a mutable reference to the underlying value.
+    ///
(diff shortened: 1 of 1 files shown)


stjepang:
This is apparently not possible to implement in a sound way, and the rationale is non-trivial:

* https://github.com/crossbeam-rs/crossbeam/issues/315#issuecomment-462369913
* https://github.com/crossbeam-rs/crossbeam/pull/332

taiki-e:
@stjepang Ah, I missed that, thanks for pointing that out.

taiki-e:
...And I was the one who removed it (https://github.com/crossbeam-rs/crossbeam/pull/506) 😅

PR #559, opened 2020-09-06
Use intra-doc links

taiki-e:
Switch to use [intra-doc links](https://github.com/rust-lang/rfcs/pull/1946) in all crates. Previously there was a big bug on cross crate re-exports (https://github.com/rust-lang/rust/issues/65983), but it has been fixed, so we can use this in crossbeam.

This also adds checks of the docs to CI.


stjepang (approved):

taiki-e:
bors r=stjepang

PR #561, opened 2020-09-07
Use "closed" instead of "disconnected"

taiki-e:
Addresses https://github.com/crossbeam-rs/crossbeam/issues/256#issuecomment-444606458

### DIFF
--- a/crossbeam-channel/examples/fibonacci.rs
+++ b/crossbeam-channel/examples/fibonacci.rs
@@ -4,7 +4,7 @@ use std::thread;
 
 use crossbeam_channel::{bounded, Sender};
 
-// Sends the Fibonacci sequence into the channel until it becomes disconnected.
+// Sends the Fibonacci sequence into the channel until it becomes closed.
 fn fibonacci(sender: Sender<u64>) {
     let (mut x, mut y) = (0, 1);
     while sender.send(x).is_ok() {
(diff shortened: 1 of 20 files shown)


stjepang:
My main worry about this change is that it would break too much existing code for very little benefit. To see if that is true, I cloned the Servo repository and grepped it for 'disconnected'. I only found one instance of breakage:

https://github.com/servo/servo/blob/a5a21a59addae0df6d9e050f17d44399db04fec3/components/script/dom/window.rs#L1735

This makes me feel better considering that Servo is a big codebase. Still... I'm a bit worried that other projects might be matching on error types more often. Crossbeam might have reached the point where it's so widespread that stability is preferable over minor API improvements.

What do you think?

taiki-e:
Hmm, given that the current term is not wrong, the cost of changing this will probably be greater than the benefit. (I found nearly 50 "Try(Send|Recv)Error::Disconnected" in tikv.)

What I was thinking was mainly about the impact on methods that may be added in the future, such as close and is_closed. (As time goes by, it becomes more difficult to standardize terms within the ecosystem. But in reality, it may already be impossible...)

stjepang:
Yeah I think that ship has sailed :) We'll probably never agree on a 'standard' channel interface, but that's okay. I don't think it's a super important problem worth solving. We're fine with a bunch similar but slightly different interfaces...

taiki-e:
Ok! Closing in favor of the above opinion that the benefits of doing this are not enough.

PR #563, opened 2020-09-08
Updates to `Parker`

### LINKED ISSUE #482: Please clarify the behavior of Parker
Hi

I have some questions about the [Parker](https://docs.rs/crossbeam/0.7.3/crossbeam/sync/struct.Parker.html) type. The `park()` method explicitly mentions the following:

> A call to park may wake up **spuriously** without consuming the token, and callers should be prepared for this possibility.

Questions:
- If I understand this correctly; when a thread blocks at `p.park()`, it *might* unblock and continue without having to directly call `u.unpark()`?
- Under what circumstances can this *spurious wake up* happen?
- If a certain process requires `park()` to work reliably, what should be used instead? What's the point of `Parker` if it's unreliable?

Thanks for your time.

Lucretiel:
- No longer a possibility for a spurious wake (previously possible if using a timeout), and removed references to spurious wake in the documentation.
- Implementation updates to `park`: now deadline based, and unified the code paths between timeout and no-timeout versions
- **Breaking Change** `park_timeout` and `park_deadline` now report the reason for their return (timeout or `unpark`) in an enum.

If you'd rather not have this breaking change, I'm happy to revert that side of it and just focus on the spurious awakening prevention stuff.

Fixes #482

### DIFF
--- a/crossbeam-utils/src/sync/mod.rs
+++ b/crossbeam-utils/src/sync/mod.rs
@@ -8,6 +8,6 @@ mod parker;
 mod sharded_lock;
 mod wait_group;
 
-pub use self::parker::{Parker, UnparkReason, Unparker};
+pub use self::parker::{Parker, Unparker};
 pub use self::sharded_lock::{ShardedLock, ShardedLockReadGuard, ShardedLockWriteGuard};
 pub use self::wait_group::WaitGroup;
--- a/crossbeam-utils/src/sync/mod.rs
+++ b/crossbeam-utils/src/sync/mod.rs
@@ -8,6 +8,6 @@ mod parker;
 mod sharded_lock;
 mod wait_group;
 
-pub use self::parker::{Parker, Unparker};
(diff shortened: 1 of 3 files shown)


taiki-e (commented):
Thanks for the PR and sorry for the delay reviewing! The implementation looks good, but we recently released a new minor version of `crossbeam-utils`, so I would prefer not to do the breaking changes right now.

It might make sense to consider doing this in a future version (filed #601 to follow-up), but for now, can you revert the changes for `UnparkReason`?

Lucretiel:
Sure thing, I'll get to it this weekend.

Lucretiel:
Done

Lucretiel:
CI failures seem unrelated; they're being caused by an issue with rustup.

taiki-e:
CI failure was fixed in #606. Could you rebase?

Lucretiel:
Done

taiki-e (approved):
Thanks! 

bors r+

PR #567, opened 2020-09-10
Port select2.go test to Rust

byusti:
A port of select2.go to Rust. Select2 tests for excess memory usage by the select statement. The original go test uses memstats.alloc within the "runtime" package  to determine the net number of bytes allocated on the heap. I used a wrapper around alloc::System to achieve a similar behavior for the Rust port.

I also added make_unbounded() to initialize unbounded channels the go way.

#201


stjepang (approved):
Thank you! :)

PR #569, opened 2020-09-20
Better plot.py with Python 3 & remove hard code

qiujiangkun:
I was trying to add my own crate into the benchmark and found out that this script was hard to read and modify. So I converted it into python 3 and removed most of the hardcode.

### DIFF
--- a/crossbeam-channel/benchmarks/plot.py
+++ b/crossbeam-channel/benchmarks/plot.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python3
 
 import sys
 import matplotlib.pyplot as plt
@@ -17,7 +17,7 @@ fig = plt.figure(figsize=(10, 10))
 def plot(subplot, title, prefix, runs):
     runs.reverse()
 
-    ys = [6 * (i+1) for i in xrange(len(runs))]
+    ys = [6 * (i + 1) for i in range(len(runs))]
     ax = fig.add_subplot(subplot)
     ax.set_title(title)
     ax.set_yticks(ys)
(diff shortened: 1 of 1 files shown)


crossbeam-channel/benchmarks/plot.py
```diff
@@ -17,68 +17,44 @@
...
+                scores[name][i] = secs
 
     opts = dict(height=0.7, align='center')
-    ax.barh([y-2 for y in ys], go, color='skyblue', **opts)
-    ax.barh([y-1 for y in ys], crossbeam_channel, color='red', **opts)
-    ax.barh([y+0 for y in ys], chan, color='orange', **opts)
-    ax.barh([y+1 for y in ys], mpsc, color='black', **opts)
-    ax.barh([y+2 for y in ys], futures_channel, color='blue', **opts)
+    for (i, score) in enumerate(scores.values()):
+        ax.barh([y + i - len(scores) // 2 for y in ys], score, **opts)
```
Vtec234, line 40: We're losing the different colors here, perhaps have a fixed `colors` list that you index with `i`?

Vtec234:
Looks good. Could you please post an example plot created with the new version just to confirm it's comparable to the existing ones?

crossbeam-channel/benchmarks/plot.py
qiujiangkun: Matplotlib will allocate different colors automatically

qiujiangkun:
Currently, it needs some tweaks to display better. But it's fine I think
image

Vtec234:
It seems there are a few problems:
- non-existent benchmarks are default initialized to `0` and then shown as empty bars, but they shouldn't be displayed at all
- the X-axis ticks are gone (just `0` on the left)
- some of the horizontal bars are too long, and as you can see the text overlaps with other graphs

With those fixed, I think it'd be good to go.

qiujiangkun:
Sure, I have noticed them and I'll fix them in a few days.











Sincerely,&nbsp;
Jack&nbsp;Quinn

---Original---

qiujiangkun:
Now the new version has come out
image

Vtec234 (approved):
LGTM. I made a small change to have cleaner names like `mpsc` instead of `Rust_mpsc`, and `chan (Go)` instead of `Go_chan`. I will merge in a week or so if noone objects.

Vtec234:
502 on crates.io, merging manually.

PR #576, opened 2020-10-09
Bump master to 1.0 and mark as 'publish = false'

taiki-e:
The current master branch contains breaking changes, but in the past, we've accidentally published a crate from master several times (https://github.com/crossbeam-rs/crossbeam/pull/537#issuecomment-687755200), and I had to yank each time.

This patch prevents releases from being accidentally created from the master branch by marking all crates as 'publish = false'.

Also, this bumps the master branch's version number to 1.0 (that should be the next major version we release). (The actual 1.0 release is blocked on #503.)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ name = "crossbeam"
 # - Update CHANGELOG.md
 # - Update README.md
 # - Create "crossbeam-X.Y.Z" git tag
-version = "0.7.3"
+version = "1.0.0"
 authors = ["The Crossbeam Project Developers"]
 edition = "2018"
 license = "MIT OR Apache-2.0"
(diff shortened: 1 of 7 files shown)


stjepang (approved):
Thank you!

jeehoonkang:
I'm a little bit hesitant to bump to 1.0 now.  One of the reasons is we merged quite a few large PRs recently.  I propose to increase the minor version now, and if it works well for a certain period of time, bump the version to 1.0.

taiki-e:
> I propose to increase the minor version now, and if it works well for a certain period of time, bump the version to 1.0.

This seems to make sense, given that we can't really release 1.0 right away in any case. I'll file a PR for that.

PR #585, opened 2020-10-15
Get rid of the hack for epoch::unprotected

coolreader18:
Noticed that this hack isn't necessary (anymore?); originally had a newtype over `Guard` that I implemented `Sync` for, and then I realized that a `static mut` works as well.


taiki-e (changes requested):
Thanks for the PR!

I would prefer to avoid using `static mut` because it is very difficult to use correctly and deprecation is being considered (https://github.com/rust-lang/rust/issues/53639, https://github.com/rust-lang/rfcs/pull/2937). (Even if the use of `static mut` is sound and the alternatives are more redundancy.)

In this case, it seems newtype is a reasonable alternative.

taiki-e (approved):
LGTM, thanks!

bors r+

PR #594, opened 2020-11-07
Fix non_autolinks warnings

taiki-e:


taiki-e:
bors r-

https://github.com/crossbeam-rs/crossbeam/runs/1368610393?check_suite_focus=true

PR #596, opened 2020-11-12
Fix a bug in seq_lock_wide (closes #595)

### LINKED ISSUE #595: crossbeam-utils atomic::seq_lock::tests::test_abort unit test fails on 32-bit architectures
From the test output:

```
test atomic::seq_lock::tests::test_abort ... FAILED
failures:
---- atomic::seq_lock::tests::test_abort stdout ----
thread 'atomic::seq_lock::tests::test_abort' panicked at 'assertion failed: `(left == right)`
  left: `(0, 0)`,
 right: `(0, 2)`: aborted write does not update the stamp', src/atomic/seq_lock_wide.rs:151:9
```

The same failure occurs on both i686 and armv7hl, on fedora rawhide. Running the tests on 64-bit architectures is fine.

Can I ignore this test failure for now or is it indicative of a real problem?

jeehoonkang:
It's quite an embarrassing bug by me...

### DIFF
--- a/crossbeam-utils/src/atomic/seq_lock_wide.rs
+++ b/crossbeam-utils/src/atomic/seq_lock_wide.rs
@@ -1,3 +1,4 @@
+use core::mem;
 use core::sync::atomic::{self, AtomicUsize, Ordering};
 
 use crate::Backoff;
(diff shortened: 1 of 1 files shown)


jeehoonkang:
It's kinda "hotfix", so I'm merging it myself.

bors r+


@decathorpe Thank you for reporting this.  Do you need this patch to be released ASAP?

decathorpe:
Awesome, thanks for fixing it so quickly!
I can include this patch where I need it, so there's no need to hurry out a point release :)

Issue #598, opened 2020-11-13
We need to test i686 linux-gnu on CI by using cross


jeehoonkang:
Ideally, we need to test i686 linux-gnu on CI by using cross. (We are not currently testing platforms other than x86_64 linux&windows.)
However, some of the channel tests (deadline/timeout-related) are pretty fragile. (After all, that's why #578 and #591 can't be merged yet. That's also why [macOS hasn't been tested in CI](https://github.com/crossbeam-rs/crossbeam/pull/518#issuecomment-633342606).)

_Originally posted by @taiki-e in https://github.com/crossbeam-rs/crossbeam/issues/597#issuecomment-726334759_

PR #599, opened 2020-11-13
Relaxed size constrait of Local (closes #597)

### LINKED ISSUE #597: crossbeam-epoch internal::local_size unit test fails on 32-bit architectures
```
failures:
---- internal::local_size stdout ----
thread 'internal::local_size' panicked at 'assertion failed: `(left == right)`
  left: `2040`,
 right: `1020`', src/internal.rs:379:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
failures:
    internal::local_size
```

The same error occurs on both i686 and armv7hl, on fedora rawhide, with stable Rust 1.47.0. The test is fine on all 64-bit architectures that are available to me (x86_64, aarch64, ppc64le, and s390x).

Looking at the source code for the failing test
https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-epoch/src/internal.rs#L375

it seems the intent is to check if `Local` is equal or smaller than 2040 bytes (at least that's what the comment says), but the code in the test actually checks for equality: `assert_eq!(2040, core::mem::size_of::<Local>());`, not a "less than or equal" test.

Should this be changed to `assert!(core::mem::size_of::<Local>() <= 2040);` or something like that? I think this value is half the expected value because pointers are half as big on 32-bit architectures ... Or would it make sense to actually compare using a `usize` value (the architecture dependent pointer size)?

PS: Since this looks like an issue in the test and not a code issue, I'll ignore the failure for now, so fixing this is not urgent :)

jeehoonkang:
@decathorpe Thank you for reporting this.  It's also kinda hotfix.  As @taiki-e said, we really need CI for i686 and other 32-bit architectures... (#598).

r? @taiki-e

### DIFF
--- a/crossbeam-epoch/src/internal.rs
+++ b/crossbeam-epoch/src/internal.rs
@@ -376,7 +376,7 @@ pub struct Local {
 // https://github.com/crossbeam-rs/crossbeam/issues/551
 #[test]
 fn local_size() {
-    assert_eq!(2040, core::mem::size_of::<Local>());
+    assert!(core::mem::size_of::<Local>() <= 2040, "An allocation of `Local` should be <= 2048 bytes.");
 }
 
 impl Local {
--- a/crossbeam-epoch/src/internal.rs
+++ b/crossbeam-epoch/src/internal.rs
@@ -376,7 +376,7 @@ pub struct Local {
 // https://github.com/crossbeam-rs/crossbeam/issues/551
 #[test]
 fn local_size() {
-    assert!(core::mem::size_of::<Local>() <= 2040, "An allocation of `Local` should be <= 2048 bytes.");
+    assert!(core::mem::size_of::<Local>() <= 2048, "An allocation of `Local` should be <= 2048 bytes.");
 }
 
 impl Local {


crossbeam-epoch/src/internal.rs
```diff
@@ -376,7 +376,7 @@ pub struct Local {
 // https://github.com/crossbeam-rs/crossbeam/issues/551
 #[test]
 fn local_size() {
-    assert_eq!(2040, core::mem::size_of::<Local>());
+    assert!(core::mem::size_of::<Local>() <= 2040, "An allocation of `Local` should be <= 2048 bytes.");
```
Vtec234, line 379: The message disagrees on the size - `2040` or `2048`?

jeehoonkang: I "guess" the reason is memory allocator would have 8 bytes of overhead.  But it's just a guess, and I'd like to ping @stjepang ...

In any case, I'll improve the assert description.  It's confusing as you said.

taiki-e: 2040 is the actual size of the current `Local`, so when comparing using `<=`, I think `<= 2048` is fine.

decathorpe:
> @decathorpe Thank you for reporting this. It's also kinda hotfix. As @taiki-e said, we really need CI for i686 and other 32-bit architectures... (#598).

Looks good to me (since this is basically the "fix" I suggested) :) Thanks!

crossbeam-epoch/src/internal.rs
taiki-e: I remembered that it was actually me who added this test (#552). IIRC, I just reduced the number of objects and adjusted it to be less than 2048 bytes, so that 8 bytes has no particular meaning.

jeehoonkang: It seems it's quite safe to make its size 2048: https://www.facebook.com/notes/facebook-engineering/scalable-memory-allocation-using-jemalloc/480222803919

jeehoonkang:
@taiki-e would you please review again?  I'd like it to be merged before #603

taiki-e (approved):
bors r+

taiki-e:
bors retry

> warning: spurious network error

PR #600, opened 2020-11-17
Make AtomicCell::is_lock_free always const fn

taiki-e:
`if` expression can only be used in const fn since Rust 1.46, so const `is_lock_free` was require Rust 1.46 (context: #531, #546).
However, when we replace uses of `if` expression with `|` and `&` operators, it seems `is_lock_free` can be used as const fn even in Rust 1.36.

r? @jeehoonkang or @Vtec234

### DIFF
--- a/crossbeam-utils/Cargo.toml
+++ b/crossbeam-utils/Cargo.toml
@@ -31,7 +31,6 @@ nightly = []
 
 [dependencies]
 cfg-if = "1"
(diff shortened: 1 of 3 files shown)


Vtec234 (approved):
I didn't know `&` was stabilised before `&&`! In any case, this looks like it should work.

taiki-e:
bors r=Vtec234

PR #604, opened 2020-11-22
Do not fail to compile if autocfg unable to determine rustc version

taiki-e:
`autocfg::new()` panics if unable to determine rustc version. However, that should not be a fatal error for the library.


taiki-e:
bors r+

I believe this is a non-controversial change.

PR #605, opened 2020-11-22
Change crossbeam-queue's license to MIT OR Apache-2.0

taiki-e:
See #536 and #537 for more. `crossbeam-queue` has the same issue.


decathorpe:
Could you please update the README file for crossbeam-queue as well? It still states that the array_queue.rs file is under the "BSD-2-Clause and Apache-2.0" license (same goes for the header in the source file itself). And since this was cleaned up to be all under the same "MIT OR Apache-2.0" license as the rest, LICENSE-THIRD-PARTY file can probably be removed?

taiki-e:
@decathorpe Done in #607

PR #607, opened 2020-11-29
Update third party licenses

taiki-e:
See #536 for details.


taiki-e:
As discussed in #536:
- change the license to `MIT OR Apache-2.0`.
- keep the link to the original source.

Strictly, we need to get the agreement of all contributors to these files, but @stjepang who wrote most of these codes, has approved this change in #537, and almost all contributors to these files we know are cool with the Rust license, so I think we should do it.

bors r+

PR #611, opened 2020-12-09
Make const_fn crate optional dependency

taiki-e:
This dependency is only actually used for the nightly feature.


taiki-e:
bors r-

https://github.com/crossbeam-rs/crossbeam/runs/1523970216?check_suite_focus=true

PR #591, opened 2020-11-03
Run sanitizers on CI

taiki-e:
This adds AddressSanitizer/MemorySanitizer/ThreadSanitizer tests to CI.

See #589 (tsan: data race in deque), #644 (tsan: data race in atomic), and #614 (asan: memory leak in skiplist) for issues reported by sanitizers.

Closes #154
Closes #589
Closes #644 

Refs: 
- https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/sanitizer.html
- https://github.com/google/sanitizers/wiki

PR #578, opened 2020-10-09
Run Miri on CI

taiki-e:

PR #613, opened 2020-12-09
Test i686-unknown-linux-gnu on CI

### LINKED ISSUE #598: We need to test i686 linux-gnu on CI by using cross

taiki-e:
Closes #598

I'm surprised that it looks like we don't have to ignore channel tests.

r? @jeehoonkang


.github/workflows/ci.yml
```diff
@@ -49,6 +49,30 @@ jobs:
...
+        crates:
+          - crossbeam
+          - crossbeam-channel
+          - crossbeam-deque
+          - crossbeam-epoch
+          - crossbeam-queue
+          - crossbeam-skiplist
+          - crossbeam-utils
+        target:
+          - i686-unknown-linux-gnu
```
taiki-e, line 65: I'm planning to submit a patch for testing on aarch64-unknown-linux-gnu once this patch is merged. Because aarch64-unknown-linux-gnu is a target [recently promoted to tier-1](https://github.com/rust-lang/rfcs/pull/2959), and aarch (ARM architecture) has a memory model different than x86.

.github/workflows/ci.yml
```diff
@@ -49,6 +49,30 @@ jobs:
       - name: Test
         run: ./ci/${{ matrix.crates }}.sh
 
+  # Test crates on non x86_64 architectures.
+  cross:
+    strategy:
+      matrix:
+        crates:
+          - crossbeam
```
jeehoonkang, line 57: I wonder if we can maintain only one set of crates for both x86-64 and the cross-target platforms.  Maybe later we will add more crates, and I'd wish we should change the CI script in only one place...

.github/workflows/ci.yml
jeehoonkang: Personally I'm fine with both splitting and merging the changes.

jeehoonkang (changes requested):

.github/workflows/ci.yml
taiki-e: Agreed. A single 'crates' set is certainly preferable. (It makes a matrix a little more complicated, but it should be possible.)

.github/workflows/ci.yml
taiki-e: I found that it works without ignoring channel tests, so I'll merge it into this PR.

.github/workflows/ci.yml
taiki-e: > It makes a matrix a little more complicated, but it should be possible.

Hmm, the way I thought would work didn't really work. I'll look for other ways.

.github/workflows/ci.yml
taiki-e: Done in 4ba3e3838fc74b08d24fdd30a4bf0473e15283bc

.github/workflows/ci.yml
taiki-e: I addressed this by testing all crates at once in cross tests. (07a1366. The way of adjusting the matrix did not work.)

This should make non-x86-64 platform test time a bit longer than x86-64 platform, at this time. However, eventually, x86-64 platform test will take more time because sanitizers (#591), miri (#578), etc. will be added to x86-64 platform test. So I think this is fine.

jeehoonkang: Thank you for the revised PR. I just want to suggest another (potential) alternative: can we also test x86-64 platform in cross?  If so, we can uniformly treat all platforms using cross.

taiki-e: cross [doesn't provide images for windows-msvc and darwin](https://github.com/rust-embedded/cross#supported-targets), so I don't think using cross on those platforms will no benefit.
However, I'll take a look for a way to merge the `test` and `cross` jobs into one job.

taiki-e:
@jeehoonkang Updated to test all crates at once.

- Tests that were previously run per crate are run at once using `--all` flag.
- Use cross if the target is set, otherwise, use cargo and test with the host target. See https://github.com/crossbeam-rs/crossbeam/pull/613#discussion_r547038713 for context.
  For now, the non-host target only runs tests.
- Replaced `sanitize` feature with `cfg(crossbeam_sanitize)`.
   Feature flag can be accidentally enabled by --all-features, so use cfg instead. (This issue actually occurred when I tried to use `--all-features` instead because `--features` nightly cannot be used together with `--all`.)
  `sanitize` feature is a feature for testing, not a feature intended to be used as a public API, so it should be okay to remove it.
- Separated check for docs into its own job.
  This is in a different area than the test and I think it makes sense to run it as a separate job.

.github/workflows/ci.yml
```diff
@@ -24,30 +24,31 @@ jobs:
...
-          - crossbeam-skiplist
-          - crossbeam-utils
-        rust:
-          - 1.36.0
-          - nightly
-        os:
-         - ubuntu-latest
-         - windows-latest
-    runs-on: ${{ matrix.os }}
+        include:
```
jeehoonkang, line 30: It seems we don't test Crossbeam for Rust stable.  Is it intended?

.github/workflows/ci.yml
```diff
@@ -24,30 +24,31 @@ jobs:
...
+          # TODO: https://github.com/crossbeam-rs/crossbeam/pull/518#issuecomment-633342606
+          # - rust: nightly
+          #   os: macos-latest
+          - rust: nightly
+            os: windows-latest
+          - rust: nightly
+            target: i686-unknown-linux-gnu
+          - rust: nightly
+            target: aarch64-unknown-linux-gnu
+    runs-on: ${{ matrix.os || 'ubuntu-latest' }}
```
jeehoonkang, line 44: We're effectively marking `ubuntu-latest` as the "default" platform here. I prefer not to select one as the default for simplicity. What do you think?

ci/test.sh
```diff
@@ -0,0 +1,50 @@
...
+cargo test --all --exclude benchmarks -- --test-threads=1
+
+if [[ "$RUST_VERSION" == "nightly"* ]]; then
+    # Some crates have `nightly` feature, so run tests with --all-features.
+    cargo test --all --all-features --exclude benchmarks -- --test-threads=1
+
+    # Benchmarks are only checked on nightly because depending on unstable features.
+    cargo check --all --benches
+    cd crossbeam-channel/benchmarks
+    cargo check --bins
```
jeehoonkang, line 28: maybe we can instead use it: https://github.com/rust-lang/cargo/issues/1485#issuecomment-89804438

jeehoonkang (commented):
Thanks for the cleanup!  I have a few comments.

.github/workflows/ci.yml
taiki-e: I don't think it was intended. Added the latest stable Rust to matrix.

.github/workflows/ci.yml
taiki-e: Indeed. Changed to always set `matrix.os`.

ci/test.sh
taiki-e: Sounds good! Switched to use --manifest-path.

jeehoonkang:
Thanks!

bors r+

PR #615, opened 2020-12-21
Port Golang chan1.go Test

SkamDart:
Rust port [chan1.go](https://github.com/golang/go/blob/master/test/ken/chan1.go)

Based on an ask in #201

### DIFF
--- a/crossbeam-channel/tests/golang.rs
+++ b/crossbeam-channel/tests/golang.rs
@@ -1527,7 +1527,6 @@ mod chan1 {
                             panic!("fail")
                         }
                         data[r] = 2;
-                        drop(data);
                     }
                 }
             }
@@ -1543,6 +1542,7 @@ mod chan1 {
                 panic!("fail");
             }
             data[r] = 1;
+            // https://github.com/crossbeam-rs/crossbeam/pull/615#discussion_r550281094
             drop(data);
             c.send(r);
         }
--- a/crossbeam-channel/tests/golang.rs
+++ b/crossbeam-channel/tests/golang.rs
@@ -428,6 +428,8 @@ mod nonblock {
             ticker.recv().unwrap();
             ticker.recv().unwrap();
             thread::yield_now();
+            thread::yield_now();
+            thread::yield_now();
         };
 
         let sync = make::<bool>(0);
--- a/crossbeam-channel/tests/golang.rs
+++ b/crossbeam-channel/tests/golang.rs
@@ -1520,14 +1520,13 @@ mod chan1 {
         loop {
             select! {
                 recv(c.rx()) -> rr => {
-                    if let Ok(r) = rr {
-                        let mut data = h.lock().unwrap();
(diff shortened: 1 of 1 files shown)


crossbeam-channel/tests/golang.rs
```diff
@@ -1507,5 +1507,59 @@ mod chan {
...
+            select! {
+                recv(c.rx()) -> rr => {
+                    if let Ok(r) = rr {
+                        let mut data = h.lock().unwrap();
+                        if data[r] != 1 {
+                            println!("r\nm={}\nr={}\nh={}\n", m, r, data[r]);
+                            panic!("fail")
+                        }
+                        data[r] = 2;
+                        drop(data);
```
SkamDart, line 1530: Someone who is more experienced with Rust than I am probably knows why, but this blocks unless I manually drop.
Any thoughts?

taiki-e: There may be an odd interaction with select macro. I'll take a look.

SkamDart: @taiki-e What do I need to do to get this PR merged in?

taiki-e: Even if I remove `drop(data)` in the `r` function, I couldn't reproduce block. I can reproduce block by removing the `drop(data)` in the `s` function, but it will not block if I disable parallel tests (`cargo test -- --test-threads=1`) or run the test individually (`cargo test --test golang -- chan1::main`).

@SkamDart Could you remove `drop(data)` in the `r` function and leave the link to this comment on the `drop(data)` in the `s` function as a comment? (Otherwise, LGTM)

crossbeam-channel/tests/golang.rs
```diff
@@ -1507,5 +1507,59 @@ mod chan {
...
+            select! {
+                recv(c.rx()) -> rr => {
+                    if let Ok(r) = rr {
+                        let mut data = h.lock().unwrap();
+                        if data[r] != 1 {
+                            println!("r\nm={}\nr={}\nh={}\n", m, r, data[r]);
+                            panic!("fail")
+                        }
+                        data[r] = 2;
+                    }
```
taiki-e, line 1530: If `rr` is an error, it would be preferable for the test to fail.
```suggestion
                    let r = rr.unwrap();
                    let mut data = h.lock().unwrap();
                    if data[r] != 1 {
                        println!("r\nm={}\nr={}\nh={}\n", m, r, data[r]);
                        panic!("fail")
                    }
                    data[r] = 2;
```

taiki-e (commented):
Thanks! LGTM aside from a nit.

crossbeam-channel/tests/golang.rs
SkamDart: Good catch! Thanks!

taiki-e (approved):
bors r+

PR #626, opened 2020-12-31
Fix clippy warnings

taiki-e:
https://github.com/crossbeam-rs/crossbeam/runs/1630617165?check_suite_focus=true


taiki-e:
Oh, `clippy::unnecessary_wraps` is not yet in stable clippy.
bors r-

https://github.com/crossbeam-rs/crossbeam/pull/626/checks?check_run_id=1630666111

PR #628, opened 2021-01-01
Add Atomic::compare_exchange(_weak) and deprecate Atomic::compare_and_set(_weak)

### LINKED ISSUE #621: Remove CompareAndSetOrdering trait and change compere_and_set to always use two orderings
Given that the standard library deprecated the methods that use a single ordering for CAS operation, I think it makes sense to remove `CompareAndSetOrdering` trait and change `compere_and_set` to always use two orderings.

Context: https://github.com/crossbeam-rs/crossbeam/issues/620#issuecomment-750888948

taiki-e:
Instead of changing the argument of `Atomic::compare_and_set(_weak)`, add new methods `Atomic::compare_exchange(_weak)` that always use two orderings.

This PR adds:

- `Atomic::compare_exchange`
  Unlike `compare_and_set`, this always receive two orderings.
- `Atomic::compare_exchange_weak`
  Unlike `compare_and_set_weak`, this always receive two orderings.
- `CompareExchangeError` (error returned by `Atomic::compare_exchange(_weak)`)
  The definition is the same as `CompareAndSetError`, just renamed `CompareExchangeError`.

This PR deprecates:

- [`Atomic::compare_and_set`](https://docs.rs/crossbeam-epoch/0.9/crossbeam_epoch/struct.Atomic.html#method.compare_and_set)
- [`Atomic::compare_and_set_weak`](https://docs.rs/crossbeam-epoch/0.9/crossbeam_epoch/struct.Atomic.html#method.compare_and_set_weak)
- [`CompareAndSetError`](https://docs.rs/crossbeam-epoch/0.9/crossbeam_epoch/struct.CompareAndSetError.html) (error returned by `Atomic::compare_and_set(_weak)`)
  This is now a type alias of `CompareExchangeError`.
- [`CompareAndSetOrdering`](https://docs.rs/crossbeam-epoch/0.9/crossbeam_epoch/trait.CompareAndSetOrdering.html)

Closes #621

### DIFF
--- a/crossbeam-epoch/src/atomic.rs
+++ b/crossbeam-epoch/src/atomic.rs
@@ -26,7 +26,11 @@ fn strongest_failure_ordering(ord: Ordering) -> Ordering {
 }
 
 /// The error returned on failed compare-and-set operation.
-pub struct CompareAndSetError<'g, T: ?Sized + Pointable, P: Pointer<T>> {
+// TODO: remove in the next major version.
+pub type CompareAndSetError<'g, T, P> = CompareExchangeError<'g, T, P>;
+
+/// The error returned on failed compare-and-swap operation.
+pub struct CompareExchangeError<'g, T: ?Sized + Pointable, P: Pointer<T>> {
     /// The value in the atomic pointer at the time of the failed operation.
     pub current: Shared<'g, T>,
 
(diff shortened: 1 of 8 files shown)


taiki-e:
I don't know why this method was originally called compare-and-set, but Wikipedia says a variant of compare-and-swap that returns a boolean indicating whether a substitution was performed is often called compare-and-set. https://en.wikipedia.org/wiki/Compare-and-swap

> The result of the operation must indicate whether it performed the substitution; this can be done either with a simple boolean response (this variant is often called **compare-and-set**), or by returning the value read from the memory location (not the value written to it).

As this method actually returns a value rather than a boolean, I think it's okay to call it `compare_exchange`. (Given that other similar methods in the standard library and crossbeam are called `compare_exchange`, It's probably more consistent.)

jeehoonkang:
std's compare_exchange receives two orderings: one for success and the other for fail cases. But this PR's compare_exchange receives only one that can be rendered into two. Do you think we need to match them?

taiki-e:
> But this PR's compare_exchange receives only one that can be rendered into two.

Uh, both methods (`compare_exchange(_weak)`) added by this PR always receive two orders, not `CompareAndSetOrdering`.
I guess what you are mentioning is the behavior of the methods (`compare_and_set(_weak)`) that will be deprecated by this PR.

https://github.com/crossbeam-rs/crossbeam/blob/4f0fcd5620662a245e815a501634199201f2a466/crossbeam-epoch/src/atomic.rs#L456-L465

https://github.com/crossbeam-rs/crossbeam/blob/4f0fcd5620662a245e815a501634199201f2a466/crossbeam-epoch/src/atomic.rs#L531-L540

taiki-e:
Updated the PR description to explain the changes in more detail.

PR #636, opened 2021-01-02
CachePadded: update alignment based on golang's CacheLinePadSize

### LINKED ISSUE #427: More complete implementation of CachePadded
Hi, I'm not sure if more platforms are interesting to you, but I've made a slightly more complete implementation of padding a struct to the cache line size here:

https://github.com/tinco/cache_line_size

I got the information on the cache line sizes from the Go compiler repository, so I'm pretty sure these are correct for all platforms Go runs on, which is a whole bunch.

taiki-e:
This updates alignment of `CachePadded` on the following architectures based on golang's [`CacheLinePadSize`](https://github.com/golang/go/blob/3510a1e32cbc86b73db143aefcc00aadc44c27bd/src/internal/cpu/cpu.go#L20).

- On [powerpc64](https://github.com/golang/go/blob/3dd58676054223962cd915bb0934d1f9f489d4d2/src/internal/cpu/cpu_ppc64x.go#L9), use 128-byte alignment.
- On [arm](https://github.com/golang/go/blob/3dd58676054223962cd915bb0934d1f9f489d4d2/src/internal/cpu/cpu_arm.go#L7), [mips](https://github.com/golang/go/blob/3dd58676054223962cd915bb0934d1f9f489d4d2/src/internal/cpu/cpu_mips.go#L7), [mips64](https://github.com/golang/go/blob/3dd58676054223962cd915bb0934d1f9f489d4d2/src/internal/cpu/cpu_mips64x.go#L9), and [riscv64](https://github.com/golang/go/blob/3dd58676054223962cd915bb0934d1f9f489d4d2/src/internal/cpu/cpu_riscv64.go#L7), use 32-byte alignment.
- On [s390x](https://github.com/golang/go/blob/3dd58676054223962cd915bb0934d1f9f489d4d2/src/internal/cpu/cpu_s390x.go#L7), use 256-byte alignment.

On x86_64 and aarch64, use 128-byte alignment as before. See #331 and #499 for reasons.
On all others, use 64-byte alignment as before.

Closes #427

cc @stjepang (when this merged, I'll send a same patch to smol-rs/cache-padded)
FYI @yoshuawuyts @cuviper (because you all were [discussing about adding `CachePadded` to the standard library](https://rust-lang.zulipchat.com/#narrow/stream/219381-t-libs/topic/.60std.3A.3Amem.3A.3ACachePadded.60))

### DIFF
--- a/crossbeam-utils/src/cache_padded.rs
+++ b/crossbeam-utils/src/cache_padded.rs
@@ -13,7 +13,9 @@ use core::ops::{Deref, DerefMut};
 ///
 /// Cache lines are assumed to be N bytes long, depending on the architecture:
 ///
-/// * On x86-64 and aarch64, N = 128.
+/// * On x86-64, aarch64, and powerpc64, N = 128.
+/// * On arm, mips, mips64, and riscv64, N = 32.
+/// * On s390x, N = 256.
 /// * On all others, N = 64.
 ///
 /// Note that N is just a reasonable guess and is not guaranteed to match the actual cache line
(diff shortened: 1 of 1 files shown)


jeehoonkang (approved):
I like this change, but I'm not an expert on this and I'd like one more +1.

yoshuawuyts (approved):
Oh amazing; yes!

Adding in the references is :100:

Vtec234 (approved):
There is still the issue of some AMD CPUs fetching just 64 (maybe? I'm not an expert), but that would have to be a runtime check since we don't compile for specific CPUs so LGTM.

novacrazy:
Indeed, all AMD CPUs still use a 64-byte cacheline size, so this implementation will use up two lines on AMD. 

Previously, I had used a build script to detect `target-cpu` and use that to control my own `CachePadded`, and in the past year there has been zero progress on better target introspection in Rust, so oh well.

taiki-e:
We may want to determine the principle regarding which cache line size `CachePadded` chooses in architectures where the actual size is implementation-defined. (Currently, on x86_64 and aarch64, we refer to architectures that use the largest cache line size.)

> Previously, I had used a build script to detect `target-cpu` and use that to control my own `CachePadded`, and in the past year there has been zero progress on better target introspection in Rust, so oh well.

I think we might be able to add an option (--cfg) to detect native CPU cache line size using a build script. (It is never enabled by default for the same reason that `-C target-cpu=native` is optional.) But for now, it may be better to implement it as a separate crate.

cuviper:
I think if you start digging that deep, it will be better served by the standard library and compiler intrinsics, where it has the canonical view of the target cpu and features.

PR #637, opened 2021-01-03
Apply unreachable_pub lint

taiki-e:
`unreachable_pub` lint warns items that are marked as `pub` but are not actually included in the public API, and suggests using `pub(crate)` instead.

`pub(crate)` is a bit more verbose than `pub`, but this should make it easier to review changes that make it difficult to tell if it will break API or not just by looking at the diff like https://github.com/crossbeam-rs/crossbeam/pull/617#discussion_r548370194. Especially, there are many such codes in crossbeam-channel and crossbeam-epoch. (Hopefully, after this patch, we can basically consider all items marked `pub` are public APIs, except for items that have been explicitly hidden from the public API by `#[doc(hidden)]`.)

This lint is known to have some false positives, but the crossbeam doesn't seem to be affected.

### DIFF
--- a/crossbeam-channel/src/flavors/mod.rs
+++ b/crossbeam-channel/src/flavors/mod.rs
@@ -9,9 +9,9 @@
 //! 5. `tick` - Channel that delivers messages periodically.
 //! 6. `zero` - Zero-capacity channel.
 
-pub mod array;
-pub mod at;
-pub mod list;
-pub mod never;
-pub mod tick;
-pub mod zero;
+pub(crate) mod array;
+pub(crate) mod at;
+pub(crate) mod list;
+pub(crate) mod never;
+pub(crate) mod tick;
+pub(crate) mod zero;
--- a/crossbeam-channel/src/flavors/never.rs
+++ b/crossbeam-channel/src/flavors/never.rs
@@ -11,17 +11,17 @@ use crate::select::{Operation, SelectHandle, Token};
 use crate::utils;
 
 /// This flavor doesn't need a token.
-pub type NeverToken = ();
+pub(crate) type NeverToken = ();
 
 /// Channel that never delivers messages.
-pub struct Channel<T> {
+pub(crate) struct Channel<T> {
     _marker: PhantomData<T>,
 }
 
 impl<T> Channel<T> {
     /// Creates a channel that never delivers messages.
     #[inline]
-    pub fn new() -> Self {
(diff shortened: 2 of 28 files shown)


crossbeam-channel/src/flavors/mod.rs
```diff
@@ -9,9 +9,9 @@
 //! 5. `tick` - Channel that delivers messages periodically.
 //! 6. `zero` - Zero-capacity channel.
 
-pub mod array;
-pub mod at;
-pub mod list;
-pub mod never;
-pub mod tick;
-pub mod zero;
+pub(crate) mod array;
```
jeehoonkang, line 12: Hmm, instead of this, can we make the `flavors` module `pub(crate)`?

crossbeam-channel/src/flavors/never.rs
```diff
@@ -11,62 +11,62 @@ use crate::select::{Operation, SelectHandle, Token};
...
-pub struct Channel<T> {
+pub(crate) struct Channel<T> {
     _marker: PhantomData<T>,
 }
 
 impl<T> Channel<T> {
     /// Creates a channel that never delivers messages.
     #[inline]
-    pub fn new() -> Self {
+    pub(crate) fn new() -> Self {
```
jeehoonkang, line 24: I'm hesitant about this change.  It seems `Channel` is already `pub(crate)`.  Then `pub(crate)` for its methods seems overkill.  What do you think?

jeehoonkang (changes requested):

Vtec234 (approved):
LGTM modulo the already raised issues.

crossbeam-channel/src/flavors/mod.rs
taiki-e: The parent modules are already private, but rustc says they are still unreachable `pub` because we can re-export those modules.

crossbeam-channel/src/flavors/never.rs
taiki-e: Indeed. If the struct is not public, then the method will never be public.

That said, there are cases where impl blocks are long, or the struct definitions are in different files, so I don't think it's really that bad to mark them as `pub(crate)` as unreachable_pub lint suggested.

jeehoonkang: Okay... Let's follow rustc's suggestions. At worst, it's not part of the public API :)

PR #638, opened 2021-01-03
[WIP] crossbeam-epoch: expose loom support via feature flag

tobz:
This PR exposes the `loom_crossbeam` feature flag to enable internal `loom` usage, rather than relying on `cfg`-based settings.

The usage of `--cfg` flags is cumbersome as it forces recompilation of all crates given that `RUSTFLAGS` has changed.

### DIFF
--- a/crossbeam-epoch/Cargo.toml
+++ b/crossbeam-epoch/Cargo.toml
@@ -36,17 +36,17 @@ nightly = ["crossbeam-utils/nightly", "const_fn"]
 # TODO: docs
 sanitize = [] # Makes it more likely to trigger any potential data races.
 
-[dependencies]
-cfg-if = "1"
-const_fn = { version = "0.4.4", optional = true }
-memoffset = "0.6"
-
 # Enable the use of loom for concurrency testing.
 #
 # This configuration option is outside of the normal semver guarantees: minor
 # versions of crossbeam may make breaking changes to it at any time.
-[target.'cfg(loom_crossbeam)'.dependencies]
-loom = "0.4"
+loom_crossbeam = []
+
+[dependencies]
+cfg-if = "1"
+const_fn = { version = "0.4.4", optional = true }
+loom = { version = "0.4", optional = true }
+memoffset = "0.6"
 
 [dependencies.crossbeam-utils]
 version = "0.8"


taiki-e:
I understand the motivation, but the cargo feature needs to be additive and I don't think the loom feature meets that requirement.

tobz:
I'm happy to stick with the `--cfg` based approach if the feature approach isn't workable, but can you explain a bit more?

From my perspective, the feature flag would enable pulling in `loom` itself as well as flipping the codepaths from std types to `loom` types.  Is there something else that I'm missing.

taiki-e:
> can you explain a bit more?

Feature flags are silently shared across the dependencies graph.
For example, if I create a crate for the test that always depends on the loom feature and I run `cargo test` with `--all`/`--workspace`, the loom feature will be enabled for all crates. This will break tests on all other crates. 
See also https://github.com/rust-lang/cargo/issues/4328, https://github.com/rust-lang/rfcs/pull/1841, https://github.com/rust-lang-nursery/lazy-static.rs/issues/150, https://github.com/serde-rs/serde/issues/918, https://rust-lang.github.io/api-guidelines/print.html#feature-names-are-free-of-placeholder-words-c-feature

> the feature flag would enable pulling in `loom` itself as well as flipping the codepaths from std types to `loom` types.

This also applies when using cfg, right?

tobz:
> Feature flags are silently shared across the dependencies graph.
> For example, if I create a crate for the test that always depends on the loom feature and I run `cargo test` with `--all`/`--workspace`, the loom feature will be enabled for all crates. This will break tests on all other crates.
> See also [rust-lang/cargo#4328](https://github.com/rust-lang/cargo/issues/4328), [rust-lang/rfcs#1841](https://github.com/rust-lang/rfcs/pull/1841), [rust-lang-nursery/lazy-static.rs#150](https://github.com/rust-lang-nursery/lazy-static.rs/issues/150), [serde-rs/serde#918](https://github.com/serde-rs/serde/issues/918), https://rust-lang.github.io/api-guidelines/print.html#feature-names-are-free-of-placeholder-words-c-feature

I see what you mean now, and this makes sense.  Appreciate the links. 👍🏻

PR #639, opened 2021-01-04
Add list of main items to docs of utils and atomic modules

taiki-e:
Both [`sync`](https://docs.rs/crossbeam/0.8.0/crossbeam/sync/index.html) and [`queue`](https://docs.rs/crossbeam/0.8.0/crossbeam/queue/index.html) modules have similar docs.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,8 +59,12 @@
 #[cfg_attr(feature = "nightly", cfg(target_has_atomic = "ptr"))]
 pub use crossbeam_utils::atomic;
 
-/// Miscellaneous utilities.
 pub mod utils {
+    //! Miscellaneous utilities.
+    //!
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -59,8 +59,12 @@
 #[cfg_attr(feature = "nightly", cfg(target_has_atomic = "ptr"))]
 pub use crossbeam_utils::atomic;
 
-/// Miscellaneous utilities.
 pub mod utils {
+    //! Miscellaneous utilities.
+    //!
+    //! * [`Backoff`], for exponential backoff in spin loops.
+    //! * [`CachePadded`], for padding and aligning a value to the length of a cache line.
```
taiki-e, line 66: By the way, this changed from outer doc to inner doc because it was needed to write link like `[`Backoff`](crate:: utils::Backoff)` when using outer doc.

PR #642, opened 2021-01-04
Add macOS to CI

taiki-e:
Currently, 1/4~1/3 tests in crossbeam-channel are ignored due to time-related assertions failed due to GitHub Actions' macOS runner is slow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,9 +40,8 @@ jobs:
             os: windows-latest
           - rust: nightly
             os: ubuntu-latest
-          # TODO: https://github.com/crossbeam-rs/crossbeam/pull/518#issuecomment-633342606
-          # - rust: nightly
-          #   os: macos-latest
+          - rust: nightly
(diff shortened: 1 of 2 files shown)


cynecx:
Is it possible to add `macos-11.0` with the Apple Silicon/M1 as target to the ci? (aka `--target aarch64-apple-darwin`).

taiki-e:
@cynecx 
I think we can add `aarch64-apple-darwin` if we can run tests and the environment is stable enough. 
However, the current `macos-11.0` is still a preview and not stable for GitHub Actions (https://github.com/actions/virtual-environments/issues/1814).
(I recently tried it, but the workflow didn't start.)

cynecx:
@taiki-e Sounds like a plan. (Yeh I've noticed the funky availability of `macos-11.0` environments too.)

taiki-e:
Seems macos-11.0 is moved to private preview: https://github.com/actions/virtual-environments/issues/2486

PR #649, opened 2021-01-24
Add `#![doc(html_root_url)]` to all crates

a1phyr:
It enables external crates' documentation to link to `crossbeam`'s one.


taiki-e:
`html_root_url` is not needed in most cases. And api-guidelines no longer recommends this. See https://github.com/rust-lang/api-guidelines/issues/229 for more.

taiki-e (commented):
As discussed in the linked issue, I don't think this will bring any benefits beyond the maintenance burden, so I'll close this. Thanks anyway @a1phyr!

PR #664, opened 2021-02-20
Add Covariance To Some Type Paramaters

danii:
crossbeam_epoc::Atomic<T>, crossbeam_deque::Worker<T> & crossbeam_deque::Stealer<T>: Covariance Over T

At my understanding, covariance is okay because Atomic<T>, Worker<T>, and Stealer<T> all appear to use owning pointers. Owning pointers such as Box<T> can safely be covariant over T. Don't take my word for this though.

### DIFF
--- a/crossbeam-deque/src/deque.rs
+++ b/crossbeam-deque/src/deque.rs
@@ -29,7 +29,7 @@ const FLUSH_THRESHOLD_BYTES: usize = 1 << 10;
 /// *not* deallocate the buffer.
 struct Buffer<T> {
     /// Pointer to the allocated memory.
-    ptr: *mut T,
+    ptr: *const T,
 
     /// Capacity of the buffer. Always a power of two.
     cap: usize,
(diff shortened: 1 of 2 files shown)


taiki-e:
This makes a change like #490 a breaking change, right?
If so, I would like to keep the current one for forward compatibility unless there are specific issues that this change actually improves.

danii:
Yes this is a breaking change on your end. These type parameters were invariant before, and making them covariant grants end users more flexibility (able to satisfy them with more values than just the one that is required).

Although now that I have thought about it, my proposal was invalid (at least for crossbeam_epoch::Atomic), because now that I've looked at it, even if that type is used as an owned pointer, it is atomic... `&mut Box<T>` is invariant because of the `&mut`, and that type is, variance wise, the same as `Atomic<T>`.
Hey at least I tried lol. I still think Buffer can be covariant even though it wouldn't change anything in the public API.

Also, the code I wrote when I thought I discovered this bug now compiles without this fix after writing some more with it, I probably should've spent more time finding and writing out the correct lifetimes.

PR #666, opened 2021-02-23
Make loom dependency optional

### LINKED ISSUE #665: Hide loom from dependency graph
The way loom is set up now, loom and its dependencies show up in my dep graph as soon as I depend on crossbeam-utils or crossbeam-epoch. It would be nice to avoid this, for example as done in https://github.com/hawkw/sharded-slab/pull/52/files. (I can do the legwork on this if you like.)

taiki-e:
The loom dependency will never compile unless users enable cfg, but the way cargo and crates.io handle cfg-ed dependencies is not very good and confuses users.

This issue seems to work around by the way proposed by @matklad in https://github.com/crossbeam-rs/crossbeam/pull/487#issuecomment-783385606.
So this PR uses that workaround.

Closes #665

r? @jeehoonkang @jonhoo

### DIFF
--- a/ci/crossbeam-epoch-loom.sh
+++ b/ci/crossbeam-epoch-loom.sh
@@ -8,4 +8,4 @@ export RUSTFLAGS="-D warnings --cfg crossbeam_loom --cfg crossbeam_sanitize"
 # With MAX_PREEMPTIONS=2 the loom tests (currently) take around 11m.
 # If we were to run with =3, they would take several times that,
 # which is probably too costly for CI.
-env LOOM_MAX_PREEMPTIONS=2 cargo test --test loom --release -- --nocapture
+env LOOM_MAX_PREEMPTIONS=2 cargo test --test loom --release --features loom -- --nocapture
--- a/crossbeam-epoch/Cargo.toml
+++ b/crossbeam-epoch/Cargo.toml
@@ -32,6 +32,12 @@ alloc = []
 # of crossbeam may make breaking changes to them at any time.
 nightly = ["crossbeam-utils/nightly", "const_fn"]
 
+# Enable the use of loom for concurrency testing.
+#
+# This configuration option is outside of the normal semver guarantees: minor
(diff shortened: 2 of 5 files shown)


taiki-e:
bors d=jonhoo

taiki-e:
@matklad @djc: Do you need me to do a new release?

matklad:
I would appreciate one!

taiki-e:
Published in crossbeam-utils 0.8.3 & crossbeam-epoch 0.9.3.

djc:
@taiki-e thanks for getting this resolved so quickly!

faern:
Do you people remember why you put loom both behind a `cfg` *and* made it optional? I see you still have this solution in place. I faced the same issue as this PR solves, but I found no downside of only making it an optional dependency: https://github.com/faern/oneshot/pull/43

taiki-e:
@faern See https://github.com/crossbeam-rs/crossbeam/pull/487#issuecomment-783365041 (and subsequent 2 comments) and https://github.com/crossbeam-rs/crossbeam/pull/638 for related discussion. (TL;DR that is not compliant with Cargo's recommendations)

faern:
I have read those, and it did not shed any light on it for me. I'm not really asking why it's optional, rather why it's behind `cfg(loom)`. I know the feature goes against the rule that features should be additive. Also internals for testing should not be exposed with features.

But changing from `[dependencies] loom = ...` to `[target.'cfg(loom)'.dependencies] loom = ...` does not help from what I can tell. The loom feature is still listed in `cargo metadata`. And my crate is built with the loom feature on if I do `cargo build --all-features`. The only difference is that if I guard loom both behind being optional and `[target.'cfg(loom)'.dependencies]` is that if a downstream user builds with the loom feature enabled, they still get standard functionality (not loom)... But that's ugly. Having them build with loom is also ugly, but yeah.

Is your motivation that by making it a cfg, even if it is exposed as a feature, enabling it does nothing for downstream?

PR #669, opened 2021-02-28
channel: discard messages eagerly

BusyJay:
If all receivers are dropped, the memory hold by channel will only be
dropped when all senders are also dropped. If there are too many
channels the memory leak can be significant.


taiki-e (approved):
LGTM, thanks!

bors r+

taiki-e:
Published in crossbeam-channel 0.5.1

PR #676, opened 2021-03-26
Format with rustfmt

taiki-e:
https://github.com/crossbeam-rs/crossbeam/runs/2198421682?check_suite_focus=true


taiki-e (commented):
bors r+

PR #683, opened 2021-04-10
Prepare for the next release

taiki-e:
Changes:

- crossbeam-channel 0.5.0 -> 0.5.1
  - Fix memory leak in unbounded channel. (#669)


taiki-e (commented):
bors r+

there is no API change

PR #684, opened 2021-04-10
Use stable compiler to installing cargo-hack

taiki-e:
I'm planning to increase the Rust version required for the installation of cargo-hack from 1.36 to 1.40. (cargo-hack is usually runnable with Cargo versions older than the Rust version required for installation. Rust 1.36 will continue to be supported after the increase the Rust version required for the installation of cargo-hack has been increased.)


taiki-e (commented):
bors r+

taiki-e:
bors retry

PR #687, opened 2021-04-28
Address clippy warning

erickt:
Clippy doesn't like this pattern:

```
for i in 0..batch_size {
  ...
  batch_size = i;
  ...
}
```

because it can trick readers into thinking the loop condition might be
modified during iteration. This avoids the warning by explicitly making
a new variable for the end condition of the loop.


taiki-e (approved):
Thanks!

bors r+

PR #690, opened 2021-04-28
epoch: Correct handling of null ptr returned from alloc::alloc

### LINKED ISSUE #689: crossbeam-epoch's atomic.rs doesn't handle alloc::alloc returning null pointers
In Fuchsia, we're doing an [update](https://fuchsia-review.googlesource.com/c/fuchsia/+/517844) of our vendored version of crossbeam. As part of it, we found some uses of `alloc::alloc` that may be incorrect, like https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-epoch/src/atomic.rs#L254. According to the [documentation](https://doc.rust-lang.org/std/alloc/trait.GlobalAlloc.html#errors) for GlobalAlloc, it's possible for allocation to return a null pointer in two cases:

* memory is exhausted
* layout doesn't match the allocator's size or alignment constraints.

This risks undefined behavior if these conditions are met.

taiki-e:
Fixes #689


jeehoonkang (approved):
bors r+

PR #691, opened 2021-04-29
Prepare for the next release

taiki-e:
- crossbeam-utils 0.8.3 -> 0.8.4
  - Bump `loom` dependency to version 0.5. (#686)
- crossbeam-epoch 0.9.3 -> 0.9.4
  - Fix UB in `<[MaybeUninit<T>] as Pointable>::init` when global allocator failed allocation (#690)
  - Bump `loom` dependency to version 0.5. (#686)

### DIFF
--- a/crossbeam-epoch/CHANGELOG.md
+++ b/crossbeam-epoch/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Version 0.9.4
+
+- Fix UB in `<[MaybeUninit<T>] as Pointable>::init` when global allocator failed allocation (#690)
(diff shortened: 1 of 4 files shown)


taiki-e (commented):
bors r+

I think it is highly likely that crossbeam's nightly feature will be broken in the latest nightly within a week (due to https://github.com/rust-lang/rust/issues/84510, https://github.com/rust-lang/rust/pull/84556). However, since rustfmt is currently not available in the latest nightly, most people will use the older nightly. Therefore, I think it would actually be reasonable to release this fix now, and then release a new version that fixes nightly functionality in a week if necessary.

PR #692, opened 2021-04-30
Use feature(const_fn_trait_bound) instead of feature(const_fn)

taiki-e:
error[E0658]: trait bounds other than `Sized` on const fn parameters are unstable
       --> crossbeam-epoch/src/atomic.rs:313:6
        |
    313 | impl<T: ?Sized + Pointable> Atomic<T> {
        |      ^
        |
        = note: see issue #57563 <https://github.com/rust-lang/rust/issues/57563> for more information
        = help: add `#![feature(const_fn_trait_bound)]` to the crate attributes to enable


taiki-e (commented):
bors r+ 

See also  rust-lang/rust#84510, rust-lang/rust#84556

PR #694, opened 2021-05-04
epoch: Fix UB in Pointable impl of [MaybeUninit<T>]

### LINKED ISSUE #693: Length mismatch in array init causes buffer overflow
When initializing an array, the value put in the length field is equal to the total number of allocated bytes, not the number of elements the array can hold.

Example:
```rust
use std::mem::MaybeUninit;
use crossbeam::epoch::*;

pub fn main() {
	let owned = Owned::<[MaybeUninit<usize>]>::init(10);
	let arr: &[MaybeUninit<usize>] = &*owned;
	println!("{}", arr.len());
}
```
This should print `10`, but prints `88`. In particular, it has only allocated space for 10 elements of data, so creation of `arr` in the example is UB.

taiki-e:
Fixes #693

### DIFF
--- a/crossbeam-epoch/src/atomic.rs
+++ b/crossbeam-epoch/src/atomic.rs
@@ -238,7 +238,8 @@ impl<T> Pointable for T {
 // [`alloc::alloc::Layout::extend`] instead.
 #[repr(C)]
 struct Array<T> {
-    size: usize,
+    /// The number of elements (not the number of bytes).
+    len: usize,
     elements: [MaybeUninit<T>; 0],
 }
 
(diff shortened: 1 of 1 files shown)


taiki-e:
@jeehoonkang Are there any changes that you'd like me to make?

crossbeam-epoch/src/atomic.rs
```diff
@@ -238,6 +238,7 @@ impl<T> Pointable for T {
 // [`alloc::alloc::Layout::extend`] instead.
 #[repr(C)]
 struct Array<T> {
+    /// The number of elements (not the number of bytes).
     size: usize,
```
jeehoonkang, line 242: @taiki-e maybe you also want to rename it to `len`?

jeehoonkang (commented):
Sorry for a late response. LGTM, but left a minor comment. Please feel free to merge this.

taiki-e:
@jeehoonkang Thanks for the review! I renamed `Array::size` field to `len` as you suggested.

bors r+

PR #695, opened 2021-05-04
Fixed array initialization bug

### LINKED ISSUE #693: Length mismatch in array init causes buffer overflow

PatrickNorton:
Should fix #693.

### DIFF
--- a/crossbeam-epoch/src/atomic.rs
+++ b/crossbeam-epoch/src/atomic.rs
@@ -248,9 +248,9 @@ impl<T> Pointable for [MaybeUninit<T>] {
     type Init = usize;
 
     unsafe fn init(size: Self::Init) -> usize {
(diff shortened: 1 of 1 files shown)


taiki-e (commented):
Thanks for the PR!
I'm going to close this in favor of #694 because it was filed a bit earlier and has a test for #693.

PR #697, opened 2021-05-11
fixed go.go to properly select over channels during the put operation in select_both

AidanGoldfarb:
Per the [crossbeam benchmark configuration](https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-channel/benchmarks/README.md) the `select_both` function is defined as: 

> select_both: T threads send N / T messages each by selecting over T channels. T other threads receive N / T messages each by selecting over the T channels.

Previously, the `select_both` function in [go.go](https://github.com/crossbeam-rs/crossbeam/blob/3e83987f258ab85a6573704ba538f2efdfe587c1/crossbeam-channel/benchmarks/go.go#L128) did not utilize the `select` statement while sending messages. Added the use of the select statement.

### DIFF
--- a/crossbeam-channel/benchmarks/go.go
+++ b/crossbeam-channel/benchmarks/go.go
@@ -139,10 +139,10 @@ func select_both(cap int) {
     var producer = func(c0 chan Message, c1 chan Message, c2 chan Message, c3 chan Message) {
         for i := 0; i < MESSAGES / THREADS; i++ {
             select {
-            case c0 <- Message(i);
-            case c1 <- Message(i);
-            case c2 <- Message(i);
-            case c3 <- Message(i);
+            case c0 <- Message(i):
+            case c1 <- Message(i):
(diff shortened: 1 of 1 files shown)


taiki-e:
Thanks for the PR! I'm curious how this change will change the benchmark results.

AidanGoldfarb:
> Thanks for the PR! I'm curious how this change will change the benchmark results.

Here are my teams benchmarking results. Note this was done in during a project involving implementing channels in the Flix programming language, so they additional test results are our additions.

results

All tests were done on a machine 'node2x18a' at the University of Rochester cluster

>node2x18 has two Intel(R) Xeon(R) CPU E5-2695 v4 @ 2.10GHz.
> 1 MiB of L1d and L1i caches. The system also has a 32 MiB L2 cache and 44 MiB L3 cache.

taiki-e (approved):
bors r+

PR #698, opened 2021-05-15
Remove uses of unstable feature(cfg_target_has_atomic)

taiki-e:
Some no-std targets (e.g., ARMv6-M) do not support atomic CAS operations and cannot use Arc, etc.

Currently, we are using an unstable feature to detect them, but it has caused breakage in the past (https://github.com/crossbeam-rs/crossbeam/issues/435).
Also, users of stable Rust are not able to compile crossbeam on those targets.

Instead of depending on unstable features of the compiler, this patch detects those targets using the TARGET environment variables provided by cargo for the build script, and a list of targets that do not support atomic CAS operations.

This way is the same as the way we recently adopted in [futures](https://github.com/rust-lang/futures-rs/pull/2400) and [valuable](https://github.com/tokio-rs/valuable/pull/12), and was originally inspired by the way [heapless](https://github.com/japaric/heapless/commit/44c66a74845eacd3f18eae5be8ae6d887f4f7f0b) and [defmt](https://github.com/knurling-rs/defmt/blob/963152f0fc530fca64ba4ff1492d9c4b7bf76062/build.rs#L42-L51) do, but this doesn't maintain the target list manually. (It's not really fully automated, but [it's very easy to update](https://github.com/crossbeam-rs/crossbeam/blob/a42dbed87a5739228b576f526b1e2fd80260a29b/.github/workflows/ci.yml#L89).)

Also, this completely removes the dependency on unstable features from crates other than crossbeam-epoch.

refs: rust-lang/rust#51953, rust-lang/futures-rs#2400, tokio-rs/valuable#12


Cightline:
Appreciate the work, I recently encountered this issue.

ci/no_atomic.sh
```diff
@@ -26,6 +26,31 @@ for target in $(rustc --print target-list); do
...
+# `"target-pointer-width" == "32"`, so assuming them `"max-atomic-width" == 32`
+# for now.
+for target in $(rustc --print target-list); do
+    res=$(rustc --print target-spec-json -Z unstable-options --target "$target" \
+        | jq -r "select(.\"max-atomic-width\" == null)")
+    [[ -z "$res" ]] || echo "    \"$target\"," >>"$file"
+done
+echo "];" >>"$file"
+
+# There is no `"max-atomic-width" == 16` or `"max-atomic-width" == 8` targets.
```
taiki-e, line 52: Both "avr-unknown-gnu-atmega328" and "msp430-none-elf", which are `"target-pointer-width" == "16"`, do not support atomic:
- https://github.com/rust-lang/rust/blob/1.52.1/compiler/rustc_target/src/spec/avr_gnu_base.rs
- https://github.com/rust-lang/rust/blob/1.52.1/compiler/rustc_target/src/spec/msp430_none_elf.rs

PR #700, opened 2021-05-23
Fix windows CI

taiki-e:


taiki-e (commented):
bors r+

This is a workaround of the issues like https://github.com/rust-lang/rustup/issues/2748, https://github.com/rust-lang/rustup/issues/2759

taiki-e:
bors retry

PR #701, opened 2021-05-23
Add {Atomic, Shared}::try_into_owned

PatrickNorton:
Sometimes, I have run into times when I have had an `Atomic` which I know to be either null or valid, and I want to convert it into an `Owned`. As far as I can tell, there is no way to do this without converting through a `Shared`, which comes at the const of an additional atomic load. This pull request also adds the same method to `Shared` for symmetry.

This pull request adds the following methods:
```rust
impl<T: ?Sized + Pointable> Owned<T> {
	pub unsafe fn try_into_owned(self) -> Option<Owned<T>>;
}

impl<T: ?Sized + Pointable> Shared<'_, T> {
	pub unsafe fn try_into_owned(self) -> Option<Owned<T>>;
}
```

### DIFF
--- a/crossbeam-epoch/src/atomic.rs
+++ b/crossbeam-epoch/src/atomic.rs
@@ -809,6 +809,52 @@ impl<T: ?Sized + Pointable> Atomic<T> {
             Owned::from_usize(self.data.into_inner())
         }
     }
+
+    /// Takes ownership of the pointee if it is non-null.
+    ///
+    /// This consumes the atomic and converts it into [`Owned`]. As [`Atomic`] doesn't have a
+    /// destructor and doesn't drop the pointee while [`Owned`] does, this is suitable for
+    /// destructors of data structures.
+    ///
+    /// # Safety
+    ///
+    /// This method may be called only if the pointer is valid and nobody else is holding a
+    /// reference to the same object, or the pointer is null..
+    ///
+    /// # Examples
+    ///
+    /// ```rust
+    /// # use std::mem;
+    /// # use crossbeam_epoch::Atomic;
+    /// struct DataStructure {
+    ///     ptr: Atomic<usize>,
+    /// }
+    ///
+    /// impl Drop for DataStructure {
+    ///     fn drop(&mut self) {
+    ///         // By now the DataStructure lives only in our thread and we are sure we don't hold
+    ///         // any Shared or & to it ourselves, but it may be null, so we have to be careful.
+    ///         let old = mem::replace(&mut self.ptr, Atomic::null());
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks for the PR and sorry for the late review!

bors r+

crossbeam-epoch/src/atomic.rs
```diff
@@ -809,6 +809,52 @@ impl<T: ?Sized + Pointable> Atomic<T> {
...
+    /// Takes ownership of the pointee if it is non-null.
+    ///
+    /// This consumes the atomic and converts it into [`Owned`]. As [`Atomic`] doesn't have a
+    /// destructor and doesn't drop the pointee while [`Owned`] does, this is suitable for
+    /// destructors of data structures.
+    ///
+    /// # Safety
+    ///
+    /// This method may be called only if the pointer is valid and nobody else is holding a
+    /// reference to the same object, or the pointer is null..
```
taiki-e, line 822:
```suggestion
    /// reference to the same object, or the pointer is null.
```

crossbeam-epoch/src/atomic.rs
```diff
@@ -809,6 +809,52 @@ impl<T: ?Sized + Pointable> Atomic<T> {
...
+    ///     ptr: Atomic<usize>,
+    /// }
+    ///
+    /// impl Drop for DataStructure {
+    ///     fn drop(&mut self) {
+    ///         // By now the DataStructure lives only in our thread and we are sure we don't hold
+    ///         // any Shared or & to it ourselves, but it may be null, so we have to be careful.
+    ///         let old = mem::replace(&mut self.ptr, Atomic::null());
+    ///         unsafe {
+    ///             if let Option::Some(x) = old.try_into_owned() {
```
taiki-e, line 839:
```suggestion
    ///             if let Some(x) = old.try_into_owned() {
```

crossbeam-epoch/src/atomic.rs
```diff
@@ -809,6 +809,52 @@ impl<T: ?Sized + Pointable> Atomic<T> {
...
+    /// }
+    /// ```
+    pub unsafe fn try_into_owned(self) -> Option<Owned<T>> {
+        // FIXME: See self.into_owned()
+        #[cfg(crossbeam_loom)]
+        let data = self.data.unsync_load();
+        #[cfg(not(crossbeam_loom))]
+        let data = self.data.into_inner();
+        if decompose_tag::<T>(data).0 == 0 {
+            Option::None
```
taiki-e, line 853:
```suggestion
            None
```

PR #704, opened 2021-05-28
Add AtomicCell::fetch_update

taiki-e:
Equivalent of [`std::sync::atomic::AtomicN::fetch_update`](https://doc.rust-lang.org/nightly/core/sync/atomic/struct.AtomicUsize.html#method.fetch_update) that stabilized in Rust 1.45.


PatrickNorton:
Would it be a good idea to add a `fetch_update` method to `Atomic` as well? If so, I'd be happy to write it.

taiki-e:
@PatrickNorton It seems a good idea. I would be happy to accept a PR to add `Atomic::fetch_update`.

PR #706, opened 2021-05-29
Add Atomic::fetch_update

PatrickNorton:
Equivalent of [`AtomicN::fetch_update`](https://doc.rust-lang.org/nightly/core/sync/atomic/struct.AtomicUsize.html#method.fetch_update) and [`AtomicCell::fetch_update`](https://github.com/crossbeam-rs/crossbeam/pull/704).

### DIFF
--- a/crossbeam-epoch/src/atomic.rs
+++ b/crossbeam-epoch/src/atomic.rs
@@ -597,7 +597,7 @@ impl<T: ?Sized + Pointable> Atomic<T> {
     /// ```
     pub fn fetch_update<'g, F>(
         &self,
-        func: F,
+        mut func: F,
         set_ord: Ordering,
         fail_ord: Ordering,
         guard: &'g Guard,
--- a/crossbeam-epoch/src/atomic.rs
+++ b/crossbeam-epoch/src/atomic.rs
@@ -588,7 +588,7 @@ impl<T: ?Sized + Pointable> Atomic<T> {
     /// # Examples
     ///
     /// ```
-    /// use crossbeam_epoch::{self as epoch, Atomic, Owned};
+    /// use crossbeam_epoch::{self as epoch, Atomic};
     /// use std::sync::atomic::Ordering::SeqCst;
     ///
     /// let a = Atomic::new(1234);
--- a/crossbeam-epoch/src/atomic.rs
+++ b/crossbeam-epoch/src/atomic.rs
@@ -594,49 +594,30 @@ impl<T: ?Sized + Pointable> Atomic<T> {
     /// let a = Atomic::new(1234);
     /// let guard = &epoch::pin();
     ///
-    /// let res1 = a.fetch_update(|x, guard| Some(x.with_tag(1)), SeqCst, SeqCst, guard);
+    /// let res1 = a.fetch_update(SeqCst, SeqCst, guard, |x, guard| Some(x.with_tag(1)));
     /// assert!(res1.is_ok());
     ///
-    /// let res2 = a.fetch_update(|x, guard| None, SeqCst, SeqCst, guard);
+    /// let res2 = a.fetch_update(SeqCst, SeqCst, guard, |x, guard| None);
     /// assert!(res2.is_err());
     /// ```
     pub fn fetch_update<'g, F>(
         &self,
-        mut func: F,
-        set_ord: Ordering,
-        fail_ord: Ordering,
+        set_order: Ordering,
+        fail_order: Ordering,
         guard: &'g Guard,
+        mut func: F,
     ) -> Result<Shared<'g, T>, Shared<'g, T>>
     where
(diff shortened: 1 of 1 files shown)


crossbeam-epoch/src/atomic.rs
```diff
@@ -561,6 +561,84 @@ impl<T: ?Sized + Pointable> Atomic<T> {
...
+    ///
+    /// let res2 = a.fetch_update(|x, guard| None, SeqCst, SeqCst, guard);
+    /// assert!(res2.is_err());
+    /// ```
+    pub fn fetch_update<'g, F>(
+        &self,
+        mut func: F,
+        set_ord: Ordering,
+        fail_ord: Ordering,
+        guard: &'g Guard,
```
taiki-e, line 608: Could you move the closure to the last argument? (like std's [`AtomicN::fetch_update`](https://doc.rust-lang.org/nightly/std/sync/atomic/struct.AtomicUsize.html#method.fetch_update). see https://github.com/rust-lang/rust/pull/71843#issue-412603003 for reasons)

crossbeam-epoch/src/atomic.rs
```diff
@@ -561,6 +561,84 @@ impl<T: ?Sized + Pointable> Atomic<T> {
...
+                }
+            }
+            Err(prev)
+        }
+
+        fetch_update(&self.data, set_ord, fail_ord, |x| {
+            func(unsafe { Shared::from_usize(x) }, guard).map(Shared::into_usize)
+        })
+        .map(|old| unsafe { Shared::from_usize(old) })
+        .map_err(|current| unsafe { Shared::from_usize(current) })
```
taiki-e, line 639: Why did you not use `Atomic::compare_exchange_weak` directly?
```rust
        let mut prev = self.load(fetch_order, guard);
        while let Some(next) = f(prev, guard) {
            match self.compare_exchange_weak(prev, next, set_order, fetch_order, guard) {
                Ok(shared) => return Ok(shared),
                Err(next_prev) => prev = next_prev.current,
            }
        }
        Err(prev)
```

crossbeam-epoch/src/atomic.rs
```diff
@@ -561,6 +561,84 @@ impl<T: ?Sized + Pointable> Atomic<T> {
...
+    /// assert!(res1.is_ok());
+    ///
+    /// let res2 = a.fetch_update(|x, guard| None, SeqCst, SeqCst, guard);
+    /// assert!(res2.is_err());
+    /// ```
+    pub fn fetch_update<'g, F>(
+        &self,
+        mut func: F,
+        set_ord: Ordering,
+        fail_ord: Ordering,
```
taiki-e, line 607: I feel that the variable names used in std [`AtomicN::fetch_update`](https://doc.rust-lang.org/nightly/std/sync/atomic/struct.AtomicUsize.html#method.fetch_update) are clearer.
```suggestion
        set_order: Ordering,
        fetch_order: Ordering,
```

taiki-e (commented):
Thanks!

crossbeam-epoch/src/atomic.rs
```diff
@@ -561,6 +561,65 @@ impl<T: ?Sized + Pointable> Atomic<T> {
...
+    /// ```
+    pub fn fetch_update<'g, F>(
+        &self,
+        set_order: Ordering,
+        fail_order: Ordering,
+        guard: &'g Guard,
+        mut func: F,
+    ) -> Result<Shared<'g, T>, Shared<'g, T>>
+    where
+        F: for<'a> FnMut(Shared<'a, T>, &'a Guard) -> Option<Shared<'a, T>>,
```
taiki-e, line 611: Why do you use a different lifetime here instead of `'g`?

PatrickNorton: I'm not entirely sure anymore; thinking about it again, I think that makes more sense.

crossbeam-epoch/src/atomic.rs
```diff
@@ -561,6 +561,65 @@ impl<T: ?Sized + Pointable> Atomic<T> {
...
+    /// ```
+    pub fn fetch_update<'g, F>(
+        &self,
+        set_order: Ordering,
+        fail_order: Ordering,
+        guard: &'g Guard,
+        mut func: F,
+    ) -> Result<Shared<'g, T>, Shared<'g, T>>
+    where
+        F: FnMut(Shared<'g, T>, &'g Guard) -> Option<Shared<'g, T>>,
```
taiki-e, line 611: Hmm, given that `&'g Guard` is an immutable reference, I'm not sure if we need to pass it as an argument to the closure.

taiki-e (approved):
LGTM. Thanks @PatrickNorton!

bors r+

PatrickNorton:
Thanks for your review help!

PR #708, opened 2021-05-30
add len() to Stealer

### LINKED ISSUE #572: Expose length in dequeue Stealer
Would it be possible to expose [this](https://github.com/crossbeam-rs/crossbeam/blob/7cc83772635bab80b28daadfd0a3d911554884be/crossbeam-deque/src/deque.rs#L703) length as part of the public API? I want to experiment with different work-stealing policies in a thread-pool (e.g. read length from 2 Stealers and steal from the longer one) in order to avoid "steal cascades", if you will.

I can do a PR for this, if you think it's generally sound.

I'm also wondering if it would make sense to expose a `steal_with_threshold` API where you can specify a minimum queue length to steal from, i.e. changing `len <= 0` to `len <= min`. I can of course emulate this with the length API described above, but I would do the atomic reads twice, in that case.

dinfuehr:
Stealer only has the method `is_empty()` at the moment, but it would also be useful to check the size of the deque to e.g. steal from the longest deque.

Closes #572

### DIFF
--- a/crossbeam-deque/src/deque.rs
+++ b/crossbeam-deque/src/deque.rs
@@ -588,6 +588,27 @@ impl<T> Stealer<T> {
         b.wrapping_sub(f) <= 0
     }
 
+    /// Returns the number of tasks in the deque.
+    ///
+    /// ```
+    /// use crossbeam_deque::Worker;
+    ///
+    /// let w = Worker::new_lifo();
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks!

bors r+

PR #709, opened 2021-05-31
Add Atomic::const_null

taiki-e:
Inspired by parking_lot's [`const_mutex`](https://docs.rs/parking_lot/0.11.1/parking_lot/fn.const_mutex.html) and [`Mutex::const_new`](https://docs.rs/lock_api/0.4.2/lock_api/struct.Mutex.html#method.const_new).

This allows creating a null atomic pointer in a constant context on stable Rust.

The `nightly` feature is still needed to create a null atomic pointer to `?Sized + Pointable` types implemented by the user in a const context, but `Atomic::const_null` will probably be enough for many users.

### DIFF
--- a/crossbeam-epoch/src/atomic.rs
+++ b/crossbeam-epoch/src/atomic.rs
@@ -309,6 +309,47 @@ impl<T> Atomic<T> {
     pub fn new(init: T) -> Atomic<T> {
         Self::init(init)
     }
+
+    /// Returns a new null atomic pointer.
+    ///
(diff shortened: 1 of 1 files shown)


taiki-e:
Hmm, if we support both `Atomic<T>::const_null` and `Atomic<[MaybeUninit<T>]>::const_null`, the user need to use the fully-qualified syntax to disambiguate. (like, `Atomic::<u8>::const_null()`, `Atomic::<[MaybeUninit<_>]>::const_null()`)

taiki-e:
closing in favor of #797.

PR #713, opened 2021-06-17
avoid bounds check in Waker::try_select method

qy3u:
Use iter instead of index to avoid bounds check.


taiki-e (approved):
Thanks!

bors r+

PR #715, opened 2021-06-21
remove unnecessary 'move' keyword

qy3u:
The newer 'f' only captures a mut reference of the former 'f' to make itself becomes a FnMut.So, the 'move' keyword here seems unnecessary.


taiki-e (approved):
Thanks.

bors r+

PR #718, opened 2021-07-13
added get_or_insert_with function

snow01:
get_or_insert_with function allows lazy creation of default value. Default values may be heavy objects, lazy creation doesn't create un-necessary objects when key is already there in the data structure.

### DIFF
--- a/crossbeam-skiplist/src/base.rs
+++ b/crossbeam-skiplist/src/base.rs
@@ -471,6 +471,11 @@ where
 
     /// Finds an entry with the specified key, or inserts a new `key`-`value` pair if none exist.
     pub fn get_or_insert(&self, key: K, value: V, guard: &Guard) -> RefEntry<'_, K, V> {
+        self.insert_internal(key, || value, false, guard)
+    }
+
+    /// Finds an entry with the specified key, or inserts a new `key`-`value` pair if none exist, where value is calculated with a function.
+    pub fn get_or_insert_with<F>(&self, key: K, value: F, guard: &Guard) -> RefEntry<'_, K, V> where F: FnOnce() -> V {
         self.insert_internal(key, value, false, guard)
     }
 
@@ -831,13 +836,13 @@ where
     /// Inserts an entry with the specified `key` and `value`.
     ///
     /// If `replace` is `true`, then any existing entry with this key will first be removed.
-    fn insert_internal(
+    fn insert_internal<F>(
         &self,
         key: K,
-        value: V,
+        value: F,
         replace: bool,
         guard: &Guard,
-    ) -> RefEntry<'_, K, V> {
+    ) -> RefEntry<'_, K, V> where F: FnOnce() -> V {
         self.check_guard(guard);
 
         unsafe {
@@ -886,7 +891,7 @@ where
 
                 // Write the key and the value into the node.
                 ptr::write(&mut (*n).key, key);
-                ptr::write(&mut (*n).value, value);
+                ptr::write(&mut (*n).value, value());
 
                 (Shared::<Node<K, V>>::from(n as *const _), &*n)
             };
@@ -1061,7 +1066,7 @@ where
     /// If there is an existing entry with this key, it will be removed before inserting the new
     /// one.
     pub fn insert(&self, key: K, value: V, guard: &Guard) -> RefEntry<'_, K, V> {
-        self.insert_internal(key, value, true, guard)
+        self.insert_internal(key, || value, true, guard)
     }
 
     /// Removes an entry with the specified `key` from the map and returns it.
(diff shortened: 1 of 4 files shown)


snow01:
I have added two test cases - (a) for valid case (b) where closure panics. Please check if this is what you are looking at.

As another note, 
I was looking to implement update_or_insert_with() - which takes two closures, 
- one for compute and update value, if key is present.
- and another to insert new value, if key is absent

Looks like there is no way to atomically get value and apply updates on it. I could alternatively create new node with updated value, but to avoid value copying I was expecting value itself, rather than its reference. Pls suggest if you have some references on it.

Thanks

snow01:
@taiki-e - pls update here.

crossbeam-skiplist/tests/base.rs
```diff
@@ -431,6 +431,43 @@ fn get_or_insert() {
...
+    s.insert(1, 1, guard);
+    s.insert(4, 4, guard);
+    s.insert(2, 2, guard);
+
+    assert_eq!(*s.get(&4, guard).unwrap().value(), 4);
+    assert_eq!(*s.insert(4, 40, guard).value(), 40);
+    assert_eq!(*s.get(&4, guard).unwrap().value(), 40);
+
+    assert_eq!(*s.get_or_insert_with(4, || panic!(), guard).value(), 40);
+    assert_eq!(*s.get(&4, guard).unwrap().value(), 40);
```
taiki-e, line 468:
```suggestion
    use std::panic;

    let s = SkipList::new(epoch::default_collector().clone());
    let res = panic::catch_unwind(panic::AssertUnwindSafe(|| {
        let guard = &epoch::pin();
        s.get_or_insert_with(4, || panic!(), guard);
    }));
    assert!(res.is_err());
    assert!(s.is_empty());
    let guard = &epoch::pin();
    assert_eq!(*s.get_or_insert_with(4, || 40, guard).value(), 40);
```

To check the behavior in case of panic, we need to actually call the closure.

crossbeam-skiplist/src/base.rs
```diff
@@ -882,6 +893,9 @@ where
                 // The reference count is initially two to account for:
                 // 1. The entry that will be returned.
                 // 2. The link at the level 0 of the tower.
+
+                // create value before creating node, so extra allocation doesn't happen if value() function panics
+                let value = value();
```
taiki-e, line 898: Could you move these two lines to before `let height = self.random_height();` and its comment?

taiki-e:
Also, it would make sense to add the same method to SkipMap and SkipSet. SkipList is a low-level API, and most users use SkipMap and SkipSet.

crossbeam-skiplist/src/base.rs
```diff
@@ -471,6 +471,14 @@ where
 
     /// Finds an entry with the specified key, or inserts a new `key`-`value` pair if none exist.
     pub fn get_or_insert(&self, key: K, value: V, guard: &Guard) -> RefEntry<'_, K, V> {
+        self.insert_internal(key, || value, false, guard)
+    }
+
+    /// Finds an entry with the specified key, or inserts a new `key`-`value` pair if none exist, where value is calculated with a function.
```
taiki-e, line 477: It would also be preferable to mention that if a value is inserted into the same key from another thread while the closure is running, the result of the closure will be discarded.

taiki-e: Test case for this behavior:
```rust
#[test]
fn t() {
    let s = Arc::new(SkipList::new(epoch::default_collector().clone()));
    let s2 = s.clone();
    let called = Arc::new(Mutex::new(false));
    let called2 = called.clone();
    let handle = std::thread::spawn(move || {
        let guard = &epoch::pin();
        assert_eq!(
            *s2.get_or_insert_with(
                7,
                || {
                    *called2.lock().unwrap() = true;
                    std::thread::sleep(std::time::Duration::from_secs(4));
                    70
                },
                guard,
            )
            .value(),
            700
        );
    });
    std::thread::sleep(std::time::Duration::from_secs(2));
    let guard = &epoch::pin();
    assert_eq!(*s.get_or_insert(7, 700, guard).value(), 700);
    handle.join().unwrap();
    assert!(*called.lock().unwrap());
}
```

snow01: Sure, taken care.

crossbeam-skiplist/src/base.rs
snow01: Taken care.

snow01:
> Also, it would make sense to add the same method to SkipMap and SkipSet. SkipList is a low-level API, and most users use SkipMap and SkipSet.

Sure, will do.

snow01:
@taiki-e - added corresponding method and test for SkipMap. For SkipSet, as value is always () it may not make sense.

Please review.

Thanks

taiki-e (approved):
LGTM. Thanks @snow01!

bors r+

PR #722, opened 2021-07-25
Simplify `AtomicCell` `LOCKS`

Daniel-Bloom-dfinity:


taiki-e:
I would prefer to use a way that does not add dependencies. We can write as follows:
https://github.com/crossbeam-rs/crossbeam/blob/4acfd0b7799a1a71821019235d23881ee2ad9670/crossbeam-utils/src/atomic/atomic_cell.rs#L641-L648

taiki-e (approved):
Thanks!

bors r+

PR #735, opened 2021-08-21
Fix memory leaks in pop_front and pop_back

topecongiro:
`SkipList` may leak memory when calling `pop_front` (or `pop_back`) concurrently. This is because we forget to decrement the reference count of `RefEntry` when the deletion of an element fails in the `pop_front`/`pop_back`.

This PR fixes it by calling `release` when the removal fails.


taiki-e (approved):
Thanks!

bors r+

PR #736, opened 2021-08-21
Fix memory leak in iterator of skiplist

topecongiro:
SkipMap's iterator may leak memory when it reads the value that gets removed from the SkipMap later:

```rust
{
    let map: SkipMap<i32, i32> = iter::once((1, 1)).collect();
    let mut iter = map.iter();
    let e = iter.next_back();
    let e = iter.next();
    map.remove(&1);
} // iter leaks memory
```

The leak happens because the implementation of next and next_back does not call release when replacing head and tail to None (see the diff).

Note that the Drop impl of SkipMap will reclaim all the values in the skiplist when it's dropped; thus, only values that are removed before dropping the skiplist get leaked.

This PR fixed the leak by decreasing the reference count when needed.

### DIFF
--- a/crossbeam-skiplist/src/base.rs
+++ b/crossbeam-skiplist/src/base.rs
@@ -1725,8 +1725,12 @@ where
             }
         }
         if finished {
-            self.head = None;
-            self.tail = None;
+            if let Some(e) = mem::replace(&mut self.head, None) {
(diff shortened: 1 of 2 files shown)


topecongiro:
Closing in favor of #738.

PR #741, opened 2021-09-01
Fix skiplist Range API

topecongiro:
This PR fixes two issues in skiplist Range API.

1. `next_back` does not work as expected
2. Range can leak memory

We fix the first one by using the correct bound when searching for the upper/lower bound (3406d84) and the second one by the same approach used in #738 (a7caaf1).


taiki-e (approved):
Thanks!

bors r+

PR #750, opened 2021-10-29
Add `Receiver::new_sender` to channels

oowekyala:
This change gives `channel::Receiver`s the ability to spawn new senders at will. 

Currently, Receivers are disconnected as soon as the Sender count reaches zero, as new Senders can only be created by cloning an existing one. Disconnected receivers do not block on `recv` and such, instead they just fail immediately.

This didn't quite fit my use case, where I need to have the ability to create new Senders when I want to, and also get the non-blocking behavior if there is no Sender alive. With the current API, I have to keep a prototype Sender which is always alive and which I can clone whenever I want. The problem is that since the prototype is always alive, the channel never gets disconnected. That means calls to `recv` stay blocking, even when we know that no message will ever be sent through the Sender prototype, because its only purpose is to produce new Senders.

This change adds support for creating new Sender instances from the Receiver directly. That means, I don't need to keep a Sender prototype anymore, so my Receiver is non-blocking when I know there is no real Sender alive. But even after the channel was disconnected, you can "reconnect" it by spawning new senders.

~~Obviously this method cannot appear in the API of channel flavours like `tick` or `never`, since the point of those is to expose only a receiver. This means, we need two different types for `Receiver`. I put that new API in a separate submodule so that it's easy to import and document. That submodule only supports creating unbounded channels for now, as that was my use case. If you're interested we may be able to add support for bounded and zero channels, but there is AFAIK no risk in keeping an incomplete API here.~~ The new method `Receiver::new_sender` is usable on bounded and unbounded channels; it panics on `tick`, `at` and `never` channels.

Please let me know if there is anything you need from me to get this merged faster, be it documentation, tests, etc. 
Cheers :)

### DIFF
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -1561,3 +1561,78 @@ pub(crate) unsafe fn read<T>(r: &Receiver<T>, token: &mut Token) -> Result<T, ()
         }
     }
 }
+
+/// Channel flavors that are reconnectable, ie, may survive
+/// without senders.
+///
+pub mod reconnectable {
+    // note: we need separate types because the At/Tick/Never
+    // channel types do not support this, but they all use the
+    // same exact Receiver type as array and list flavours.
+
+    // The intended usage of this module is just to replace any `use crossbeam_channel::...`
+    // with `use crossbeam_channel::reconnectable::...` and have everything work out of the box.
+
+    // note: these re-exports omit
+    // at, tick, never: cannot be supported
+    // bounded: may be added later in this module
+    use std::ops::Deref;
+
+    use super::{Receiver as NormalReceiver, ReceiverFlavor, SenderFlavor};
+    pub use crate::{
+        Sender,
+        channel::{IntoIter, Iter, TryIter},
+        err::{
+            ReadyTimeoutError, RecvError, RecvTimeoutError, SelectTimeoutError, SendError,
+            SendTimeoutError, TryReadyError, TryRecvError, TrySelectError, TrySendError,
+        },
+        select::{Select, SelectedOperation},
+    };
+
+    /// An [unbounded](super::unbounded) channel whose receiver
+    /// also is reconnectable.
+    pub fn unbounded<T>() -> (Sender<T>, Receiver<T>) {
+        let (s, r) = super::unbounded();
+        (s, Receiver(r))
+    }
+
+    /// A [receiver](super::Receiver) that can survive periods
+    /// of time where no [Sender]s are alive. New senders may
+    /// be created from the receiver directly ([Self::new_sender]).
+    /// The channel is only deallocated when the last receiver dies.
+    /// If there are live senders at that point, they start producing
+    /// [SendError]s as usual.
+    #[derive(Debug)]
+    pub struct Receiver<T>(NormalReceiver<T>);
+
+    impl<T> Receiver<T> {
+        /// Returns a new sender that communicates with this
+        /// receiver.
+        pub fn new_sender(&self) -> Sender<T> {
+            match &self.0.flavor {
+                ReceiverFlavor::Array(chan) => Sender {
+                    flavor: SenderFlavor::Array(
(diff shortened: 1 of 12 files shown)


crossbeam-channel/src/channel.rs
```diff
@@ -1506,3 +1507,63 @@ pub(crate) unsafe fn read<T>(r: &Receiver<T>, token: &mut Token) -> Result<T, ()
...
+pub mod reconnectable {
+    // note: we need separate types because the At/Tick/Never
+    // channel types do not support this, but they all use the
+    // same exact Receiver type as array and list flavours.
+
+    // The intended usage of this module is just to replace any `use crossbeam_channel::...`
+    // with `use crossbeam_channel::reconnectable::...` and have everything work out of the box.
+
+    use super::*;
+    pub use crate::*;
```
taiki-e, line 1523: Please expand this glob re-export.

Also, re-exporting `bounded` here will make the future reconnectable bounded channel addition a breaking change.

crossbeam-channel/src/channel.rs
```diff
@@ -1506,3 +1507,63 @@ pub(crate) unsafe fn read<T>(r: &Receiver<T>, token: &mut Token) -> Result<T, ()
...
+    #[derive(Debug)]
+    pub struct Receiver<T>(NormalReceiver<T>);
+
+    impl<T> Receiver<T> {
+        /// Returns a new sender that communicates with this
+        /// receiver.
+        pub fn new_sender(&self) -> Sender<T> {
+            match &self.0.flavor {
+                ReceiverFlavor::Array(chan) => Sender {
+                    flavor: SenderFlavor::Array(chan.new_sender(|_| unreachable!())),
```
taiki-e, line 1549: > unreachable!()

Does this mean that it is unimplemented?

oowekyala: Yes it is unimplemented, but given the current api to create reconnectable receivers, it is also unreachable and shouldn't cause problems. Maybe `todo!` is more appropriate?

crossbeam-channel/src/channel.rs
oowekyala: Thanks for spotting this!

ryoqun:
hey, this api is interesting to me as well.

> Obviously this method cannot appear in the API of channel flavours like `tick` or `never`, since the point of those is to expose only a receiver. This means, we need two different types for `Receiver`. I put that new API in a separate submodule so that it's easy to import and document. That submodule only supports creating unbounded channels for now, as that was my use case. If you're interested we may be able to add support for bounded and zero channels, but there is AFAIK no risk in keeping an incomplete API here.

However, I didn't particularly like this submodule arrangement... How about making `new_sender` returning some dummy sender for `tick` and `never`? maybe this will be called `full()`?

ryoqun:
ping @oowekyala @taiki-e regarding my api idea: https://github.com/crossbeam-rs/crossbeam/pull/750#issuecomment-1876415286

crossbeam-channel/src/channel.rs
```diff
@@ -1506,3 +1506,76 @@ pub(crate) unsafe fn read<T>(r: &Receiver<T>, token: &mut Token) -> Result<T, ()
...
+                },
+                ReceiverFlavor::List(chan) => Sender {
+                    flavor: SenderFlavor::List(chan.new_sender(|c| c.reconnect_senders())),
+                },
+                ReceiverFlavor::Zero(chan) => Sender {
+                    flavor: SenderFlavor::Zero(
+                        chan.new_sender(|_| unimplemented!("but unreachable")),
+                    ),
+                },
+                _ => unreachable!("This type cannot be built with at/never/tick"),
```
taiki-e: The code calling `new_sender` with `tick/at/never` is broken anyway, so I think it's okay to use  `track_caller` and simply panic without separating the module. (I'm not in favor of releasing without implementing list and zero.)

oowekyala: I implemented your suggestion and removed the separate module. I also implemented this feature for Array and Zero flavors

PR #751, opened 2021-11-03
crossbeam-utils: check only the architecture, not the whole target string

kanavin:
There can be custom targets in use, and it's not possible to make a list
of them; for the check only the first item in the target string is actually
relevant (the architecture of the target, which is generally consistent
between standard and custom targets).

Signed-off-by: Alexander Kanavin <alex@linutronix.de>

### DIFF
--- a/crossbeam-epoch/build.rs
+++ b/crossbeam-epoch/build.rs
@@ -30,11 +30,16 @@ fn main() {
         }
     };
 
+    let target_arch = target.split('-').next().unwrap_or_default();
+
     // Note that this is `no_*`, not `has_*`. This allows treating
     // `cfg(target_has_atomic = "ptr")` as true when the build script doesn't
     // run. This is needed for compatibility with non-cargo build systems that
     // don't run the build script.
-    if NO_ATOMIC_CAS.contains(&&*target) {
+    if NO_ATOMIC_CAS
+        .iter()
+        .any(|t| t.split('-').next().unwrap_or_default() == target_arch)
+    {
         println!("cargo:rustc-cfg=crossbeam_no_atomic_cas");
     }
 
--- a/crossbeam-queue/build.rs
+++ b/crossbeam-queue/build.rs
@@ -30,11 +30,16 @@ fn main() {
         }
     };
 
+    let target_arch = target.split('-').next().unwrap_or_default();
+
     // Note that this is `no_*`, not `has_*`. This allows treating
     // `cfg(target_has_atomic = "ptr")` as true when the build script doesn't
     // run. This is needed for compatibility with non-cargo build systems that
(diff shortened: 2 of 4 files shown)


taiki-e (changes requested):
In some cases, the kinds of atomic operations supported depend on the OS. For example:

- x86_64-*: cmpxchg16b is available for all x86_64 macOS devices, so Atomic128 is available for x86_64 macOS but not for x86_64 Linux.
- riscv32imc-*: [riscv32imc-esp-espidf has atomic, but riscv32imc-unknown-none-elf does not.](https://github.com/rust-lang/rust/blob/473eaa42e9365c47d129f72693b5d163a20cf369/compiler/rustc_target/src/spec/riscv32imc_esp_espidf.rs#L20-L26)

Currently, the recommended way to build for custom targets is to use rustflags to pass the [cfgs](https://github.com/crossbeam-rs/crossbeam/blob/dba57a2c65e863209dac483c9fd9c86927a53b7e/crossbeam-utils/build.rs#L7-L26).
I would accept a patch to improve the situation, but we need to handle the edge cases mentioned above (in a way that does not manually maintain a list of such targets, whenever possible).

kanavin:
> Currently, the recommended way to build for custom targets is to use rustflags to pass the [cfgs](https://github.com/crossbeam-rs/crossbeam/blob/dba57a2c65e863209dac483c9fd9c86927a53b7e/crossbeam-utils/build.rs#L7-L26). I would accept a patch to improve the situation, but we need to handle the edge cases mentioned above (in a way that does not manually maintain a list of such targets, whenever possible).


Thanks. The context is that this patch comes from the efforts to cross-compile librsvg in the Yocto project, where crossbeam is included into the librsvg tarball as a vendored library. For various reasons Yocto cannot use the standard rust targets (e.g. the compiler binary is hardcoded in them to 'cc' which is something we cannot use). 

I guess I need to find out how to convince librsvg's build system to pass those custom flags to crossbeam, instead of running build.rs - if you can take the librsvg tarball and check what it does, I'd appreciate:
https://download.gnome.org/sources/librsvg/2.52/

Just for reference, you can see our configuration ('recipe' in yocto parlance) for librsvg here:
http://git.yoctoproject.org/cgit/cgit.cgi/poky-contrib/tree/meta/recipes-gnome/librsvg/librsvg_2.52.3.bb?h=akanavin/package-version-updates

kanavin:
So yes, it's actually easier to just pass the needed flags via RUSTFLAGS on architectures where it's needed:
```
RUSTFLAGS:append:mips = " --cfg crossbeam_no_atomic_64"
RUSTFLAGS:append:powerpc = " --cfg crossbeam_no_atomic_64"
```
Thanks!

PR #755, opened 2021-12-02
Remove SC fence in try_advance

tomtomjhj:
Since an SC fence is issued when obtaining the guard, try_advance
doesn't need to issue another fence if it uses the guard's local epoch
for advancing the global epoch.

`cargo bench` results:
* master
  ```
  test multi_alloc_defer_free  ... bench:   3,007,952 ns/iter (+/- 128,911)
  test multi_defer             ... bench:   1,595,657 ns/iter (+/- 24,119)
  test single_alloc_defer_free ... bench:          41 ns/iter (+/- 0)
  test single_defer            ... bench:          12 ns/iter (+/- 0)
  test multi_flush             ... bench:  17,557,440 ns/iter (+/- 344,195)
  test single_flush            ... bench:          76 ns/iter (+/- 0)
  test multi_pin               ... bench:   3,376,356 ns/iter (+/- 93,846)
  test single_pin              ... bench:           4 ns/iter (+/- 0)
  ```

* this patch
  <pre>
  test multi_alloc_defer_free  ... bench:   2,990,410 ns/iter (+/- 106,550)
  test multi_defer             ... bench:   1,593,113 ns/iter (+/- 20,050)
  test single_alloc_defer_free ... bench:          40 ns/iter (+/- 0)
  test single_defer            ... bench:          12 ns/iter (+/- 0)
  test multi_flush             ... bench:  <strong>15,167,083</strong> ns/iter (+/- 690,042)
  test single_flush            ... bench:          <strong>42</strong> ns/iter (+/- 2)
  test multi_pin               ... bench:   3,300,440 ns/iter (+/- 48,145)
  test single_pin              ... bench:           4 ns/iter (+/- 0)
  </pre>


jeehoonkang:
@tomtomjhj would you please look at the CI failures?

tomtomjhj:
The first one is a rust edition issue (old rustc doesn't support 2021 edition?) and the other is clippy lint for crossbeam-utils. I think it's better to address these in another PR.

taiki-e:
CI failure has been fixed in #758.

tomtomjhj:
[`pin_holds_advance`](https://github.com/crossbeam-rs/crossbeam/blob/858bc6e03e1a688368262a4f43e8e35db69ac522/crossbeam-epoch/src/collector.rs#L199) test fails when I add artificial delay here:
https://github.com/crossbeam-rs/crossbeam/blob/858bc6e03e1a688368262a4f43e8e35db69ac522/crossbeam-epoch/src/internal.rs#L466-L468

This happens when the global epoch decreases, which is in contrary to this comment: https://github.com/crossbeam-rs/crossbeam/blob/858bc6e03e1a688368262a4f43e8e35db69ac522/crossbeam-epoch/src/internal.rs#L338-L343

So crossbeam is relying on the additional global epoch load and fence in `try_advance` for the monotonicity of global epoch. This is an important assumption in [Jeehoon's proof](https://github.com/crossbeam-rs/rfcs/blob/master/text/2017-07-23-relaxed-memory.md). However, it seems that this point is not proved in that document, and I couldn't figure out why that works.

A relatively simple solution for enforcing monotonicity is to update the global epoch with a (relaxed) CAS instead of store.

PR #772, opened 2022-01-19
Add IntoIter implementations for ArrayQueue & SegQueue

PatrickNorton:
As it says in the title—this is a feature I've wanted a few times.

This adds the following implementations:
```rust
impl<T> IntoIterator for ArrayQueue<T>;

impl<T> IntoIterator for SegQueue<T>;
```


PatrickNorton:
Am I correct in figuring that the clippy fail isn't due to this? It's in a separate file & a relatively new nursery lint, and nothing else seems to be failing.

taiki-e (approved):
Thanks!

bors r+

Issue #433, opened 2019-10-16
Use AtomicU64 for epochs in 32-bit architectures


jeehoonkang:
@stjepang [said](https://github.com/crossbeam-rs/crossbeam/pull/359#issuecomment-485958500):

> Just so we don't forget, in all the other places where we have atomic timestamps, we should use `AtomicU64` (I believe we can't use wide sequences like in this PR, unfortunately):
>
> - epoch representation
> - head/tail index in deque, channel, and queues
>
> But that can come later in another PR. It will also require Rust 1.34 minimum.

jeehoonkang:
Since MSRV = 1.36 now, we can use `AtomicU64` now: https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html

taiki-e:
Note that some 32-bit architectures do not support `AtomicU64`: https://github.com/tokio-rs/tokio/pull/1421#issuecomment-527911780

jeehoonkang:
Very sad. Do you think it's worth it to use AtomicU64 for those platforms that support it? My opinion is yes, because, e.g., major 32-bit architectures support AtomicU64.

taiki-e:
I think it's good to use it when `AtomicU64` is available, but we have to decide what to do if `AtomicU64` is not available.
We can probably use `AtomicCell<u64>` for this, but performance may be degraded on platforms that `AtomicU64` is not available.

pkubaj:
Note that there are already issues with building this crate on 32-bit powerpc:
```
error[E0432]: unresolved imports `core::sync::atomic::AtomicI64`, `core::sync::atomic::AtomicU64`
  --> /wrkdirs/usr/ports/benchmarks/inferno/work/inferno-0.10.6/cargo-crates/crossbeam-utils-0.8.5/src/lib.rs:79:49
   |
79 |             pub(crate) use core::sync::atomic::{AtomicI64, AtomicU64};
   |                                                 ^^^^^^^^^  ^^^^^^^^^ no `AtomicU64` in `sync::atomic`
   |                                                 |
   |                                                 no `AtomicI64` in `sync::atomic`
   |
help: a similar name exists in the module
   |
79 |             pub(crate) use core::sync::atomic::{AtomicI8, AtomicU64};
   |                                                 ^^^^^^^^
help: a similar name exists in the module
   |
79 |             pub(crate) use core::sync::atomic::{AtomicI64, AtomicU8};
   |                                                            ^^^^^^^^

[...]

PR #777, opened 2022-01-23
Use AtomicU64 for head/tail index in deque, channel, and queues

taiki-e:
Part of #433.


crossbeam-channel/src/utils.rs
```diff
@@ -61,6 +61,46 @@ pub(crate) fn sleep_until(deadline: Option<Instant>) {
...
+#[cfg(crossbeam_no_atomic_64)]
+impl AtomicU64 {
+    pub(crate) const fn new(v: u64) -> Self {
+        Self {
+            inner: crossbeam_utils::atomic::AtomicCell::new(v),
+        }
+    }
+    pub(crate) fn load(&self, _order: Ordering) -> u64 {
+        self.inner.load()
+    }
```
taiki-e, line 83: Is it preferable to define this in crossbeam_utils::atomic?

taiki-e:
Closed in favor of https://github.com/crossbeam-rs/crossbeam/pull/1321, for now.

Issue #749, opened 2021-10-27
Old crossbeam-channel could return a crazy length because of integer overflow


NingLin-P:
In `crossbeam-channel` 0.3.x and 0.4.x (also happen to `crossbeam_queue::SegQueue`), when computing the length of the channel, if the indices fall onto the block end, they will be fixed up by:

https://github.com/crossbeam-rs/crossbeam/blob/48b13dc43cf19e0544d03d16f4b3da3d8b23f076/crossbeam-channel/src/flavors/list.rs#L511-L518

Although the indices have been rotated with `wrapping_sub` previously, if both the head and the tail index are equal to `BLOCK_CAP` here, then `tail -= LAP` will overflow and wrap into `usize::MAX` when build in release mode (or panic when build in debug mode).

https://github.com/crossbeam-rs/crossbeam/commit/1df3ee7b09feabf71d85bac46e36f1eef6257bca has been changed the way to fix up indices, so this problem won't happen on 0.5.x. We may need to cherry-pick this commit to 0.3.x and 0.4.x branches.

taiki-e:
Closing -- Given that the first version of 0.5 that fixed this issue was released over a year ago, the major version is 0, and that there have been few breaking changes between 0.4 and 0.5, there is no motivation to continue maintaining 0.4.

PR #781, opened 2022-02-05
Fix unsoundness of AtomicCell<*64> arithmetics on 32-bit targets that support Atomic*64

taiki-e:
The alignment of u64/i64 on a 32-bit target can be smaller than AtomicU64/AtomicI64.

32-bit targets without Atomic*64 (https://github.com/crossbeam-rs/crossbeam/pull/767) and 64-bit targets are not affected by this issue.

### DIFF
--- a/crossbeam-utils/src/atomic/atomic_cell.rs
+++ b/crossbeam-utils/src/atomic/atomic_cell.rs
@@ -465,8 +465,24 @@ macro_rules! impl_arithmetic {
             /// ```
             #[inline]
             pub fn fetch_add(&self, val: $t) -> $t {
-                let a = unsafe { &*(self.value.get() as *const $atomic) };
-                a.fetch_add(val, Ordering::AcqRel)
+                if can_transmute::<$t, $atomic>() {
+                    let a = unsafe { &*(self.value.get() as *const $atomic) };
+                    a.fetch_add(val, Ordering::AcqRel)
+                } else {
+                    #[cfg(crossbeam_loom)]
+                    {
+                        let _ = val;
+                        unimplemented!("loom does not support non-atomic atomic ops");
+                    }
+                    #[cfg(not(crossbeam_loom))]
+                    {
+                        let _guard = lock(self.value.get() as usize).write();
+                        let value = unsafe { &mut *(self.value.get()) };
+                        let old = *value;
+                        *value = value.wrapping_add(val);
+                        old
+                    }
+                }
             }
 
             /// Decrements the current value by `val` and returns the previous value.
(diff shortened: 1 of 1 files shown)


jlebon:
Were the previous releases yanked because of this? It'd be good to note that in the `CHANGELOG` under each version yanked for folks looking for more context. Thanks!

taiki-e:
Originally I was going to write a security advisory on this, but I was busy with work and it got delayed. (I requested CVE today, so it should be published in the next few days)

That said, I agree that mentioning the yanked version is a good idea ([I usually do it on my projects](https://github.com/taiki-e/pin-project/blob/main/CHANGELOG.md#101---2020-10-15)), so I did that in https://github.com/crossbeam-rs/crossbeam/pull/788.

taiki-e:
Advisory published as https://github.com/crossbeam-rs/crossbeam/security/advisories/GHSA-qc84-gqf4-9926.

schopin-pro:
Hi!

Is there a reason why the 0.7.* versions haven't been yanked? I thought it was because they weren't affected by the issue but looking a bit more closely to the changelog for 0.8.0 I don't see why it would be affected and not the 0.7.* versions.

When browsing the reverse-dependency list on crates.io, there are still quite a few crates that specify a ^0.7.0 dependency. Rust 1.58.1, for instance, had both 0.8.5 and 0.7.2 in its vendored dependency tree.

taiki-e:
@schopin-pro

> I thought it was because they weren't affected by the issue but looking a bit more closely to the changelog for 0.8.0 I don't see why it would be affected and not the 0.7.* versions.

As "Affected versions" in [the advisory](https://github.com/crossbeam-rs/crossbeam/security/advisories/GHSA-qc84-gqf4-9926) says, 0.7.x releases are also affected. However, in 0.7.0 and older releases, the affected functionality requires the nightly feature. So the only release that is not yanked and still considered affected is 0.7.2.

For the same reason as https://github.com/crossbeam-rs/crossbeam/issues/749#issuecomment-959591158, we no longer maintain 0.7.x releases. So no backport to 0.7.x has been made. However, if someone sends a backport to the 0.7 branch and also fixes the CI-related issues (which we expect will probably occur), then we can accept it and release a new 0.7.x release.

> When browsing the reverse-dependency list on crates.io, there are still quite a few crates that specify a ^0.7.0 dependency.

As far as I know, many of them are crates that are not maintained, or use crossbeam-utils as a dev-dependency.

> Rust 1.58.1, for instance, had both 0.8.5 and 0.7.2 in its vendored dependency tree.

That has already been fixed: https://github.com/rust-lang/rust/commit/f3b8812f2473b49453e79c3f57acf4fa4c2ca227
Also, the rustc's case (rustc-rayon) does not seem to be affected by this issue.

taiki-e:
> I feel like the 0.7.2 should have been yanked as well, so that people that are indirect consumers lagging a bit behind would have the yanked warning.

Unfortunately, yanking does not actually work that way.
If we yank 0.7.2, a crate that requires crossbeam-utils ^0.7.2 as a dependency will fail to resolve dependencies if Cargo.lock is not yet present.

pinkforest:
Hey @taiki-e could you release the advisory as Public Domain so we can publish this in RustSec ?

https://github.com/rustsec/advisory-db/issues/1203

Or please feel free to send PR to https://github.com/rustsec/advisory-db

Thanks!

taiki-e:
Sigh. I thought those were in the public domain.

I re-licensed all advisories in this repository as the public domain.

PR #787, opened 2022-02-09
Do not provide AtomicCell when cfg(crossbeam_loom) is enabled

taiki-e:
We cannot provide AtomicCell under `cfg(crossbeam_loom)` because loom's atomic types have a different in-memory representation than the underlying type. (https://github.com/tokio-rs/loom/pull/253)

https://github.com/crossbeam-rs/crossbeam/blob/bd83e1f2328e90298fc8a2f535a25d94ece09152/crossbeam-utils/src/atomic/atomic_cell.rs#L35-L42

fyi @jonhoo

### DIFF
--- a/crossbeam-utils/src/atomic/mod.rs
+++ b/crossbeam-utils/src/atomic/mod.rs
@@ -24,9 +24,14 @@ cfg_if::cfg_if! {
 }
 
 #[cfg(not(crossbeam_no_atomic_cas))]
+// We cannot provide AtomicCell under cfg(crossbeam_loom) because loom's atomic
+// types have a different in-memory representation than the underlying type.
+// TODO: The latest loom supports fences, so fallback using seqlock may be available.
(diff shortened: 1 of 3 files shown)


crossbeam-utils/src/atomic/mod.rs
```diff
@@ -24,9 +24,14 @@ cfg_if::cfg_if! {
 }
 
 #[cfg(not(crossbeam_no_atomic_cas))]
+// We cannot provide AtomicCell under cfg(crossbeam_loom) because loom's atomic
+// types have a different in-memory representation than the underlying type.
+// TODO: The latest loom supports fences, so fallback using seqlock may be available.
+#[cfg(not(crossbeam_loom))]
```
taiki-e, line 30: (Well, this may be difficult because seqlock has [another issue](http://www.hpl.hp.com/techreports/2012/HPL-2012-68.pdf).)

PR #802, opened 2022-03-17
channel: Do not call current_thread_id inside iteration

taiki-e:
This should result in the same behavior as before 10009bd.

This works around an upstream bug related to TLS access on aarch64 linux, which was [reported in firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1757571).

### DIFF
--- a/crossbeam-channel/src/waker.rs
+++ b/crossbeam-channel/src/waker.rs
@@ -77,26 +77,32 @@ impl Waker {
     /// Attempts to find another thread's entry, select the operation, and wake it up.
     #[inline]
     pub(crate) fn try_select(&mut self) -> Option<Entry> {
-        self.selectors
-            .iter()
-            .position(|selector| {
-                // Does the entry belong to a different thread?
-                selector.cx.thread_id() != current_thread_id()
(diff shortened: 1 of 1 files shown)


glandium:
> when LTO is enabled,

FWIW, since my comment about that, I actually confirmed that LTO is _not_ involved. It might be dependent on the rust version, though (not fully confirmed yet).

glandium:
FWIW, something else that works is changing the `current_thread_id` function to not have a thread-local cache.

taiki-e:
@glandium Thanks for investigating this!
If there is a problem with the inlining of TLS access, it may work fine on rustc before https://github.com/rust-lang/rust/commit/641d3b09f41b441f2c2618de32983ad3d13ea3f8 (1.54) and if LTO is disabled.

bors r+

taiki-e:
(Possibly related: https://github.com/rust-lang/rust/issues/48967)

glandium:
For the record, taking 0.5.4 works, and changing `try_select` to have `#[inline(never)]` breaks it again.

PR #804, opened 2022-03-17
Only call current_thread_id once in try_select

glandium:
For some unclear reason as of writing, calling current_thread_id on each
iteration makes Firefox block before displaying a window on aarch64
Linux.


cynecx:
This is a duplicate of https://github.com/crossbeam-rs/crossbeam/pull/802.

glandium:
@taiki-e Someone contacted you about https://bugzilla.mozilla.org/show_bug.cgi?id=1757571. This is where we are investigation-wise. Applying this patch fixes the issue. That would suggest some kind of race condition somewhere else, but I'm yet to find out the root cause of the problem. The change itself is good to apply anyways.

glandium:
Heh, I guess I should have checked pull requests.

Issue #803, opened 2022-03-17
error[E0658]: trait bounds other than `Sized` on const fn parameters are unstable


ArunGust:
Compile error

   Compiling crossbeam-epoch v0.9.8
error[E0658]: trait bounds other than `Sized` on const fn parameters are unstable
   --> C:\SYSTEM\RUST\CARGO\registry\src\github.com-1ecc6299db9ec823\crossbeam-epoch-0.9.8\src\atomic.rs:314:6
    |
314 | impl<T: ?Sized + Pointable> Atomic<T> {
    |      ^
...
346 |     pub const fn null() -> Atomic<T> {
    |     -------------------------------- function declared as const here
    |
    = note: see issue #93706 <https://github.com/rust-lang/rust/issues/93706> for more information
    = help: add `#![feature(const_fn_trait_bound)]` to the crate attributes to enable

For more information about this error, try `rustc --explain E0658`.
error: could not compile `crossbeam-epoch` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed

taiki-e:
This can be fixed by using nightly-2022-03-08 or later.

There is no guarantee that crossbeam can be compiled with the old nightly compiler, so I'm going to close this.

Aidiakapi:
This is incredibly annoying, because these nightly versions are currently having various components broken.

taiki-e:
Users of the old nightly toolchains can downgrade crossbeam-epoch.

```
cargo update -p crossbeam-epoch:0.9.8 --precise 0.9.7
```

PR #809, opened 2022-03-24
Update README.md of crossbeam epoch to have correct version requirement of rust compiler

TotalKrill:


taiki-e:
This is not correct. It is the *old nightly* compiler that is not supported; the *stable* compiler is supported in all versions since 1.36. See also https://github.com/crossbeam-rs/crossbeam/issues/803.

TotalKrill:
Well, It doesnt build on stable now, so I don't know how you can say that stable 1.36 is supported?

taiki-e:
Our CI (which runs daily) makes sure crossbeam-* crates can build on stable 1.36, latest stable (currently 1.59), and latest nightly): https://github.com/crossbeam-rs/crossbeam/actions/runs/2031722592

> It doesnt build on stable now

Hmm, that's odd. Can you provide the full error message and the version of rustc (`rustc -vV`)? Also, under what situations did you get the error? (CLI or editor, CI or local, native or WSL, etc.)

TotalKrill:
Never mind me, I was so sure I was running stable by default, turns out some script I have used changed to an older version of nightly by default. 

So basically, my system was wrongly configured and I was mistaken.

PR #811, opened 2022-04-06
channel: use `AtomicUsize::get_mut` in `flavors::{array,list}::Channel::drop`

sb64:
As access to `&mut self` in `Drop` impls gives exclusive access, it's
unnecessary to atomically load `self.head` and `self.tail` there.

### DIFF
--- a/crossbeam-channel/src/flavors/array.rs
+++ b/crossbeam-channel/src/flavors/array.rs
@@ -521,8 +521,11 @@ impl<T> Channel<T> {
 impl<T> Drop for Channel<T> {
     fn drop(&mut self) {
         // Get the index of the head.
-        let hix = *self.head.get_mut() & (self.mark_bit - 1);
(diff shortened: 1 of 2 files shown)


sb64:
I stumbled across this as I was reading the code for educational purposes

taiki-e:
Thanks! I'll merge this once the upstream bugfix (https://github.com/rust-lang/rust/pull/95621) which fixes the breakage on miri CI merged.

RalfJung:
You could temporarily remove the `-Zmiri-check-number-validity` to work around the MPSC bug.
(Sorry for this, I didn't realize how widely used that flag is so I wasn't too worried when I noticed MPSC breaking with it.)

taiki-e:
Thanks for the advice. This time I chose to wait for the patch to be merged since it had already been submitted that looked reasonable. It makes sense to temporarily remove the flag if the fix is going to take a long time.

PR #817, opened 2022-04-23
Replace lazy_static with once_cell

### LINKED ISSUE #623: Switch from lazy_static to once_cell
With tokio having switched from lazy_static to once_cell in 0.3, the dependencies on lazy_static are going way down in many of my projects. It would be nice to get rid of it completely in favor of once_cell. Would you be open to a PR?

taiki-e:
Closes #623


taiki-e:
Ah, lazy_static 1.5.0 is not actually released yet so the compatibility issue with Rust 1.36 will become later.

PR #824, opened 2022-05-08
Golang test chan

alygin:
Port of the [chan.go](https://github.com/golang/go/blob/master/test/ken/chan.go) test.

Part of #201


taiki-e (approved):
Thanks!

bors r+

Issue #825, opened 2022-05-09
Unreliable doc test for tick()


alygin:
There's a failure in one of the test runs in [this PR](https://github.com/crossbeam-rs/crossbeam/pull/824). I decided to file a separate issue for fixing this failure since it was not introduced by changes in the PR.

The [failed doc test](https://github.com/crossbeam-rs/crossbeam/blob/78d000b06d23bdcf1d98cb3765206ba0f3fec1f3/crossbeam-channel/src/channel.rs#L293-L318) depends on proper timing and has a 50-msec margin for several events.

It seems that the test failure was caused by `thread::sleep()` delay inacuracies that added up to more than 50 msec. There're four `thread::sleep()` calls in this test (explicit and hidden inside `recv()`). Each of them may introduce extra delay up to 16 msec on Windows, so the margin should be at least 64 msec instead of 50 to make the test reliable.

I've got a Windows machine where this test fails once in ~20 runs.

taiki-e:
https://github.com/crossbeam-rs/crossbeam/pull/826 merged

PR #826, opened 2022-05-09
Make doc test for tick channel more reliable

alygin:
Increase time margin for checking event times to make the doc test more reliable

ref #825


taiki-e (approved):
Thanks!

bors r+

PR #827, opened 2022-05-14
Remove unused macro rules

alygin:
The [new `unused_macro_rules` lint](https://github.com/rust-lang/rust/issues/73576) that recently landed in nightly [breaks](https://github.com/crossbeam-rs/crossbeam/actions/runs/2316949414) CI builds.

This fix removes unused macro arms in tests.

### DIFF
--- a/crossbeam-channel/tests/golang.rs
+++ b/crossbeam-channel/tests/golang.rs
@@ -199,14 +199,6 @@ macro_rules! defer {
 }
 
 macro_rules! go {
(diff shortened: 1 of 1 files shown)


taiki-e (commented):
Thanks for the PR! This lint is now allowed by default (https://github.com/rust-lang/rust/pull/97032), but I think addressing some of them is reasonable.

taiki-e (approved):
bors r+

PR #831, opened 2022-05-17
Add spin loop hints in tests for Miri

cbeuw:
This is a better way to do #829 

Miri does not have a pre-emptive scheduler, so once the execution falls into a spin loop it'll hang forever: https://github.com/rust-lang/miri/issues/1388

Similar measures (`yield_now()`) are already present in [some other tests](https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-queue/tests/array_queue.rs), but it's missing here


taiki-e (approved):
Thanks!

bors r+

PR #832, opened 2022-05-19
Fix links to macro rules in docs

alygin:
The fix provides explicit links to macro rules in docs because they [are not resolved globally anymore](https://github.com/rust-lang/rust/pull/96676).


taiki-e (approved):
Thanks!

bors r+

Issue #748, opened 2021-10-20
`AtomicCell` accessing uninitialized memory


CeleritasCelery:
Run the follow code sample with `cargo +nightly miri run`

```rust
use crossbeam_utils::atomic::AtomicCell;

#[allow(dead_code)]
#[repr(align(8))]
#[derive(Copy, Clone, Debug)]
enum Test {
    Field(u32),
    FieldLess,
}

fn main() {
    assert!(AtomicCell::<Test>::is_lock_free());
    let x = AtomicCell::new(Test::FieldLess);
    println!("{:?}", x.load());
}
```

I see the following error from Miri
```
error: Undefined Behavior: type validation failed at .value.<enum-tag>: encountered uninitialized bytes, but expected a valid enum tag
note: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
```

I think the root of this issue is that `AtomicCell` is trying to do a `transmute_copy` on the `None` value, but the upper bits are uninitialized, therefore `AtomicCell` is reading uninitialized data.

crossbeam-utils = "0.8.5"
cargo 1.56.0-nightly (cc17afbb0 2021-08-02)

taiki-e:
This seems an instance of https://github.com/rust-lang/unsafe-code-guidelines/issues/69?

I don't know if there is a way to fix this issue other than to restrict the types that can be used in AtomicCell using something like unsafe marker traits.

taiki-e:
cc @RalfJung @jeehoonkang @Amanieu

RalfJung:
> This seems an instance of rust-lang/unsafe-code-guidelines#69?

Well, having uninitialized tags violates [Rust's UB rules](https://doc.rust-lang.org/reference/behavior-considered-undefined.html): enum values must have valid discriminants.

It seems like the `AtomicCell` implementation should be using `MaybeUninit` somewhere, to properly express that it is dealing with potentially uninit data.

adamreichold:
>  enum values must have valid discriminants.

But isn't the discriminant valid in the above example? My understanding is that the rest of the enum is uninitialized (because writing `Test::FieldLess` only writes the discriminant) and `AtomicCell::<T>` always reads the full `size_of::<T>()` bytes.

RalfJung:
> But isn't the discriminant valid in the above example?

Yes, and a `transmute_copy` of that data at the given type would be fine. But `AtomicCell` evidently does something more interesting.

[...]

PR #834, opened 2022-05-22
Use `UnsafeCell<MaybeUninit<T>>` in AtomicCell

### LINKED ISSUE #833: AtomicCell is unsound on types containing niches
I believe it's incorrect to implement `AtomicCell<T>` in terms of `UnsafeCell<T>`, and it needs to be `UnsafeCell<MaybeUninit<T>>` instead, to prevent code outside the cell from observing partially initialized state.

Here is an example of safe code that reproduces UB:

```rust
use crossbeam_utils::atomic::AtomicCell;
use std::num::NonZeroU128;
use std::thread;

enum Enum {
    NeverConstructed,
    Cell(AtomicCell<NonZeroU128>),
}

static STATIC: Enum = Enum::Cell(AtomicCell::new(match NonZeroU128::new(1) {
    Some(nonzero) => nonzero,
    None => unreachable!(),
}));

fn main() {
    thread::spawn(|| {
        let cell = match &STATIC {
            Enum::NeverConstructed => unreachable!(),
            Enum::Cell(cell) => cell,
        };
        let x = NonZeroU128::new(0xFFFFFFFF_FFFFFFFF_00000000_00000000).unwrap();
        let y = NonZeroU128::new(0x00000000_00000000_FFFFFFFF_FFFFFFFF).unwrap();
        loop {
            cell.store(x);
            cell.store(y);
        }
    });

    loop {
        if let Enum::NeverConstructed = STATIC {
            unreachable!(":(");
        }
    }
}
```

```console
$ cargo run

warning: variant is never constructed: `NeverConstructed`
 --> src/main.rs:6:5
  |
6 |     NeverConstructed,
  |     ^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(dead_code)]` on by default

warning: `repro` (bin "repro") generated 1 warning
    Finished dev [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/repro`

thread 'main' panicked at 'internal error: entered unreachable code: :(', src/main.rs:31:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

taiki-e:
Fixes #833

Note: This contains two breaking changes:
- Unsized values are no longer allowed.
  This is because `MaybeUninit` doesn't allow it.
- `AtomicCell` now implements `Drop`.
  This is because `MaybeUninit` prevents `T` from being dropped, so we need to implement `Drop` for `AtomicCell` to avoid leaks of non-`Copy` types.

Breakages are allowed because this is a soundness bug fix. However, given the amount of breakage, we would not be able to yank the affected releases and would only create an advisory.

### DIFF
--- a/crossbeam-utils/src/atomic/atomic_cell.rs
+++ b/crossbeam-utils/src/atomic/atomic_cell.rs
@@ -5,7 +5,7 @@ use crate::primitive::sync::atomic::{self, AtomicBool};
 use core::cell::UnsafeCell;
 use core::cmp;
 use core::fmt;
-use core::mem::{self, MaybeUninit};
+use core::mem::{self, ManuallyDrop, MaybeUninit};
 use core::sync::atomic::Ordering;
 
 use core::ptr;
@@ -39,6 +39,10 @@ pub struct AtomicCell<T> {
     ///
     /// Using MaybeUninit to prevent code outside the cell from observing partially initialized state:
     /// <https://github.com/crossbeam-rs/crossbeam/issues/833>
+    ///
+    /// Note:
+    /// - we'll never store uninitialized `T` due to our API only using initialized `T`.
+    /// - this `MaybeUninit` does *not* fix <https://github.com/crossbeam-rs/crossbeam/issues/315>.
     value: UnsafeCell<MaybeUninit<T>>,
 }
 
@@ -68,6 +72,9 @@ impl<T> AtomicCell<T> {
 
     /// Consumes the atomic and returns the contained value.
     ///
+    /// This is safe because passing `self` by value guarantees that no other threads are
+    /// concurrently accessing the atomic data.
+    ///
     /// # Examples
     ///
     /// ```
@@ -79,8 +86,13 @@ impl<T> AtomicCell<T> {
     /// assert_eq!(v, 7);
     /// ```
     pub fn into_inner(self) -> T {
-        // SAFETY: we'll never store uninitialized `T`
-        unsafe { self.value.into_inner().assume_init() }
+        let this = ManuallyDrop::new(self);
+        // SAFETY:
+        // - passing `self` by value guarantees that no other threads are concurrently
+        //   accessing the atomic data
+        // - the raw pointer passed in is valid because we got it from an owned value.
+        // - `ManuallyDrop` prevents double dropping `T`
+        unsafe { this.as_ptr().read() }
     }
 
     /// Returns `true` if operations on values of this type are lock-free.
@@ -291,6 +303,22 @@ impl<T: Copy + Eq> AtomicCell<T> {
             }
         }
         Err(prev)
+    }
+}
+
+// `MaybeUninit` prevents `T` from being dropped, so we need to implement `Drop`
+// for `AtomicCell` to avoid leaks of non-`Copy` types.
+impl<T> Drop for AtomicCell<T> {
+    fn drop(&mut self) {
+        if mem::needs_drop::<T>() {
+            // SAFETY:
+            // - the mutable reference guarantees that no other threads are concurrently accessing the atomic data
+            // - the raw pointer passed in is valid because we got it from a reference
+            // - `MaybeUninit` prevents double dropping `T`
+            unsafe {
+                self.as_ptr().drop_in_place();
+            }
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


crossbeam-utils/src/atomic/atomic_cell.rs
```diff
@@ -292,6 +306,22 @@ impl<T: Copy + Eq> AtomicCell<T> {
...
+// for `AtomicCell` to avoid leaks of non-`Copy` types.
+impl<T> Drop for AtomicCell<T> {
+    fn drop(&mut self) {
+        if mem::needs_drop::<T>() {
+            // SAFETY:
+            // - the mutable reference guarantees that no other threads are concurrently accessing the atomic data
+            // - the raw pointer passed in is valid because we got it from a reference
+            // - `MaybeUninit` prevents double dropping `T`
+            unsafe {
+                drop(self.as_ptr().read());
```
dtolnay, line 319: I don't think this should copy the data before dropping. It could be quite big.

```suggestion
                ptr::drop_in_place(self.as_ptr());
```

dtolnay (changes requested):

RalfJung:
> Fixes https://github.com/crossbeam-rs/crossbeam/issues/833

I am a bit surprised to see this fix on the crossbeam side. I thought this should be fixed on the rustc side by fixing https://github.com/rust-lang/rust/issues/87341?

However, I think the same change is *still* needed in crossbeam, to fix the program that AtomicCell is unsound for types containing padding (https://github.com/crossbeam-rs/crossbeam/issues/748).

RalfJung:
Also, Miri still fails on CI, since this still uses `AtomicI64` and friends under the hood -- the problem is just moved one layer down.

We need new language features to support atomic operations on `MaybeUninit<$int>`.

taiki-e:
@RalfJung 

> I am a bit surprised to see this fix on the crossbeam side. I thought this should be fixed on the rustc side by fixing [rust-lang/rust#87341](https://github.com/rust-lang/rust/issues/87341)?

I agree that this should be fixed on the rustc side. (However, AFAIK, until this PR was created, using `MaybeUninit` was misunderstood to be a sufficient way -- the issue was [marked I-unsound a few hours after this PR was created](https://github.com/rust-lang/rust/issues/87341#issuecomment-1133869785))

> However, I think the same change is _still_ needed in crossbeam, to fix the program that AtomicCell is unsound for types containing padding (#748).

IIUC, that problem is about treating types containing uninitialized bytes as (initialized) integers, and what this PR does is to wrap types containing uninitialized bytes in `MaybeUninit`. So, I believe this PR is not sufficient to fix that (we need `AtomicMaybeUninit`).

> Also, Miri still fails on CI, since this still uses `AtomicI64` and friends under the hood -- the problem is just moved one layer down.

The test added in this PR uses `AtomicCell<NonZeroU128>`, so the `AtomicCell` uses global locks instead of atomic operations (128-bit atomics is not stabilized, so crossbeam doesn't use it). So, failure is probably due to just uses of detatched threads.

> We need new language features to support atomic operations on `MaybeUninit<$int>`.

I'm working on [implementing `AtomicMaybeUninit` at the library level (by using inline assembly)](https://github.com/taiki-e/atomic-maybe-uninit). That said, providing it as a language feature would be ideal.

crossbeam-utils/src/atomic/atomic_cell.rs
taiki-e: Changed to use drop_in_place. Thanks for the suggestion!

dtolnay:
> I am a bit surprised to see this fix on the crossbeam side. I thought this should be fixed on the rustc side by fixing https://github.com/rust-lang/rust/issues/87341?

Even if fixed in rustc, this crate would remain prone to data races/UB on stable for the next 11 weeks and on older compilers permanently. I think it's worth considering a fix in this crate even if a fix in rustc is happening. This isn't a theoretical "hypothetically a future rustc could miscompile your code because there is theoretically UB here" — it's already observably doing the wrong thing on code that's reasonably typical of real-world usage of this crate (putting `AtomicCell` inside a struct inside an Option is not a rare situation).

The fact that this currently unavoidably breaks `?Sized` support would be the only hesitation about doing a fix in crossbeam, but I think that discussion would only be about whether to release the workaround in a patch vs an incompatible semver version. How disruptive a patch for this would be depends on how much existing code involves `AtomicCell<T>` with `T: ?Sized`. I don't know of any, but I wouldn't necessarily know of any. The `?Sized` has been there since #360 which doesn't seem like it was motivated by any particular use case.

Separately we should definitely fix rustc, and also fix the standard library to make `MaybeUninit` support `T: ?Sized`.

RalfJung:
> fix the standard library to make MaybeUninit support T: ?Sized.

That is very hard at best. See https://github.com/rust-lang/rust/issues/80158.

PR #835, opened 2022-05-22
channel: Replace Spinlock with Mutex

taiki-e:
Addresses one of @thomcc's reviews in https://github.com/rust-lang/rust/pull/93563#discussion_r799137958 (sorry for the late response!)

Historically, the use of spinlock was introduced when the dependency on parking_lot was removed (https://github.com/crossbeam-rs/crossbeam/commit/5208895dbcf568fc92e70b1644217dfe1b634e51).
However, given that it is used in Waker, which includes vectors that may be reallocated, using mutex seems to be the right choice here.

r? @ibraheemdev
bors d=ibraheemdev

### DIFF
--- a/crossbeam-channel/src/flavors/zero.rs
+++ b/crossbeam-channel/src/flavors/zero.rs
@@ -5,6 +5,7 @@
 use std::cell::UnsafeCell;
 use std::marker::PhantomData;
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Mutex;
 use std::time::Instant;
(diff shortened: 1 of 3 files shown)


ibraheemdev:
A better design might be to use an intrusive linked list for wakers, which could avoid allocations entirely in the critical section, but this seems like a good change either way.

taiki-e:
bors r=ibraheemdev

(The command of bors used in rust-lang is `@bors`, but the command of bors-ng used in the community is `bors`)

PR #836, opened 2022-05-22
channel: Do not cache thread id on aarch64 linux

taiki-e:
From https://github.com/crossbeam-rs/crossbeam/pull/802#issuecomment-1074579963

> For the record, taking 0.5.4 works, and changing `try_select` to have `#[inline(never)]` breaks it again.

Given that an annotation about inlining reintroduces the issue, it appears that this may also be reintroduced by options such as opt-level and LTO.

Disabling cache on aarch64 linux until the problem is fixed upstream is probably a more robust approach.

### DIFF
--- a/crossbeam-channel/src/waker.rs
+++ b/crossbeam-channel/src/waker.rs
@@ -273,6 +273,7 @@ impl Drop for SyncWaker {
 }
 
 /// Returns the id of the current thread.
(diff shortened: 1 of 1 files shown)


taiki-e:
cc @glandium

taiki-e:
The current one seems to be working anyway, so will close this for now.

PR #843, opened 2022-05-31
Fix miri tests

alygin:
There were several changes in Miri recently, including [breaking ones](https://github.com/rust-lang/miri/pull/2105) and affecting its speed. They caused CI builds failures.

This fix makes the most expensive Miri tests less intensive, so they have decent execution time again.


taiki-e (approved):
Thanks!

bors r+

PR #844, opened 2022-05-31
Change panic handling in scoped thread

### LINKED ISSUE #816: Memory leak in `crossbeam_utils::thread::scope()`: Unbounded memory usage with long-lived scopes
`crossbeam_utils::thread::Scope` owns a vector of shared join handles:

https://github.com/crossbeam-rs/crossbeam/blob/450d237d1546ac135d21f75b61fdf5a944e8421c/crossbeam-utils/src/thread.rs#L193-L203

Each call to `scope.spawn()` pushes a new handle to the vec:

https://github.com/crossbeam-rs/crossbeam/blob/450d237d1546ac135d21f75b61fdf5a944e8421c/crossbeam-utils/src/thread.rs#L458-L459

The vec never shrinks until it is drained all at once at the end of `scope()`:

https://github.com/crossbeam-rs/crossbeam/blob/450d237d1546ac135d21f75b61fdf5a944e8421c/crossbeam-utils/src/thread.rs#L167-L176

Thus, any long-lived thread scope will exhibit unbounded memory usage as it spawns more and more threads, even if those threads finish executing or are joined by the user. This is a huge problem for servers or other such long-running applications that want to spawn scoped threads.

I would understand if there are reasons that `scope()` can't be implemented in such a way where these handles are cleaned up as threads complete. But if that is the case, then I feel this needs to be a **very explicitly warned about** in the crate's documentation and README. Cuz this is really bad. We believe that this issue is the root cause of a recent out-of-memory crash on an enterprise customer's deployment of our software product.

taiki-e:
- Panic instead of returning result
- Do not own join handles in main thread

TODO: if we don't do https://github.com/crossbeam-rs/crossbeam/pull/844#issuecomment-1142382387, we probably shouldn't change the signature of scope function.

Fixes #816
Closes #724

### DIFF
--- a/crossbeam-channel/benches/crossbeam.rs
+++ b/crossbeam-channel/benches/crossbeam.rs
@@ -64,8 +64,7 @@ mod unbounded {
                 }
             });
             drop(s1);
-        })
-        .unwrap();
+        });
     }
 
     #[bench]
@@ -93,8 +92,7 @@ mod unbounded {
                 r2.recv().unwrap();
             });
             drop(s1);
-        })
-        .unwrap();
+        });
     }
 
     #[bench]
@@ -129,8 +127,7 @@ mod unbounded {
                 }
             });
             drop(s1);
-        })
-        .unwrap();
+        });
     }
 
     #[bench]
@@ -165,8 +162,7 @@ mod unbounded {
                 }
             });
             drop(s1);
-        })
-        .unwrap();
+        });
     }
 
     #[bench]
@@ -208,8 +204,7 @@ mod unbounded {
                 }
             });
             drop(s1);
-        })
-        .unwrap();
+        });
     }
 }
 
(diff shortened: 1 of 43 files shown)


taiki-e:
BTW, in addition to this, I would like to port the <https://github.com/rust-lang/rust/pull/94559>'s changes to crossbeam.

taiki-e:
@m-ou-se Thanks! I pushed [a fix](https://github.com/crossbeam-rs/crossbeam/compare/1aca853b7a81881410ff8196d01fa459fd37d330..e637a68d020df7589a781041608ba0d1736e4716) for that bug

taiki-e:
bors r- 

miri failed due to thread leaks: https://github.com/crossbeam-rs/crossbeam/runs/7479452681?check_suite_focus=true

PR #855, opened 2022-06-26
Miri: enable preemption again

RalfJung:
Miri has a [work-around](https://github.com/rust-lang/miri/pull/2248) for https://github.com/rust-lang/rust/issues/55005 now.

Also, "check-number-validity" is on by default these days.  And I am not sure why "compare-exchange-weak-failure-rate=0.0" was set so let's see what happens when we remove it. :)

### DIFF
--- a/ci/miri.sh
+++ b/ci/miri.sh
@@ -5,35 +5,30 @@ cd "$(dirname "$0")"/..
 
 export RUSTFLAGS="${RUSTFLAGS:-} -Z randomize-layout"
 
-# disable preemption due to https://github.com/rust-lang/rust/issues/55005
-MIRIFLAGS="-Zmiri-strict-provenance -Zmiri-symbolic-alignment-check -Zmiri-disable-isolation -Zmiri-preemption-rate=0" \
+MIRIFLAGS="-Zmiri-strict-provenance -Zmiri-symbolic-alignment-check -Zmiri-disable-isolation" \
     cargo miri test \
     -p crossbeam-queue \
     -p crossbeam-utils
 
 # -Zmiri-ignore-leaks is needed because we use detached threads in tests/docs: https://github.com/rust-lang/miri/issues/1371
-# disable preemption due to https://github.com/rust-lang/rust/issues/55005
-MIRIFLAGS="-Zmiri-strict-provenance -Zmiri-symbolic-alignment-check -Zmiri-disable-isolation -Zmiri-ignore-leaks -Zmiri-preemption-rate=0" \
+MIRIFLAGS="-Zmiri-strict-provenance -Zmiri-symbolic-alignment-check -Zmiri-disable-isolation -Zmiri-ignore-leaks" \
     cargo miri test \
     -p crossbeam-channel
 
 # -Zmiri-ignore-leaks is needed for https://github.com/crossbeam-rs/crossbeam/issues/579
 # -Zmiri-disable-stacked-borrows is needed for https://github.com/crossbeam-rs/crossbeam/issues/545
-# disable preemption due to https://github.com/rust-lang/rust/issues/55005
-MIRIFLAGS="-Zmiri-check-number-validity -Zmiri-symbolic-alignment-check -Zmiri-disable-isolation -Zmiri-disable-stacked-borrows -Zmiri-ignore-leaks -Zmiri-preemption-rate=0" \
+MIRIFLAGS="-Zmiri-symbolic-alignment-check -Zmiri-disable-isolation -Zmiri-disable-stacked-borrows -Zmiri-ignore-leaks" \
     cargo miri test \
     -p crossbeam-epoch \
     -p crossbeam-skiplist
 
 # -Zmiri-ignore-leaks is needed for https://github.com/crossbeam-rs/crossbeam/issues/579
 # -Zmiri-disable-stacked-borrows is needed for https://github.com/crossbeam-rs/crossbeam/issues/545
-# disable preemption due to https://github.com/rust-lang/rust/issues/55005
-MIRIFLAGS="-Zmiri-check-number-validity -Zmiri-symbolic-alignment-check -Zmiri-disable-stacked-borrows -Zmiri-ignore-leaks -Zmiri-compare-exchange-weak-failure-rate=0.0 -Zmiri-preemption-rate=0" \
+MIRIFLAGS="-Zmiri-symbolic-alignment-check -Zmiri-disable-stacked-borrows -Zmiri-ignore-leaks -Zmiri-compare-exchange-weak-failure-rate=0.0" \
     cargo miri test \
     -p crossbeam-deque
 
 # -Zmiri-ignore-leaks is needed for https://github.com/crossbeam-rs/crossbeam/issues/579
-# disable preemption due to https://github.com/rust-lang/rust/issues/55005
-MIRIFLAGS="-Zmiri-check-number-validity -Zmiri-symbolic-alignment-check -Zmiri-ignore-leaks -Zmiri-preemption-rate=0" \
+MIRIFLAGS="-Zmiri-symbolic-alignment-check -Zmiri-ignore-leaks" \
     cargo miri test \
     -p crossbeam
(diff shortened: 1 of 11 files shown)


taiki-e:
Thanks!

> I am not sure why "compare-exchange-weak-failure-rate=0.0" was set

This is because some single-threaded tests (e.g., doctest of Stealer::steal) assume that weak CAS will not fail. [It is not strictly a correct assumption](https://github.com/rust-lang/miri/pull/1686#issuecomment-767394783), but there were many tests (especially doctests) that depended on that behavior, so I had to set it. (sorry I forgot to leave a comment explaining this)

RalfJung:
Also looks like I failed to actually remove that flag.^^ I'll add a comment instead.

RalfJung:
That failure in crossbeam-dequeue stampede looks real though?
```
error: Undefined Behavior: type validation failed: encountered a dangling box (use-after-free)
    --> /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:1504:9
     |
1504 |         intrinsics::volatile_load(src)
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ type validation failed: encountered a dangling box (use-after-free)
     |
     = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
     = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
     = note: inside `std::ptr::read_volatile::<std::boxed::Box<usize>>` at /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/mod.rs:1504:9
note: inside `crossbeam_deque::deque::Buffer::<std::boxed::Box<usize>>::read` at /home/runner/work/crossbeam/crossbeam/crossbeam-deque/src/deque.rs:76:9
    --> /home/runner/work/crossbeam/crossbeam/crossbeam-deque/src/deque.rs:76:9
     |
76   |         ptr::read_volatile(self.at(index))
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: inside `crossbeam_deque::Stealer::<std::boxed::Box<usize>>::steal` at /home/runner/work/crossbeam/crossbeam/crossbeam-deque/src/deque.rs:652:29
    --> /home/runner/work/crossbeam/crossbeam/crossbeam-deque/src/deque.rs:652:29
     |
652  |         let task = unsafe { buffer.deref().read(f) };
     |                             ^^^^^^^^^^^^^^^^^^^^^^
note: inside closure at crossbeam-deque/tests/fifo.rs:126:41
    --> crossbeam-deque/tests/fifo.rs:126:41
     |
126  |                     if let Success(x) = s.steal() {
     |                                         ^^^^^^^^^
note: inside closure at /home/runner/work/crossbeam/crossbeam/crossbeam-utils/src/thread.rs:438:31
    --> /home/runner/work/crossbeam/crossbeam/crossbeam-utils/src/thread.rs:438:31
     |
438  |                     let res = f(&scope);
     |                               ^^^^^^^^^
```
Is there an issue for that?

RalfJung:
I think I fixed a soundness bug. :D  And the code also became cleaner by removing the need for a bunch of `mem::forget`.
Let me know if you'd prefer that to be a separate PR.

taiki-e:
Thanks. IIUC, that was real use-after-free before https://github.com/crossbeam-rs/crossbeam/security/advisories/GHSA-pqqp-xmhj-wgcw fixed. After that fix, that value is never used, but your fix to use `MaybeUninit` looks correct because it is possible to read a value that is not valid as `T`.

RalfJung:
Hm, but something seems to not be right with the patch...

taiki-e:
I think you need to change this line:
https://github.com/crossbeam-rs/crossbeam/blob/f12133c95fb0f57b8c10791273c49fadedfb2863/crossbeam-deque/src/deque.rs#L127

to: 

```rs
(buffer.deref().at(i) as *mut T).drop_in_place();
```

crossbeam-deque/src/deque.rs
```diff
@@ -62,7 +64,7 @@ impl<T> Buffer<T> {
     /// technically speaking a data race and therefore UB. We should use an atomic store here, but
     /// that would be more expensive and difficult to implement generically for all types `T`.
     /// Hence, as a hack, we use a volatile write instead.
-    unsafe fn write(&self, index: isize, task: T) {
+    unsafe fn write(&self, index: isize, task: MaybeUninit<T>) {
```
taiki-e, line 67: Since the value written to `self.at(index)` must be valid `T` (otherwise, the destructor cannot be sure that the value is valid), it may be better to use `T` instead of `MaybeUninit<T>` in the argument (and use `MaybeUninit::new(task)` in the next line).

RalfJung: I did this because there are some places that just `read` the data and `write` it back immediately. By taking `MaybeUninit<T>` here we can avoid having to assert that this data is valid.

RalfJung:
> I think you need to change this line:

Ah, yes!

crossbeam-deque/src/deque.rs
taiki-e: Ah, that makes sense.

RalfJung:
Btw, Miri might at some point actually notice that these volatile reads/writes should be atomic. We do have a data race detector, after all. I am not sure why it doesn't kick in here; probably it needs some really specific interleaving to actually be a race.

RalfJung:
Well, I think that is exactly what is happening in the latest test failure. :)

RalfJung:
I added timing because the tests became a lot slower, but now they are fast again... whatever.^^
This still helps to identify slow tests, so if you don't mind I will keep it in.

taiki-e (approved):
Thanks again!

bors r+

ci/miri.sh
```diff
@@ -3,37 +3,40 @@ set -euxo pipefail
 IFS=$'\n\t'
 cd "$(dirname "$0")"/..
 
+# We need 'ts' for the per-line timing
+sudo apt-get -y install moreutils
+echo
```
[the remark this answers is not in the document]
RalfJung: Ah, I didn't know about this flag.
However (a) it only works with isolation disabled (since for regular tests it actually measures the time *inside* Miri), (b) it is harder to read without colors and libtest output in Miri is never colored for $REASONS, (c) the colors might not be appropriate for Miri; 1.3s seems still pretty good to me. ;)

Up to you, I am fine with removing the `ts` thing again. You have isolation disabled anyway.

RalfJung:
4h? Either there is a non-deterministic infinite loop somewhere or the test runner just got stuck. Let me still reduce some test sizes.

Argh, Rust 1.36 is *ancient*...

RalfJung:
What is clippy talking about, I didn't even change that code?!?

taiki-e:
bors r+

It seems that a new clippy lint was added in Rust 1.62 released about 10 hours ago.

taiki-e:
bors r-

https://github.com/crossbeam-rs/crossbeam/runs/7142722550?check_suite_focus=true

```
test arc ... error: unsupported operation: integer-to-pointer casts and `ptr::from_exposed_addr` are not supported with `-Zmiri-strict-provenance`
0.000047      --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/once_cell-1.12.0/src/imp_std.rs:175:30
0.000007       |
0.000005   175 |             let mut waiter = (queue & !STATE_MASK) as *const Waiter;
0.000006       |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ integer-to-pointer casts and `ptr::from_exposed_addr` are not supported with `-Zmiri-strict-provenance`
0.000005       |
0.000006       = help: use Strict Provenance APIs (https://doc.rust-lang.org/nightly/std/ptr/index.html#strict-provenance, https://crates.io/crates/sptr) instead
0.000005    
```

RalfJung:
Yeah once_cell does not work with strict provenance. It never did. But that only became apparent with a preemptive scheduler (the casts only occur when threads actually race on a once_cell).

PR #858, opened 2022-07-01
Ignore clippy::let_unit_value lint

taiki-e:


taiki-e:
auto retry is not triggered or is very slow.

bors r-
bors r+

Issue #852, opened 2022-06-22
Noticeable overhead in crossbeam-epoch under many Locals


ryoqun:
Hi, nice to meet you and thanks for maintaining this crate. :) 

I found some kind of performance issue spanning cross-crate code interactions, ultimately resulting in many cpu cycles being wasted at `crossbeam-epoch`. And I'm wondering where the proper fix could be placed across those related crates.

In short, there are circumstances where `crossbeam-epoch`'s epoch bookkeeping code exposes significant overhead. And our production code was hit at it.

Dirtiest hack would be reducing frequency of gc collection (= `global().collect(&guard);`) by increasing `PINNINGS_BETWEEN_COLLECT` like this:

```patch
$ git diff
diff --git a/crossbeam-epoch/src/internal.rs b/crossbeam-epoch/src/internal.rs
index de208b1..0394c03 100644
--- a/crossbeam-epoch/src/internal.rs
+++ b/crossbeam-epoch/src/internal.rs
@@ -387,7 +387,7 @@ fn local_size() {
 impl Local {
     /// Number of pinnings after which a participant will execute some deferred functions from the
     /// global queue.
-    const PINNINGS_BETWEEN_COLLECT: usize = 128;
+    const PINNINGS_BETWEEN_COLLECT: usize = 128 * 128;
 
     /// Registers a new `Local` in the provided `Global`.
     pub(crate) fn register(collector: &Collector) -> LocalHandle {

```

However, I'm not fully sure this is the right fix (esp, regarding its ramifications due to reduced garbage collections).

Let me explain the rather complicated context a bit.

Firstly, `crossbeam-epoch` is used by `crossbeam-deque` (duh), which in turn by `rayon` (task scheduler library) as task queues , then by `solana-validator` (which experinced the performance issue; https://github.com/solana-labs/solana/issues/22603).

so far, so good. it's just normal use case of rayon for multi cores by an application code.

The twist is that `solana-validator` is holding *many* `rayon` thread pools managed by its internal subsystems. So, it's well exceeding system's core count by great factor like 2000 threads on a 64 core machine.

[...]

PR #862, opened 2022-07-11
Add reduced-collection-frequency for niche use case

ryoqun:
to be documented

related to https://github.com/crossbeam-rs/crossbeam/issues/852


ibraheemdev:
Couldn't this be added as a config option on `Collector`, and maybe make it atomic so the global collector's can be updated?

PR #796, opened 2022-03-03
epoch: Remove ptr-to-int casts

taiki-e:
Use [this hack](https://github.com/rust-lang/miri/issues/1866#issuecomment-985802751) to fix compatibility issues with Miri (see https://github.com/crossbeam-rs/crossbeam/pull/490#issuecomment-755435349 for details). 

Due to the https://github.com/crossbeam-rs/crossbeam/issues/545, still not compatible with stacked borrows. This will be fixed by the subsequent PR (#871).

Note: this is a breaking change because changes API of Pointable and Pointer traits

Fixes #579

PR #871, opened 2022-07-20
epoch: Fix stacked borrows violations

### LINKED ISSUE #545: MIRI reports UB in crossbeam::epoch::Collector's deallocation
Hello.

Another MIRI unbounded behavior for destructors, this time in Collector.
```
 let x1_0: crossbeam::epoch::Collector = crossbeam::epoch::Collector::new();//LAYER:0
 let x2_0: & crossbeam::epoch::Collector = & x1_0;//LAYER:1
 let x3_0: crossbeam::epoch::LocalHandle = x2_0.register();//LAYER:2
```

Running this gives the MIRI error at deallocation:
```
error: Undefined Behavior: deallocating while item is protected: [SharedReadWrite for <235744> (call 76130)]
<trace>
```
I am not super familiar with your internal memory structure, but this may be related to a similar issue seen in [another garbage collection implementation](https://github.com/fitzgen/bacon-rajan-cc/issues/12). Maybe it will help with the patch.

A [code file with a full trace](https://github.com/crossbeam-rs/crossbeam/files/5099590/4_261362.rs.zip) attached. It looks a little weird because the test case was automatically generated.

As for the impact of this issue, it doesn't seem to be too big of an issue as it is mostly contained. Use-after-free **should not be possible** unless something else happens during dealloc. I can't be 100% sure because Miri stops at that point, but looking at the code in list.rs, it looks okay. It may become more problematic if people start adding more unsafe functions that expose internal memory in some way.

Thanks
~Yoshi

taiki-e:
Fixes #545

Based on #796 (the first and second commits are from #796)


crossbeam-epoch/src/atomic.rs
```diff
@@ -113,26 +115,26 @@ pub trait Pointable {
     ///
     /// - The given `ptr` should have been initialized with [`Pointable::init`].
     /// - `ptr` should not have yet been dropped by [`Pointable::drop`].
-    /// - `ptr` should not be mutably dereferenced by [`Pointable::deref_mut`] concurrently.
-    unsafe fn deref<'a>(ptr: *mut ()) -> &'a Self;
+    /// - `ptr` should not be mutably dereferenced by [`Pointable::as_mut_ptr`] concurrently.
+    unsafe fn as_ptr(ptr: *mut ()) -> *const Self;
 
     /// Mutably dereferences the given pointer.
```
taiki-e, line 121: This is no longer a dereference, but I'm not sure what is a good API name and description.

taiki-e:
(Given that Miri appears to be trying to move to TB in the long term, I don't see the need to rush to merge this.)

jonasmalacofilho:
> (Given that Miri appears to be trying to move to TB in the long term, I don't see the need to rush to merge this.)

Is this documented anywhere and current?

Miri's README currently says:

> While Tree Borrows is still sound in the sense of catching all aliasing violations that current versions of the compiler might exploit, **it is likely that the eventual final aliasing model of Rust will be stricter than Tree Borrows.** In other words, **if you use Tree Borrows, even if your code is accepted today, it might be declared UB in the future. This is much less likely with Stacked Borrows.** Using Tree Borrows currently implies -Zmiri-strict-provenance because integer-to-pointer casts are not supported in this mode, but that may change in the future.

taiki-e:
> > (Given that Miri appears to be trying to move to TB in the long term, I don't see the need to rush to merge this.)
>
> That's wrong. There is no plan to move to TB entirely, 

That comment was my understanding based on reading several discussions and blog posts at 2023-06 and before, and I may have misunderstood something. (My understanding at the time was that they wanted to adopt a somewhat stricter version of TB.)

Well, to be honest, that comment should no longer be relevant to this PR. The remaining issues with the epoch and issues in skiplist that made me hesitate to progress this PR at the time have been fixed in the last two commits in this PR (thanks @Imberflur for the skiplist fix), so there is no longer any need to consider TB-only support.

> and in any case TB doesn't help here because the implementation in Miri (currently, at least) doesn't support the int-to-ptr that crossbeam-epoch currently relies on.

That issue has already been resolved in the master branch: https://github.com/crossbeam-rs/crossbeam/pull/796

PR #877, opened 2022-07-22
Bump MSRV to Rust 1.38

taiki-e:
It was released [nearly 3 years ago](https://blog.rust-lang.org/2019/09/26/Rust-1.38.0.html) and meets our MSRV policy.

This allows using const_in_array_repeat_expressions (https://github.com/rust-lang/rust/issues/49147) and `<pointer>.cast()`.


taiki-e:
FYI @cuviper (because this PR affects rayon's MSRV)

Also, aside from this PR, I would like to update our MSRV policy to support 1 year instead of 6 months. Otherwise, it is impossible for rayon to actually maintain their MSRV policy (because our current MSRV policy allows us to increase the MSRV as a minor change).

cuviper:
Thanks for the heads-up. I have no problem with this tiny bump. :slightly_smiling_face: 

FWIW, I'm looking at 1.56 in rayon-rs/rayon#964, but that will wait a few months yet to meet our 1-year MSRV policy.

dhardy:
FYI this has broken rand's CI (using Rust 1.36). It's not critical since it's only through `rand` using `rayon` as a dev-dependency, but still a little annoying seeing a patch release bump the MSRV.

taiki-e:
@dhardy

> still a little annoying seeing a patch release bump the MSRV.

We had treated MSRV bump as a major change in the past, but maintaining that policy was a pain for both maintainers and users, so we changed our policy. See https://github.com/crossbeam-rs/crossbeam/pull/444#discussion_r343737057 and https://github.com/crossbeam-rs/crossbeam/issues/503#issuecomment-631106295 for more.

(FYI, https://github.com/rust-lang/libc/pull/2845 and https://github.com/rust-lang/libs-team/issues/72 are the latest discussions on MSRV for crate widely used in the ecosystem. )

> It's not critical since it's only through rand using rayon as a dev-dependency

FWIW, AFAIK, dev-deps often have much more lenient MSRV policy than the crate under test, so there are many projects that do not run `cargo test` on MSRV, but only run `cargo build`. (e.g., [serde](https://github.com/serde-rs/serde/blob/e3d871ff7bf10dadf10bdc234a55692228358d0e/.github/workflows/ci.yml#L63-L78), [syn](https://github.com/dtolnay/syn/blob/2c83686bede4f64353e9ad10033a23f77e56eb94/.github/workflows/ci.yml), [regex](https://github.com/rust-lang/regex/blob/3bcae3b6954ea3628ef0974f0a2604a6ae248408/.github/workflows/ci.yml#L99-L103), [tokio](https://github.com/tokio-rs/tokio/blob/479c545c20b2cb44a8f09600733adc8c8dcb5aa0/.github/workflows/ci.yml#L182), [futures](https://github.com/rust-lang/futures-rs/blob/master/.github/workflows/ci.yml), etc.)

dhardy:
Committing only to support rustc versions no more than 6 months or 1 year old honestly sounds like a sensible option to me. We've seen very little concern about bumping the MSRV of rand in the past.

> but still a little annoying seeing a patch release bump the MSRV

Sorry, I shouldn't complain about something already using the best apparent solution!

tatsuya6502:
@dhardy

Just FYI. I am not sure if this is the best way, but instead of bumping the MSRV of my library, I have been working around it by pinning dependencies to some older versions in `Cargo.lock` when testing the MSRV.

[moka-rs/moka/blob/78050701a/.github/workflows/CI.yml#L55-L67](https://github.com/moka-rs/moka/blob/78050701ad38dba8f6d52a03ae5ccf91fde0f1d7/.github/workflows/CI.yml#L55-L67)

```yaml
      - name: Pin some dependencies to specific versions (MSRV only)
        if: ${{ matrix.rust == '1.51.0' }}
        # hashbrown >= v0.12 requires Rust 2021 edition.
        # native-tls >= v0.2.9 requires more recent Rust version.
        # async-global-executor >= 2.1 requires Rust 2021 edition.
        # pull-down-cmark >= 0.9.2 requires Rust 2021 edition.
        run: |
          cargo update -p dashmap --precise 5.2.0
          cargo update -p indexmap --precise 1.8.2
          cargo update -p hashbrown --precise 0.11.2
          cargo update -p native-tls --precise 0.2.8
          cargo update -p async-global-executor --precise 2.0.4
          cargo update -p pulldown-cmark --precise 0.9.1
```

PR #901, opened 2022-08-27
Implement "consume" loads on PowerPC

taylordotfish:
Like ARM, PowerPC is also a weakly ordered architecture where "acquire" loads are more expensive than "consume" loads, which require no special instructions.

### DIFF
--- a/crossbeam-utils/src/atomic/consume.rs
+++ b/crossbeam-utils/src/atomic/consume.rs
@@ -1,4 +1,9 @@
-#[cfg(any(target_arch = "arm", target_arch = "aarch64"))]
+#[cfg(any(
+    target_arch = "arm",
+    target_arch = "aarch64",
+    target_arch = "powerpc",
+    target_arch = "powerpc64",
+))]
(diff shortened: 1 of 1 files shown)


taiki-e:
Thanks for the PR! Unfortunately, the resulting code is less efficient than acquire load that uses `isync` because LLVM uses `lwsync` for powerpc's compiler_fence: https://godbolt.org/z/9MEnsWrT5

taylordotfish:
@taiki-e Huh… but why is `compiler_fence` issuing any kind of `*sync` at all? Based on the [documentation](https://doc.rust-lang.org/std/sync/atomic/fn.compiler_fence.html), I would've expected it not to generate any special instructions:

> `compiler_fence` does not emit any machine code, but restricts the kinds of memory re-ordering the compiler is allowed to do.

But empirically, that's clearly not the case…

taiki-e:
compiler_fence's documentation is incorrect, see https://github.com/rust-lang/rust/issues/62256 for more.

PR #903, opened 2022-08-30
deque: add new method to limit batch steal size

glorv:


crossbeam-deque/src/deque.rs
```diff
@@ -1639,15 +1754,15 @@ impl<T> Injector<T> {
...
             } else {
                 let len = (tail - head) >> SHIFT;
                 // Steal half of the available tasks.
-                advance = ((len + 1) / 2).min(MAX_BATCH + 1);
+                advance = ((len + 1) / 2).min(limit + 1);
             }
         } else {
             // We can steal all tasks till the end of the block.
-            advance = (BLOCK_CAP - offset).min(MAX_BATCH + 1);
+            advance = (BLOCK_CAP - offset).min(limit + 1);
```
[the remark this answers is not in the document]
glorv: fixed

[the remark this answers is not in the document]
taiki-e: Thanks. Could you add tests for this case (limit=usize::MAX)?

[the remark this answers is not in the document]
glorv: Done. While adding the new test, I found there is a small difference between the original `steal_batch_and_pop` for Stealer and Injecter, the limit for Stealer is MAX_BATCH while Injector is "MAX_BATCH+1". For the `steal_batch_and_limit_and_pop` method, I think the extra poped elements should be counted in the limit. So I changed the code a little. Please take a look again, thanks~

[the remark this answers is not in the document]
glorv: @taiki-e friendly ping.

crossbeam-deque/src/deque.rs
```diff
@@ -693,6 +693,44 @@ impl<T> Stealer<T> {
...
+    /// let w2 = Worker::new_fifo();
+    ///
+    /// let _ = s.steal_batch_with_limit(&w2, 2);
+    /// assert_eq!(w2.pop(), Some(1));
+    /// assert_eq!(w2.pop(), Some(2));
+    /// assert_eq!(w2.pop(), None);
+    ///
+    /// w1.push(7);
+    /// w1.push(8);
+    /// // set a large limit will pop half of the elements
```
taiki-e, line 727: "will pop half of the elements" is an implementation detail that may be changed in the future...

Consider the comments something like the following:

```
Setting a large limit does not guarantee that all elements will be popped. In this case, 
half of the elements are currently popped, but the number of popped elements is considered 
an implementation detail that may be changed in the future.
```

crossbeam-deque/src/deque.rs
```diff
@@ -1603,6 +1715,41 @@ impl<T> Injector<T> {
     /// assert_eq!(w.pop(), Some(2));
     /// ```
     pub fn steal_batch_and_pop(&self, dest: &Worker<T>) -> Steal<T> {
+        self.steal_batch_with_limit_and_pop(dest, MAX_BATCH)
```
taiki-e, line 1718: I'm not sure if the default of `Injector::steal_batch_and_pop` needs to be changed from `MAX_BATCH + 1`.

glorv: Yes, I also thought about it, as the number of elements in each call is not guaranteed by default and the caller should not depend on it, use `MAX_BATCH` or `MAX_BATCH + 1` should both ok here, but I prefer use `MAX_BATCH` to be the same with Stealer, `MAX_BATCH + 1` seems a bit weird to me.

crossbeam-deque/src/deque.rs
```diff
@@ -693,6 +693,47 @@ impl<T> Stealer<T> {
...
+    /// assert_eq!(w2.pop(), Some(1));
+    /// assert_eq!(w2.pop(), Some(2));
+    /// assert_eq!(w2.pop(), None);
+    ///
+    /// w1.push(7);
+    /// w1.push(8);
+    /// // Setting a large limit does not guarantee that all elements will be popped. In this case,
+    /// // half of the elements are currently popped, but the number of popped elements is considered
+    /// // an implementation detail that may be changed in the future.
+
```
taiki-e, line 730:
```suggestion
```

crossbeam-deque/src/deque.rs
```diff
@@ -891,6 +932,45 @@ impl<T> Stealer<T> {
...
+    /// let s = w1.stealer();
+    /// let w2 = Worker::new_fifo();
+    ///
+    /// assert_eq!(s.steal_batch_with_limit_and_pop(&w2, 2), Steal::Success(1));
+    /// assert_eq!(w2.pop(), Some(2));
+    /// assert_eq!(w2.pop(), None);
+    ///
+    /// w1.push(7);
+    /// w1.push(8);
+    /// assert_eq!(s.steal_batch_with_limit_and_pop(&w2, usize::MAX), Steal::Success(3));
```
taiki-e, line 966:
```suggestion
    /// // Setting a large limit does not guarantee that all elements will be popped. In this case,
    /// // half of the elements are currently popped, but the number of popped elements is considered
    /// // an implementation detail that may be changed in the future.
    /// assert_eq!(s.steal_batch_with_limit_and_pop(&w2, usize::MAX), Steal::Success(3));
```

crossbeam-deque/src/deque.rs
```diff
@@ -1444,6 +1524,41 @@ impl<T> Injector<T> {
...
+    ///
+    /// let w = Worker::new_fifo();
+    /// let _ = q.steal_batch_with_limit(&w, 2);
+    /// assert_eq!(w.pop(), Some(1));
+    /// assert_eq!(w.pop(), Some(2));
+    /// assert_eq!(w.pop(), None);
+    ///
+    /// q.push(7);
+    /// q.push(8);
+    /// // set the limit to usize::MAX will pop half of the elements
```
taiki-e, line 1556:
```suggestion
    /// // Setting a large limit does not guarantee that all elements will be popped. In this case,
    /// // half of the elements are currently popped, but the number of popped elements is considered
    /// // an implementation detail that may be changed in the future.
```

crossbeam-deque/src/deque.rs
```diff
@@ -1603,6 +1718,41 @@ impl<T> Injector<T> {
...
+    /// q.push(5);
+    /// q.push(6);
+    ///
+    /// let w = Worker::new_fifo();
+    /// assert_eq!(q.steal_batch_with_limit_and_pop(&w, 2), Steal::Success(1));
+    /// assert_eq!(w.pop(), Some(2));
+    /// assert_eq!(w.pop(), None);
+    ///
+    /// q.push(7);
+    /// // set limit to usize::MAX will pop half of the elements
```
taiki-e, line 1748:
```suggestion
    /// // Setting a large limit does not guarantee that all elements will be popped. In this case,
    /// // half of the elements are currently popped, but the number of popped elements is considered
    /// // an implementation detail that may be changed in the future.
```

crossbeam-deque/src/deque.rs
```diff
@@ -693,6 +693,47 @@ impl<T> Stealer<T> {
...
+    /// Steals no more than `limit` of tasks and pushes them into another worker.
+    ///
+    /// How many tasks exactly will be stolen is not specified. That said, this method will try to
+    /// steal around half of the tasks in the queue, but also not more than the given limit.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use crossbeam_deque::Worker;
+    ///
```
taiki-e, line 708: `usize::MAX` (associated constant) is not available for older compilers, I would suggest adding the following hidden import to allow testing doctest on those versions.

```suggestion
    /// # use std::usize;
    ///
```

crossbeam-deque/src/deque.rs
```diff
@@ -693,6 +693,47 @@ impl<T> Stealer<T> {
...
+    /// w1.push(7);
+    /// w1.push(8);
+    /// // Setting a large limit does not guarantee that all elements will be popped. In this case,
+    /// // half of the elements are currently popped, but the number of popped elements is considered
+    /// // an implementation detail that may be changed in the future.
+
+    /// let _ = s.steal_batch_with_limit(&w2, usize::MAX);
+    /// assert_eq!(w2.len(), 3);
+    ///
```
taiki-e, line 733:
```suggestion
```

crossbeam-deque/src/deque.rs
```diff
@@ -891,6 +932,45 @@ impl<T> Stealer<T> {
...
+    /// assert_eq!(w2.pop(), Some(2));
+    /// assert_eq!(w2.pop(), None);
+    ///
+    /// w1.push(7);
+    /// w1.push(8);
+    /// assert_eq!(s.steal_batch_with_limit_and_pop(&w2, usize::MAX), Steal::Success(3));
+    /// assert_eq!(w2.pop(), Some(4));
+    /// assert_eq!(w2.pop(), Some(5));
+    /// assert_eq!(w2.pop(), None);
+    ///
```
taiki-e, line 970:
```suggestion
```

crossbeam-deque/src/deque.rs
taiki-e: It may have been a performance-related tweak (https://github.com/crossbeam-rs/crossbeam/commit/af8ce60f8b32104e489859b60e3d9ae7e3de59a8). So, I tend to keep the previous one for now, but it might make sense to have a TODO comment indicating that it may need to be revisited in the future.

glorv: done

crossbeam-deque/src/deque.rs
glorv: changed to use `std::usize::MAX` directly to avoid an extra import.

taiki-e (approved):
LGTM, thanks @glorv!

bors r+

taiki-e:
bors retry

BusyJay:
@taiki-e Hi, can we get a release for this patch?

taiki-e:
Oh, sorry, I missed this had not been released. I will release it tonight.

taiki-e:
Published in crossbeam-deque 0.8.3.

PR #904, opened 2022-09-05
Improve docs of channel creation functions

yjhn:
Add note that the channel is multi-producer multi-consumer to the functions' docs. Also point to the parent module docs since they contain a lot of info on how to use the channel. These changes make functions' docs more self-contained (previously one needed to know to look at the module docs to discover this info).


taiki-e (approved):
Thanks.

bors r+

taiki-e:
bors retry

taiki-e:
bors retry

PR #907, opened 2022-09-11
Update crossbeam-channel benchmarks

yjhn:
Changes:
- updated dependencies
- updated README instructions and results
- removed `chan` benchmarks because `chan` is unmaintained since 2018
- added `flume` to the plot; I tried to include other libraries as well, but that makes graphs overlap:
plot_example


arthurprs (approved):
LGTM

taiki-e:
bors r=arthurprs

PR #916, opened 2022-09-29
Bump crossbeam-utils

MarcusGrass:
crossbeam-epoch was using a yanked version of crossbeam-utils, updated to latest released version


taiki-e:
Could you explain why this is necessary? The "0.8.5" requirement means "^0.8.5" requirement and cargo does not select a yanked version when generating lockfile or updating versions. Also, if we do this, we should do the same for other yanked versions.

MarcusGrass:
> Could you explain why this is necessary? The "0.8.5" requirement means "^0.8.5" requirement and cargo does not select a yanked version when generating lockfile or updating versions. Also, if we do this, we should do the same for other yanked versions.

No this one's on me, was checking a lib with cargo deny and the lockfile remained unupdated on some version fixing, thanks for responding

PR #917, opened 2022-10-03
Improve commented example for the never<T>() function

jimmykiang:
Fixes the following clippy warnings when the commented code example is used from the never function :

- warning: redundant closure 
- warning: use of `unwrap_or` followed by a function call

The functionality remains intact.


taiki-e (approved):
Thanks.

bors r+

riyahope77 (approved):

PR #922, opened 2022-11-01
When building for linux, replace all vendors with 'unknown'

kanavin:
Some build systems such as the Yocto project define custom rust targets, of the form 'arch-somevendor-linux-{gnu|musl}' (only the 'somevendor' part is custom, the rest matches rust's definitions).

This causes mismatches with crossbeam's lists of targets which are built from the standard rust target list, and always have 'unknown' as the vendor.

This change simply replaces such custom vendors with 'unknown' when the third component of the target is 'linux'.

---
Note from Alex: this is a reworked and hopefully better/more precise/acceptable version of https://github.com/crossbeam-rs/crossbeam/pull/751

### DIFF
--- a/build-common.rs
+++ b/build-common.rs
@@ -1,0 +1,13 @@
+// The target triplets have the form of 'arch-vendor-system'.
+//
+// When building for Linux (e.g. the 'system' part is
+// 'linux-something'), replace the vendor with 'unknown'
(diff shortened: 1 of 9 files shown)


taiki-e:
Published in crossbeam-utils 0.8.13, crossbeam-queue 0.3.7, and crossbeam-epoch 0.9.12.

PR #926, opened 2022-11-16
crossbeam-epoch: Bump memoffset to 0.7

glittershark:
mostly to keep the number of dupes of crates in my dependency graph down


taiki-e:
Thanks.

bors r+

PR #932, opened 2022-11-22
Fix panicking when target triplets has less than three elements

### LINKED ISSUE #931: Handling all Rust targets.
I notice that in [this commit](https://github.com/crossbeam-rs/crossbeam/commit/e78e511e30031785253f767caf37e7efa0d08191), we try to replace vendor in targets. But it presumes that all targets are triplets, which doesn't hold for some targets like `wasm32-wasi` and `x86_64-fuchsia`. This causes problems when [compiling crossbeam for these targets](https://github.com/second-state/microservice-rust-mysql/issues/7).

I can fix this if needed or maybe there are better solutions.

Cryptex-github:
Currently the build script panic when the target triplets have less than three elements, for example `wasm32-wasi`, this PR fixes that issue.

Fixes #931

### DIFF
--- a/build-common.rs
+++ b/build-common.rs
@@ -5,8 +5,8 @@
 // so that mapping to rust standard targets happens correctly.
 fn convert_custom_linux_target(target: String) -> String {
     let mut parts: Vec<&str> = target.split('-').collect();
-    let system = parts[2];
-    if system == "linux" {
+    let system = parts.get(2);
+    if system == Some(&"linux") {
         parts[1] = "unknown";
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks!

bors r+

taiki-e:
bors retry

taiki-e:
Published in crossbeam-utils 0.8.14, crossbeam-queue 0.3.8, and crossbeam-epoch 0.9.13.

PR #937, opened 2022-12-05
no_atomic: add `mipsel-unknown-linux`

M1cha:
We're using yocto dunfell with the unofficial meta-rust layer and it doesn't add a libc suffix.

### DIFF
--- a/no_atomic.rs
+++ b/no_atomic.rs
@@ -38,6 +38,7 @@ const NO_ATOMIC_64: &[&str] = &[
     "mips-unknown-linux-uclibc",
     "mipsel-sony-psp",
     "mipsel-sony-psx",
+    "mipsel-unknown-linux",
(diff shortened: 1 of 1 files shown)


taiki-e:
Thanks for the PR. As said in https://github.com/crossbeam-rs/crossbeam/pull/751#pullrequestreview-796739843, I would prefer not to accept a way to manually manage such a list of platforms.



That said, as said in https://github.com/crossbeam-rs/crossbeam/issues/931#issuecomment-1322991616, the current way is not ideal, and I would accept a PR that adopts the same way that the [portable-atomic](https://github.com/taiki-e/portable-atomic) has adopted.

> I believe the ideal solution here is to use `cfg_target_has_atomic` on Rust 1.60+ like my [portable-atomic](https://github.com/taiki-e/portable-atomic) crate does.

Issue #795, opened 2022-03-03
channel: Panic in .recv_timeout(Duration::MAX)


MightyPork:
https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-channel/src/channel.rs#L863

Calling `recv_timeout()` with `Duration::MAX` causes panic due to `Instant` overflow.

This happens quite naturally when the duration comes from some calculation, meaning "wait forever".

I don't think it should crash, insted wait the longest possible time.

taiki-e:
It would be ideal if there were a method like Instant::saturating_add, but unfortunately it does not exist, so for now it seems we have to [wait a few decades](https://github.com/tokio-rs/tokio/blob/fc4deaa1d0e35d3404d1d32ddcb4b083081355c0/tokio/src/time/instant.rs#L57-L63) or wait forever.

MightyPork:
There's checked add, so I'm thinking something like

```rust
self.recv_deadline(match Instant::now().checked_add(timeout) {
    Some(deadline) => deadline,
    None => Instant::now() + Duration::from_secs(86400 * 365 * 30) // hack from tokio-rs
})
```

taiki-e:
Sounds good to me.

taiki-e:
Fixed in crossbeam-channel 0.5.3.

MightyPork:
I was planning to make a PR xD

thanks for fixing it, this was quite the footgun :+1:

ibraheemdev:
I fixed this in the PR to std with:
```diff
    pub fn recv_timeout(&self, timeout: Duration) -> Result<T, RecvTimeoutError> {
-        self.recv_deadline(Instant::now() + timeout)
+        match Instant::now().checked_add(timeout) {
+            Some(deadline) => self.recv_deadline(deadline),
+            // So far in the future that it's practically the same as waiting indefinitely.
+            None => self.recv().map_err(RecvTimeoutError::from),
+        }
    }
```

PR #953, opened 2023-01-18
Wait forever on very large timeout

taiki-e:
Adopts the same approach as the standard library: https://github.com/crossbeam-rs/crossbeam/issues/795#issuecomment-1069514998

r? @ibraheemdev
bors d=ibraheemdev


taiki-e:
bors d=ibraheemdev

Seems bors did not read the PR comments?

taiki-e:
bors retry

PR #958, opened 2023-02-02
Mark some structures with `#[clippy::has_significant_drop]`

c410-f3r:
`#[clippy::has_significant_drop]` tells that a structure should be considered when evaluating some lints. Examples of such behavior are the existent `clippy::significant_drop_in_scrutinee` and the soon-to-be-finished https://github.com/rust-lang/rust-clippy/issues/9399.


taiki-e (approved):
Thanks!

bors r+

PR #967, opened 2023-03-04
CachePadded: Use 32-byte alignment on riscv32/sparc/hexagon, 16-byte alignment on m68k

taiki-e:
- riscv32 is assumed not to exceed the cache line size of riscv64.
- sparc, hexagon, and m68k's cache line sizes are from Linux kernel's __cacheline_aligned ([SMP_CACHE_BYTES, or L1_CACHE_BYTES if SMP_CACHE_BYTES is not set](https://github.com/torvalds/linux/blob/3516bd729358a2a9b090c1905bd2a3fa926e24c6/include/linux/cache.h)).

### DIFF
--- a/crossbeam-utils/src/cache_padded.rs
+++ b/crossbeam-utils/src/cache_padded.rs
@@ -14,7 +14,8 @@ use core::ops::{Deref, DerefMut};
 /// Cache lines are assumed to be N bytes long, depending on the architecture:
 ///
 /// * On x86-64, aarch64, and powerpc64, N = 128.
-/// * On arm, mips, mips64, and riscv64, N = 32.
+/// * On arm, mips, mips64, riscv32, riscv64, sparc, and hexagon, N = 32.
+/// * On m68k, N = 16.
 /// * On s390x, N = 256.
 /// * On all others, N = 64.
 ///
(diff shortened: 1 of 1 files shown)


taiki-e:
> linux's __cacheline_aligned ([SMP_CACHE_BYTES, or L1_CACHE_BYTES if SMP_CACHE_BYTES is not set](https://github.com/torvalds/linux/blob/3516bd729358a2a9b090c1905bd2a3fa926e24c6/include/linux/cache.h))

I'm aware of the values defined by the Linux kernel and those defined by us are different for some architectures (e.g., [aarch64](https://github.com/torvalds/linux/blob/0988a0ea791999ebbf95693f2676381825b05033/arch/arm64/include/asm/cache.h#L8), [riscv](https://github.com/torvalds/linux/blob/0988a0ea791999ebbf95693f2676381825b05033/arch/riscv/include/asm/cache.h#L10) [^1]), but for now, I have preferred the source we are currently referring to.

[^1]: As far as I know, the actual cache line size for both aarch64 and riscv is implementation dependent (like other architectures such as [x86](https://github.com/torvalds/linux/blob/0988a0ea791999ebbf95693f2676381825b05033/arch/x86/include/asm/cache.h#L8), [arm](https://github.com/torvalds/linux/blob/0988a0ea791999ebbf95693f2676381825b05033/arch/arm/include/asm/cache.h#L8), [powerpc](https://github.com/torvalds/linux/blob/0988a0ea791999ebbf95693f2676381825b05033/arch/powerpc/include/asm/cache.h#L9-L24), [mips](https://github.com/torvalds/linux/blob/0988a0ea791999ebbf95693f2676381825b05033/arch/mips/include/asm/cache.h#L14), [xtensa](https://github.com/torvalds/linux/blob/0988a0ea791999ebbf95693f2676381825b05033/arch/xtensa/include/asm/cache.h#L16), [loongarch](https://github.com/torvalds/linux/blob/0988a0ea791999ebbf95693f2676381825b05033/arch/loongarch/include/asm/cache.h#L8))

PR #972, opened 2023-04-08
channel: unbounded: synchronize receiver disconnect with initialization 

### LINKED ISSUE #971: Segfault in `crossbeam_channel::flavors::list`
The unbounded channel implementation from `crossbeam_channel` can segfault due to what looks like a race condition when dropping channel senders and receivers.

Since that also affects the `std::sync::mpsc` port, I also opened an issue in the rust-lang repo: https://github.com/rust-lang/rust/issues/110001. Please refer to this one for the details.

petrosagg:
Receiver disconnection relies on the incorrect assumption that `head.index != tail.index` implies that the channel is initialized (i.e `head.block` and `tail.block` point to allocated blocks). However, it can happen that `head.index != tail.index` and `head.block == null` at the same time which leads to a segfault when a channel is dropped in that state.

This can happen because initialization is performed in two steps. First, the tail block is allocated and the `tail.block` is set. If that is successful `head.block` is set to the same pointer. Importantly, initialization is skipped if `tail.block` is not null.

Therefore we can have the following situation:

1. Thread A starts to send the first value of the channel, observes that `tail.block` is null and begins initialization. It sets `tail.block` to point to a newly allocated block and then gets preempted. `head.block` is still null at this point.
2. Thread B starts to send the second value of the channel, observes that `tail.block` *is not* null and proceeds with writing its value in the allocated tail block and sets `tail.index` to 1.
3. Thread B drops the receiver of the channel which observes that `head.index != tail.index` (0 and 1 respectively), therefore there must be messages to drop. It starts traversing the linked list from `head.block` which is still a null pointer, leading to a segfault.

The first commit of this PR includes a reproduction of this scenario.

This PR fixes this problem by waiting for initialization to complete when `head.index != tail.index` and the `head.block` is still null. A similar check exists in `start_recv` for similar reasons.

Fixes #971

### DIFF
--- a/crossbeam-channel/src/flavors/list.rs
+++ b/crossbeam-channel/src/flavors/list.rs
@@ -236,6 +236,7 @@ impl<T> Channel<T> {
                     .compare_exchange(block, new, Ordering::Release, Ordering::Relaxed)
                     .is_ok()
                 {
+                    std::thread::sleep(std::time::Duration::from_secs(2));
                     self.head.block.store(new, Ordering::Release);
                     block = new;
                 } else {
--- a/crossbeam-channel/src/flavors/list.rs
+++ b/crossbeam-channel/src/flavors/list.rs
@@ -236,7 +236,6 @@ impl<T> Channel<T> {
                     .compare_exchange(block, new, Ordering::Release, Ordering::Relaxed)
                     .is_ok()
                 {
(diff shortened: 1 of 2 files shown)


taiki-e (approved):
Thanks!

bors r+

PR #976, opened 2023-04-13
skiplist: add compare_insert

Tianion:
The purpose is support `compare and swap` / `compare and exchange`.
I notice that the function `insert_internal` is core function, so I change `replace` to support CAS


Tianion:
hhh...`CI / san (pull_request) Cancelled after 360m`. There seems no problem with this commit
example
```rust
let map = SkipMap::new();
map.insert("key", 1);
map.compare_insert("key", 0, |x| x < &0);
assert_eq!(*map.get("key").unwrap().value(), 1);
map.compare_insert("key", 2, |x| x < &2);
assert_eq!(*map.get("key").unwrap().value(), 2);
```

taiki-e:
Thanks for the PR! CI failure should be fixed by rebasing.

taiki-e:
Could you also add a test for a case where the key is not present? (should work the same as `insert`)

Also, it would be nice to clarify in the documentation that the closure will not be called if the key is not present.

Tianion:
The rustc version in the CI has been upgraded from 1.71 to 1.72, which resulted in CI/Clippy checks failing. I attempted to fix it, but there are certain issues below that I'm unsure how to address. 
As a workaround, I had to use `#[allow(clippy::arc_with_non_send_sync)]` to pass the `CI/clippy`. I tried replacing Arc with Rc, but this would likely introduce other problems.
```
error: usage of an `Arc` that is not `Send` or `Sync`
  --> crossbeam-epoch/src/collector.rs:32:21
   |
32 |             global: Arc::new(Global::new()),
   |                     ^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: the trait `Sync` is not implemented for `Global`
   = note: required for `Arc<Global>` to implement `Send` and `Sync`
   = help: consider using an `Rc` instead or wrapping the inner type with a `Mutex`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#arc_with_non_send_sync
   = note: `-D clippy::arc-with-non-send-sync` implied by `-D warnings`
```

taiki-e:
Could you remove clippy-related commits? They have been addressed in #1021.

Tianion:
> Could you remove (or revert) clippy-related commits? They have been addressed in #1021.

Done，thanks for your detailed review

taiki-e (approved):
Thanks!

PR #982, opened 2023-05-25
chore: fix CI

Jules-Bertholet:
CI needs to be appeased, it seems.


Jules-Bertholet:
It seems we are at an impasse… old and new `rustc` can't be kept happy at the same time.

taiki-e (approved):
Thanks!

bors r+

Issue #30, opened 2016-01-20
Feature request: `peek()`


jedisct1:
Is there a way to get data from a queue without actually removing it from the queue?

aturon:
Unfortunately, there's no safe way to do this: if you hand out a reference to the data, another thread could gain ownership of that same data by dequeuing it.

PR #1013, opened 2023-08-10
Draft: Add peek to ArrayQueue<T> where T: Copy

estk:
Hi 👋 , I'm not sure if there is appetite in general for this feature but I personally found it useful. I would be happy to add the symmetric method to `SegQueue` as well. I noticed issue #30 but I figured it does not apply to this specialization. Also happy to just leave this be if it's not useful to anyone.


estk:
So after a bit of rework, I added a custom spinlock and was able to make miri happy locally. I'm skeptical if this is something that  the crossbeam team will really be interested in since it requires an extra spinlock which will cause contention between `.peek()` and `.push()` operations. Nonetheless I would be interested in feedback.

estk:
yep thanks for the consideration.

Issue #859, opened 2022-07-05
`AtomicCell::load()` can return a value that was modified by a concurrent racing `store()` inside the loader's critical section


cbeuw:
It is well known that SeqLocks - the underlying algorithm of `AtomicCell` - rely on data races which is technically always UB, though in practice it is assumed that a volatile read not ordered with a write is fine to perform, only that the read value must be discarded if it is determined after-the-fact that a writer was concurrently running with the volatile read.

However, crossbeam's SeqLock cannot reliably detect the presence of a writer running concurrently while a reader is in its  critical section, this is due to the atomic load of the timestamp in `validate_read()` being too weakly ordered/synchronised, and can miss a writer's modification to the timestamp.

The following test fails with a modified Miri. We need to stop Miri from eagerly throwing a data-race UB when the unordered volatile read or write happens (by simply changing the throw ub macro to println), since we need to keep it running. You can clone the modified Miri here: https://github.com/cbeuw/miri/tree/no-throw-on-data-race. Running `./miri install` will compile and install Miri and cargo-miri binaries under a new rustup toolchain called `miri`

`MIRIFLAGS="-Zmiri-seed=1" cargo +miri miri test --lib -- atomic::seq_lock::tests::test_racy_read_discarded`
```rust
  use core::ptr;
  use std::thread::{self, yield_now};

  #[derive(Copy, Clone)]
  struct EvilSend<T>(T);

  unsafe impl<T> Send for EvilSend<T> {}
  unsafe impl<T> Sync for EvilSend<T> {}

  #[test]
  fn test_racy_read_discarded() {
      static LK: SeqLock = SeqLock::new();

      let mut a = 0u32;
      let b = ptr::addr_of_mut!(a);
      let shared = EvilSend(b);

      let writer = thread::spawn(move || {
          // Same as AtomicCell::atomic_store()
          let _guard = LK.write();
          unsafe { ptr::write(shared.0, 1) };
      });

      {
          // Same as AtomicCell::atomic_load()
          if let Some(stamp) = LK.optimistic_read() {
              if stamp != 0 {
                  // Writer has already run
                  return;
              }

              // We entered the critical section before the writer did, so
              // *shared.0 == 0 (although we shouldn't be asserting it here since
              // we could be technically using a value obtained from a data-racing read)

              for _ in 0..100 {
                  // Give the writer some chance to ram in...
                  yield_now();
              }

              // If the writer did come into our critical section, this val will be 1
[...]

Issue #744, opened 2021-09-07
AtomicU8 is stable (and supported by the MSRV)


steffahn:
Apparently, crossbeam has MSRV `1.36`, which [includes `AtomicU8`](https://doc.rust-lang.org/1.36.0/std/sync/atomic/struct.AtomicU8.html). Hence the following comment is outdated:

https://github.com/crossbeam-rs/crossbeam/blob/2653a6cfe7f8756d27f53d51cc84fcabf5d1d549/crossbeam-utils/src/atomic/atomic_cell.rs#L796-L801

Considering that this comment legitimizes code that is officially UB, I propose that this should be fixed, if possible.

nico-abram:
This very naive patch passes the tests:

```patch
-                let val = ptr::read_volatile(src);
+                let mut val = std::mem::MaybeUninit::uninit();
+                let mut val_bytes = val.as_mut_ptr() as *mut u8;
+                let mut src_bytes = src as *mut core::sync::atomic::AtomicU8;
+                let src_end = src_bytes.offset(std::mem::size_of::<T>() as isize);
+                let val = unsafe {
+                    while src_bytes != src_end {
+                            val_bytes.write((*src_bytes).load(Ordering::Relaxed));
+                            val_bytes = val_bytes.offset(1);
+                            src_bytes = src_bytes.offset(1);
+                    }
+                    val.assume_init()
+                };
```

We could try using larger loads like thomcc's tearcell does: https://github.com/thomcc/tearor/blob/c558f6105230a49bb79b8afe989194e7656fdda7/src/lib.rs#L122-L162 instead of a naive byte by byte copy.

However, when I asked for review in the community discord server's #dark-arts a problem with padding bytes was mentioned: reading uninit/padding memory with AtomicU8 is not documented to not be UB. No way to fix this without a new language primitive or the language allowing these reads for atomics. For example, for this program
```rs
use std::mem::MaybeUninit;
use std::sync::atomic::{AtomicU8, Ordering};

fn main() {
    #[repr(align(2))]
    #[derive(Debug)]
    struct WithPadding {
        _d: u8,
    }
    let mut x = WithPadding { _d: 3u8 };
    let x_p = std::ptr::addr_of_mut!(x);
    let mut y = MaybeUninit::<WithPadding>::uninit();
    unsafe {
        *y.as_mut_ptr().cast::<u8>() = (*x_p.cast::<AtomicU8>()).load(Ordering::Relaxed);
[...]

PR #1015, opened 2023-08-13
AtomicCell: Use atomic-maybe-uninit

### LINKED ISSUE #315: Soundness of AtomicCell::compare_exchange is dubious

### LINKED ISSUE #748: `AtomicCell` accessing uninitialized memory

taiki-e:
This fixes UB when lock-free and the value has uninitialized bytes, by using [atomic-maybe-uninit](https://github.com/taiki-e/atomic-maybe-uninit) crate.
Fixes #315
Fixes #748

Also, this makes 128-bit atomics lock-free on some platforms.
Closes https://github.com/crossbeam-rs/rfcs/issues/13

Currently, this patch always uses fallback (seqlock) on environments that do not support inline assembly, such as Miri.
It is possible to use atomic-maybe-uninit crate to fix UB of concurrent access with volatile read/write in seqlock/deque (#744, #859), but that is not included in this PR and is left to subsequent PRs.

---

cc @RalfJung

### DIFF
--- a/crossbeam-utils/src/atomic/atomic_cell.rs
+++ b/crossbeam-utils/src/atomic/atomic_cell.rs
@@ -99,6 +99,8 @@ impl<T> AtomicCell<T> {
     /// # Examples
     ///
     /// ```
+    /// # // Always use fallback for now on environments that do not support inline assembly.
+    /// # if cfg!(any(miri, crossbeam_loom, crossbeam_atomic_cell_force_fallback)) { return; }
     /// use crossbeam_utils::atomic::AtomicCell;
     ///
     /// // This type is internally represented as `AtomicUsize` so we can just use atomic
@@ -307,13 +309,29 @@ macro_rules! atomic {
         loop {
             atomic!(@check, $t, AtomicUnit, $a, $atomic_op);
 
-            atomic!(@check, $t, atomic::AtomicU8, $a, $atomic_op);
-            atomic!(@check, $t, atomic::AtomicU16, $a, $atomic_op);
-            atomic!(@check, $t, atomic::AtomicU32, $a, $atomic_op);
-            #[cfg(target_has_atomic = "64")]
-            atomic!(@check, $t, atomic::AtomicU64, $a, $atomic_op);
-            // TODO: AtomicU128 is unstable
-            // atomic!(@check, $t, atomic::AtomicU128, $a, $atomic_op);
+            // Always use fallback for now on environments that do not support inline assembly.
+            #[cfg(not(any(
+                miri,
+                crossbeam_loom,
+                crossbeam_atomic_cell_force_fallback,
+            )))]
+            atomic_maybe_uninit::cfg_has_atomic_cas! {
+                atomic_maybe_uninit::cfg_has_atomic_8! {
+                    atomic!(@check, $t, atomic_maybe_uninit::AtomicMaybeUninit<u8>, $a, $atomic_op);
+                }
+                atomic_maybe_uninit::cfg_has_atomic_16! {
+                    atomic!(@check, $t, atomic_maybe_uninit::AtomicMaybeUninit<u16>, $a, $atomic_op);
+                }
+                atomic_maybe_uninit::cfg_has_atomic_32! {
+                    atomic!(@check, $t, atomic_maybe_uninit::AtomicMaybeUninit<u32>, $a, $atomic_op);
+                }
+                atomic_maybe_uninit::cfg_has_atomic_64! {
+                    atomic!(@check, $t, atomic_maybe_uninit::AtomicMaybeUninit<u64>, $a, $atomic_op);
+                }
+                atomic_maybe_uninit::cfg_has_atomic_128! {
+                    atomic!(@check, $t, atomic_maybe_uninit::AtomicMaybeUninit<u128>, $a, $atomic_op);
+                }
+            }
 
             break $fallback_op;
         }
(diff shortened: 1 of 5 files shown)


taiki-e:
Currently, it is not possible to use MaybeUninit in the input registers of an inline assembly (i.e., when processing MaybeUninit, values must be [passed through memory](https://github.com/taiki-e/atomic-maybe-uninit/blob/main/src/arch/aarch64.rs#L121-L122)), so this worsens performance.

Before:
```
test concurrent_load_u8     ... bench:     600,658 ns/iter (+/- 11,986)
test concurrent_load_usize  ... bench:     601,385 ns/iter (+/- 11,719)
test concurrent_store_u8     ... bench:   1,358,933 ns/iter (+/- 192,961)
test concurrent_store_usize  ... bench:   1,356,045 ns/iter (+/- 20,230)
```

After (pass by memory):

```
test concurrent_load_u8     ... bench:     868,554 ns/iter (+/- 150,633)
test concurrent_load_usize  ... bench:     851,380 ns/iter (+/- 7,139)
test concurrent_store_u8     ... bench:   2,029,710 ns/iter (+/- 178,712)
test concurrent_store_usize  ... bench:   2,038,341 ns/iter (+/- 196,265)
```

If I change the atomic-maybe-uninit implementation to pass by value using transmute, performance will be equivalent to the current one, but it is UB when the input value contains uninitialized bytes (because it is transmuted to integer).

After (pass by value):

```
test concurrent_load_u8     ... bench:     601,504 ns/iter (+/- 12,824)
test concurrent_load_usize  ... bench:     600,052 ns/iter (+/- 15,885)
test concurrent_store_u8     ... bench:   1,357,697 ns/iter (+/- 211,465)
test concurrent_store_usize  ... bench:   1,357,504 ns/iter (+/- 252,722)
```

It would be nice if the inline assembly could allow MaybeUninit registers as thomcc had [previously proposed](https://rust-lang.zulipchat.com/#narrow/stream/216763-project-inline-asm/topic/MaybeUninit.20in.20asm!).

taiki-e:
Filed https://github.com/rust-lang/rust/pull/114790 to allow MaybeUninit in input and output of inline assembly.

crossbeam-utils/src/atomic/atomic_cell.rs
```diff
@@ -691,124 +928,275 @@ macro_rules! impl_arithmetic {
...
-    }
+atomic_maybe_uninit::cfg_has_atomic_64! {
+    #[cfg(not(crossbeam_no_atomic_64))]
+    impl_arithmetic!(u64, atomic::AtomicU64, "let a = AtomicCell::new(7u64);");
+    #[cfg(not(crossbeam_no_atomic_64))]
+    impl_arithmetic!(i64, atomic::AtomicI64, "let a = AtomicCell::new(7i64);");
+    #[cfg(crossbeam_no_atomic_64)]
+    impl_arithmetic!(u64, fetch_update, "let a = AtomicCell::new(7u64);");
+    #[cfg(crossbeam_no_atomic_64)]
+    impl_arithmetic!(i64, fetch_update, atomic::AtomicI64, "let a = AtomicCell::new(7i64);");
```
RalfJung, line 966: With `fetch_update` it doesn't take a type, does it?

Also there's a subtle invariant here that `fallback` is only used if regular loads/stores also use the fallback, right? How is that ensured?

taiki-e:
Since this PR was opened we have made many improvements to both atomic-maybe-uninit and the compiler. Resolving performance issue (https://github.com/rust-lang/rust/pull/114790), improving and stabilizing inline assembly support on tier 2 targets (https://github.com/rust-lang/rust/pull/131258, https://github.com/rust-lang/rust/pull/131781, etc.), fixing or avoiding rustc/LLVM bugs, improving code quality (https://github.com/taiki-e/atomic-maybe-uninit/pull/30, etc.), improving platform support (atomic-maybe-uninit currently supports all CPU architectures supported by Rust other than C-SKY), etc.

I think we can now finally merge this.

PR #1016, opened 2023-08-14
add support for csky arch

Dirreke:
add support for csky-unknown-linux-gnuabiv2


taiki-e:
FWIW, when a new builtin target is added, our CI handles it automatically: https://github.com/crossbeam-rs/crossbeam/pull/1017

Dirreke:
Thanks!

PR #1024, opened 2023-08-28
Add StaticArrayQueue, a queue that can be used as a global static variable

Finomnis:
# Motivation

Currently there is no queue (that I could find) that fulfills all the following criteria:
- `no_std` compatible (without `alloc`)
- lock-free
- usable as a static variable

`crossbeam_queue::ArrayQueue` is very close, but has the following problems:
- Not usable as a static variable (`new()` function isn't `const`)
- Requires `alloc`

# Content

This change introduces `StaticArrayQueue`, which is identical to `ArrayQueue` except that it contains the array unboxed as a member and has a `const` constructor.

This allows the following pattern:
```rust
use std::{
    thread::{self, sleep},
    time::Duration,
};

use crossbeam_queue::StaticArrayQueue;

static QUEUE: StaticArrayQueue<u32, 10> = StaticArrayQueue::new();

fn main() {
    thread::spawn(|| loop {
        while let Some(element) = QUEUE.pop() {
            println!("Got element: {}", element);
        }
        sleep(Duration::from_millis(1));
    });

    for i in 0..10 {
        QUEUE.push(i).unwrap();
        sleep(Duration::from_millis(10));
    }
}
```

# Open questions/tasks

- [ ] Const generics are not supported by Rust `1.38`. Bump MSRV? Publish as external crate? (in its current state, the code requires Rust `1.61`. So most likely publish this as an external crate)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,9 +40,9 @@ jobs:
       matrix:
         # aarch64/x86_64 macOS and aarch64 Linux are tested on Cirrus CI
         include:
-          - rust: '1.57'
+          - rust: '1.59'
             os: ubuntu-latest
-          - rust: '1.57'
+          - rust: '1.59'
             os: windows-latest
           - rust: stable
             os: ubuntu-latest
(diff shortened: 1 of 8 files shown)


Finomnis:
Never mind. It seems like this problem is already solved by [`heapless::mpmc`](https://docs.rs/heapless/latest/heapless/mpmc/index.html). Will use that one in my project instead.

Issue #821, opened 2022-05-04
crossbeam-channel uses pure userspace spinlocks, which perform very poorly in some situations


bsilver8192:
crossbeam-channel appears to use pure spinlocks (which never block in the kernel), from userspace, which is a bad idea. Looking at the behavior with `strace`, it appears to be `sched_yield`ing in a loop.

On my Linux system (Debian 5.15 kernel, standard configuration) with a simple little benchmark I wrote, I see this:
```console
brian[13116] rico (master) ~/Desktop/crossbeam
$ cargo run --manifest-path crossbeam-channel/benchmarks/Cargo.toml --bin crossbeam-channel
[warnings because I commented stuff out elided]
warning: `benchmarks` (bin "crossbeam-channel") generated 7 warnings
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/crossbeam-channel`
unbounded_spsc_sync       Rust crossbeam-channel   4.473 sec
brian[13117] rico (master) ~/Desktop/crossbeam
$ taskset 1 cargo run --manifest-path crossbeam-channel/benchmarks/Cargo.toml --bin crossbeam-channel
[warnings because I commented stuff out elided]
warning: `benchmarks` (bin "crossbeam-channel") generated 7 warnings
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/crossbeam-channel`
unbounded_spsc_sync       Rust crossbeam-channel 106.810 sec
```

It takes over 20x longer with both threads pinned to the same CPU, because each thread spins until its scheduler tick expires repeatedly. In my full application, I'm using SCHED_RR in one thread, which makes this a full-on deadlock because it will never yield and let the other thread run.

I'm running https://github.com/bsilver8192/crossbeam/tree/crossbeam-channel-spin-benchmark, which is recent master (450d237d1546ac135d21f75b61fdf5a944e8421c) + [my benchmark](https://github.com/bsilver8192/crossbeam/commit/d11015fd09a3145b31fc97a525e61141726be655).

This seems related to #366 and #447.

[...]

PR #1038, opened 2023-11-07
Reduce backoff spinning/usage

ibraheemdev:
This PR is a first step to reducing a lot of the spinning-related issues that have cropped up. It does two things:
- Remove optimistic spinning completely from `Context::wait_until`. There are already spin loops in `recv`/`send`, so another set of spins after entering the park queue is excessive.
- Reduce the amount of spinning in `Backoff::snooze`. As seen in https://github.com/crossbeam-rs/crossbeam/issues/821#issuecomment-1776524621, the amount of `PAUSE` instructions issued is quite high. This change reduces it from 6 to 3, similar to [parking-lot](https://github.com/Amanieu/parking_lot/blob/fa859df5893f28df3e7b32224bb237a0efaf71a2/core/src/spinwait.rs#L53-L57). It also reduces the amount of total amount of spinning, reducing `YIELD_LIMIT` to 6. This differs from parking-lot as it is common for a channel, unlike a mutex, to remain empty/full for long periods of times, and so if yielding doesn't allow progress after a few times it is likely better to park properly.

### DIFF
--- a/crossbeam-channel/src/context.rs
+++ b/crossbeam-channel/src/context.rs
@@ -138,21 +138,6 @@ impl Context {
     /// If the deadline is reached, `Selected::Aborted` will be selected.
     #[inline]
     pub fn wait_until(&self, deadline: Option<Instant>) -> Selected {
-        // Spin for a short time, waiting until an operation is selected.
-        let backoff = Backoff::new();
-        loop {
-            let sel = Selected::from(self.inner.select.load(Ordering::Acquire));
-            if sel != Selected::Waiting {
-                return sel;
-            }
-
-            if backoff.is_completed() {
-                break;
-            } else {
-                backoff.snooze();
-            }
-        }
-
         loop {
             // Check whether an operation has been selected.
             let sel = Selected::from(self.inner.select.load(Ordering::Acquire));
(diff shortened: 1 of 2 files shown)


ibraheemdev:
I'll try to post some relevant benchmark/profiling results to see how much this helps.

taiki-e:
Thanks! As for SPIN_LIMIT, I remember that 4 looked good when I looked into it before. https://github.com/taiki-e/crossbeam/commit/0362ffe0f5c7febcf9b8dac9db61547fd421546b

That said, either is fine if performance is not that different.

codehobbyist06:
Hi @ibraheemdev , any update on this PR? When are the chances that this PR could be merged?

taiki-e (approved):
Thanks!

ibraheemdev:
I ran the benchmarks locally and there are some large performance regressions, especially on the spsc benchmarks. However, I'm not we should be optimizing for those benchmarks as opposed to the real world cases where excessive spinning is problematic. That said I'd like to experiment a bit more, maybe just removing the double use of `Backoff` and the yield limit is enough?

taiki-e:
The removal of double use is definitely a non-controversial change. I think the others have different optimal solutions depending on the nature of the benchmark and the platform. I think splitting this PR and merging less controversial changes first is one of the reasonable solutions here.

PR #1040, opened 2023-11-25
Add select_biased! macro

ryoqun:
~(this is a draft; there's too much of code duplication...)~

much like https://github.com/rust-lang/futures-rs/pull/1976, i want biased select in `crossbeam-channel` as well. :)

PR #1047, opened 2023-12-08
Add disconnected flavor which immediately fails

ryoqun:
Yet another small niceties, i'd like to add to `crossbeam_channel`.

This doesn't look meaningful by itself at first glance. However, this is useful for conditionally inserting an artificial select arm at predetermined priority in the `select_biased!()` #1040, combined with `never()` like this:

```rust
select_biased! {
    recv(important_channel) -> urgent_message => {
        ...
    },
    recv(if is_local_task_available { disconnected() } else { never() }) -> _dummy_signal => {
        ...
    },
    recv(normal_channel) -> normal_message => {
        ...
    },
}
```

Note that using `tick()` like below introduces needless calls to `Instant::now()` (which is a vdso basically) _per select invocation_ on top of the needless stack memory consumption. (On the other hand, the zst & `const`-code-heavy `disconnected()` will be optimized out to great degree after the macro expansion)..:

```rust
fn always() -> Receiver<Instant> {
    tick(Duration::default())
}

```

Also, unrolling the select is possible, but creates code duplication. Lastly, checking `is_local_task_available` after `select_biased!()` alone would introduce a priority inversion.

Thus, i think there's no acceptable alternative solution currently.

### DIFF
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -274,6 +274,13 @@ pub fn never<T>() -> Receiver<T> {
     }
 }
 
+/// Creates a receiver that is disconnected always.
+pub fn disconnected<T>() -> Receiver<T> {
+    Receiver {
+        flavor: ReceiverFlavor::Disconnected(flavors::disconnected::Channel::new()),
+    }
+}
+
 /// Creates a receiver that delivers messages periodically.
 ///
 /// The channel is bounded with capacity of 1 and never gets disconnected. Messages will be
@@ -731,6 +738,9 @@ enum ReceiverFlavor<T> {
 
     /// The never flavor.
     Never(flavors::never::Channel<T>),
+
+    /// The disconnected flavor.
+    Disconnected(flavors::disconnected::Channel<T>),
 }
 
 unsafe impl<T: Send> Send for Receiver<T> {}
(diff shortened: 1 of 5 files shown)


crossbeam-channel/src/channel.rs
```diff
@@ -274,6 +274,13 @@ pub fn never<T>() -> Receiver<T> {
     }
 }
 
+/// Creates a receiver that is disconnected always.
```
ryoqun, line 277: I'm happy to expand this doccomment with an example, which is in the pr description with polishes, once it's confirmed that this pr has some hope to be merged.

crossbeam-channel/src/flavors/disconnected.rs
```diff
@@ -0,0 +1,108 @@
...
+/// This flavor doesn't need a token.
+pub(crate) type DisconnectedToken = ();
+
+/// Channel that always delivers messages.
+pub(crate) struct Channel<T> {
+    _marker: PhantomData<T>,
+}
+
+impl<T> Channel<T> {
+    /// Creates a channel that always delivers messages.
```
ryoqun, line 21: oops forgot to update..

taiki-e:
Thanks for the PR. I guess this could be replaced by never + disconnect once the API for explicit disconnection is exposed?

ryoqun:
> Thanks for the PR. I guess this could be replaced by never + disconnect once the API for explicit disconnection is exposed?

hey, really thanks for replying!

yeah, assuming the disconnected never is implemented in a hard-coded way without any state much like the never itself for maximum optimization opportunity by rustc.

ryoqun:
i think _disconnected_ dummy sender might also be useful. 

how about this api? `let (sender, receiver) = disconnected()`? (i.e. alight with `unbounded()` and `bounded(...)`)

PR #1065, opened 2023-12-29
ShardedLock doesn't need shards to be boxed

noelzubin:


taiki-e:
Thanks for the PR.

- This will probably reduce the performance of move, etc. as the size will be very large.
- When going through Vec, there is not much advantage to this as the code will eventually do the allocation unless the compiler can remove the allocation with optimization; it is more reasonable to raise the MSRV to use const rwlock new, but it is not clear how desirable this really is given the above points.

However, it might be reasonable to use `Box<[_;NUM_SHARDS]>` (1 pointer) instead of `Box<[_]>` (1 pointer + 1 usize metadata) since it would reduce the size.

noelzubin:
Ah. yes . What about using MaybeUnint ?.

PR #1069, opened 2024-01-07
Move CachePadded to separate sub-crate

### LINKED ISSUE #1066: Move CachePadded to separate sub-crate?
I have been using https://github.com/smol-rs/cache-padded/ as a dependency until it has been deprecated (https://github.com/smol-rs/cache-padded/pull/11) and then, as suggested, I switched to `crossbeam_utils::CachePadded`.

I didn't really like that this forced an unneeded dependency (`cfg-if`, mentioned at https://github.com/smol-rs/cache-padded/issues/10#issuecomment-1237286495) onto my crate.

I kinda accepted this, but recently, I found another inconvenience: it also forces me to bump the MSRV of my crate due to #1037, which is AFAICT unrelated to `CachePadded`.

Would it be possible to move `CachePadded` to a separate sub-crate to avoid those problems?

Alternatively, both problems might be avoidable by moving parts of `crossbeam_utils` to separate "features" (which I could deactivate for my crate), but I don't know if that's feasible?

mgeier:
This is (for now) just a proof of concept to see whether moving `CachePadded` to a separate sub-crate is feasible.

Fixes #1066, which contains related discussion.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ std = [
 alloc = ["crossbeam-epoch/alloc", "crossbeam-queue/alloc"]
 
 [dependencies]
+crossbeam-cachepadded = { version = "0.0", path = "crossbeam-cachepadded", default-features = false, optional = true }
 crossbeam-channel = { version = "0.5.10", path = "crossbeam-channel", default-features = false, optional = true }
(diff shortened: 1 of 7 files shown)


taiki-e:
Closing in favor of https://github.com/crossbeam-rs/crossbeam/pull/1075.

PR #1072, opened 2024-01-08
Remove dependency on cfg-if

taiki-e:
IMO, while cfg-if is powerful where there are complex mutual exclusion conditions such as "a or b or c or ...", its benefits are not great for simple use cases like ours.

There are also compatibility issues with tools such as rustfmt (and r-a IIRC?), and interoperability with doc_cfg, which is what I plan to use. The rustfmt issue affects not only the code within the macro, but also the module defined within the macro. It is why [tokio's contributing guideline recommends the direct use of rustfmt with git ls-files instead of cargo fmt](https://github.com/tokio-rs/tokio/blob/84c5674c601dfc36ab417ff0ec01763c2dd30a5c/CONTRIBUTING.md#cargo-commands).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,8 +40,6 @@ crossbeam-epoch = { version = "0.9.17", path = "crossbeam-epoch", default-featur
 crossbeam-queue = { version = "0.3.10", path = "crossbeam-queue", default-features = false, optional = true }
 crossbeam-utils = { version = "0.8.18", path = "crossbeam-utils", default-features = false }
 
-cfg-if = "1"
-
 [dev-dependencies]
 rand = "0.8"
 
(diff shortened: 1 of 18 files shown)


taiki-e:
```console
$ cargo tree -p crossbeam | sed -E 's/ \(.*//g'
crossbeam v0.8.3
├── crossbeam-channel v0.5.10
│   └── crossbeam-utils v0.8.18
├── crossbeam-deque v0.8.4
│   ├── crossbeam-epoch v0.9.17
│   │   └── crossbeam-utils v0.8.18
│   └── crossbeam-utils v0.8.18
├── crossbeam-epoch v0.9.17
├── crossbeam-queue v0.3.10
│   └── crossbeam-utils v0.8.18
└── crossbeam-utils v0.8.18
[dev-dependencies]
<omitted>
```

PR #1075, opened 2024-01-08
Create separate sub-crate for CachePadded (second attempt)

### LINKED ISSUE #1066: Move CachePadded to separate sub-crate?

mgeier:
This is an alternative to #1069.

This simply creates a symbolic link to `cache_padded.rs`.

This way, `crossbeam-utils` is unaffected.

Fixes #1066, which contains related discussion.

This is also a proof of concept, some details still need to be ironed out.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ workspace = true
 resolver = "2"
 members = [
   ".",
+  "crossbeam-cachepadded",
   "crossbeam-channel",
   "crossbeam-channel/benchmarks",
   "crossbeam-deque",
--- a/crossbeam-cachepadded/Cargo.toml
+++ b/crossbeam-cachepadded/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "crossbeam-cachepadded"
+# When publishing a new version:
+# - Update CHANGELOG.md
+# - Update README.md
+# - Create "crossbeam-cachepadded-X.Y.Z" git tag
+version = "0.0.0"
+edition = "2018"
+rust-version = "1.38"
+license = "MIT OR Apache-2.0"
+repository = "https://github.com/crossbeam-rs/crossbeam"
+homepage = "https://github.com/crossbeam-rs/crossbeam/tree/master/crossbeam-cachepadded"
+documentation = "https://docs.rs/crossbeam-cachepadded"
+description = "Prevent false sharing by padding and aligning to the length of a cache line"
+keywords = ["cache", "padding", "lock-free", "atomic"]
+categories = ["concurrency", "no-std"]
+
+[lib]
+# The tests are executed in the crossbeam-util crate.
+doctest = false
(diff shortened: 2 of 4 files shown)


mgeier:
Are there any comments on this?

The goal of this new sub-crate was to avoid unnecessary MSRV bumps (for now using version 1.38), but this conflicts with the CI check for MSRV, which checks all sub-crates at once and fails because Rust 1.38 doesn't know yet about edition 2021.
Any ideas how to fix this?

mgeier:
Just for reference: I have tried to come up with a hack that moves the maintenance burden to my crate: https://github.com/mgeier/rtrb/pull/115

It's ugly, but it seems to work.
It still would be great if this could work out-of-the-box with crossbeam.

taiki-e:
I'm thinking that if we put crossbeam-utils' atomic-related APIs behind a feature flag, adopt the approach I adopted in futures-rs (https://github.com/rust-lang/futures-rs/pull/2811), revert the edition bump (https://github.com/crossbeam-rs/crossbeam/commit/c5c1361e66897b56cfaa47c2f0f7a824cef11d2b, https://github.com/crossbeam-rs/crossbeam/commit/a57e655eef415c21babddc4ba0217b6ca7acd0a2), and disable unsafe_op_in_unsafe_fn in some crates, we can revert the MSRV bump, at least in crossbeam-utils and crossbeam-queue. Of course, the first one is a breaking change, but the master branch already has some unreleased breaking changes.

As for the currently proposed approach (adding a new CachePadded-only crate), I'm not sure it's worth having a new crate, considering that few users were using the cache-padded crate even before the deprecation.

mgeier:
Yes, that sounds great!

>  considering that few users were using the cache-padded crate even before the deprecation.

This might be true, but I still consider it a very important foundational crate.

It would be great to have a widely usable and unrestricted common crate for that instead of everyone copying (parts of) the code to their own crates (like e.g. https://github.com/tqtrungse/omango-util/blob/master/src/cache_padded.rs).

taiki-e:
Closing in favor of the idea mentioned in https://github.com/crossbeam-rs/crossbeam/pull/1075#issuecomment-2039050221 (implementation is in https://github.com/crossbeam-rs/crossbeam/pull/1156).

Thanks anyway for the PR.

PR #1077, opened 2024-01-15
array::Channel: remove redundant cap field

calebsander:
`cap` stores the channel capacity, the number of slots in `buffer`. There's no additional cost to extracting the length of a slice, so remove `cap` and use `buffer.len()` instead.

### DIFF
--- a/crossbeam-channel/src/flavors/array.rs
+++ b/crossbeam-channel/src/flavors/array.rs
@@ -73,9 +73,6 @@ pub(crate) struct Channel<T> {
     /// The buffer holding slots.
     buffer: Box<[Slot<T>]>,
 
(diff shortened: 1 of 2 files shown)


taiki-e:
Thanks! Could you do the same for ArrayQueue?
https://github.com/crossbeam-rs/crossbeam/blob/3aa2598e3943c791e68d668d9f92e84e748bab65/crossbeam-queue/src/array_queue.rs#L71

taiki-e (approved):

taiki-e:
IIRC, the main difference would be the presence of code for select and disconnect support. Aside from the latter, the former would be a complicated API to include.

taiki-e (approved):

PR #1078, opened 2024-01-15
counter: use NonNull

calebsander:
`counter::{Sender, Receiver}` hold shared references to a channel, similar to an `Arc`. Their raw pointers to the `Counter` are never null, so change them from `*mut Counter<C>` to `NonNull<Counter<C>>`. This makes it easier to verify the safety requirements for dereferences, and allows the niche value optimization to apply.


taiki-e (approved):
Thanks. Considering our use case, I don't think niche optimization will be used, but it would indeed somewhat clarify the code.

PR #1079, opened 2024-01-16
Fix lockfree benchmark output

Osteoporosis:
I am trying to run the lockfree benchmark and plot the results, but the output file has an incorrect name tag.
This causes the lockfree results to disappear. And if both lockfree and crossbeam-channel results coexist, the crossbeam-channel result may be overwritten and corrupted on the benchmark graph.


taiki-e (approved):
Thanks.

PR #1084, opened 2024-02-25
Fix race between block initialization and receiver disconnection

ibraheemdev:
Reported in https://github.com/rust-lang/rust/issues/121582. This fixes a race that can occur if a sender installs the first block right after all receivers have disconnected and checked for any existing messages, causing a memory leak.


taiki-e (approved):
Thanks!

PR #1089, opened 2024-02-29
Document potential memory leak in unbounded channel

ibraheemdev:
Per https://github.com/rust-lang/rust/pull/121646#discussion_r1503682759.


taiki-e (approved):
Thanks.

Issue #1090, opened 2024-03-05
crossbeam::skiplist: make iterator sendable to anther threads


SpadeA-Tang:
We are currently implementing an in-memory engine feature in TiKV, and use skiplist as the underlying engine. 
We need the iterator to be sendable among threads. 
However, the current skiplist implementation carries the borrow of the skiplist,  so the iterator can only be used in the thread creating it.
I implement a wrapper of the current skiplist, making it support creating iterators supporting the senario described above.
Would you mind reviewing it and seeing it is needed in crossbeam skiplist.

SpadeA-Tang:
@taiki-e  PTAL, thanks.

PR #1091, opened 2024-03-05
skiplist: a new iteator which is sendable among threads

SpadeA-Tang:
Ref https://github.com/crossbeam-rs/crossbeam/issues/1090

This PR implements a wrapper of the skiplist which supports creating iteator that is sendable among threads.
More tests or comment can be added if this is needed.

### DIFF
--- a/crossbeam-skiplist/src/base.rs
+++ b/crossbeam-skiplist/src/base.rs
@@ -1,6 +1,7 @@
 //! A lock-free skip list. See [`SkipList`].
 
 use alloc::alloc::{alloc, dealloc, handle_alloc_error, Layout};
+use alloc::sync::Arc;
 use core::borrow::Borrow;
 use core::cmp;
 use core::fmt;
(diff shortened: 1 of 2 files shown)


crossbeam-skiplist/src/base.rs
```diff
@@ -2121,3 +2122,252 @@ fn below_upper_bound<T: Ord + ?Sized>(bound: &Bound<&T>, other: &T) -> bool {
...
+    /// Returns a reference to the value.
+    pub fn value(&self) -> &V {
+        unsafe { &(*self.node).value }
+    }
+}
+
+/// A lock-free skip list.
+pub struct ConcurrentSkipList<K, V> {
+    /// The underlying skip list
+    list: Arc<SkipList<K, V>>,
```
taiki-e, line 2167: I think we could add a method that takes `self: Arc<Self>` instead of creating a new type. Like:

- https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#method.lock_owned
- https://docs.rs/async-lock/latest/async_lock/struct.Mutex.html#method.lock_arc

SpadeA-Tang: Thanks for the reply. It's a bit late here, I will see it tomorrow.

SpadeA-Tang: I have modified the code. PTAL again, thanks @taiki-e . If this is desired, I will add more tests.

crossbeam-skiplist/src/base.rs
```diff
@@ -2121,3 +2196,184 @@ fn below_upper_bound<T: Ord + ?Sized>(bound: &Bound<&T>, other: &T) -> bool {
...
+impl<K, V> Drop for OwnedEntry<K, V> {
+    fn drop(&mut self) {
+        assert!(self.released);
+    }
+}
+
+/// A iterator with a clone of the concurrent skip list
+pub struct OwnedIter<T, K, V>
+where
+    T: AsRef<SkipList<K, V>>,
```
taiki-e, line 2264: Why use a generic parameter here? There seems to be only one possible instance.

SpadeA-Tang: I forgot... I think it's unnecessary now. Have removed.

PR #1097, opened 2024-04-01
feat: impl Display for CachePadded<T>

### LINKED ISSUE #1043: Implement fmt::Display for CachePadded where T: Display
Would it be possible to implement Display for CachePadded<T> where T: Display.

This makes adding CachePadded around a variable slightly less intrusive a change.

joshua-mo-143:
Aims to solve #1043.

I have followed the Debug implementation. Some guidance may be required on whether we want to print the inner value or not.

Closes #1043

### DIFF
--- a/crossbeam-utils/src/cache_padded.rs
+++ b/crossbeam-utils/src/cache_padded.rs
@@ -207,3 +207,9 @@ impl<T> From<T> for CachePadded<T> {
         Self::new(t)
     }
 }
+
+impl<T: fmt::Display> fmt::Display for CachePadded<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Cache padding with length {}", &self.value)
+    }
+}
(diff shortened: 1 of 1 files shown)


crossbeam-utils/src/cache_padded.rs
```diff
@@ -207,3 +207,9 @@ impl<T> From<T> for CachePadded<T> {
         Self::new(t)
     }
 }
+
+impl<T: fmt::Display> fmt::Display for CachePadded<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Cache padding with length {}", &self.value)
```
taiki-e, line 213: I would have expected it to work the same way as when called via deref... Like:
```suggestion
        fmt::Display::fmt(&self.value, f)
```

cc @akanalytics

taiki-e (approved):
Thanks!

taiki-e:
Published in crossbeam-utils 0.8.20.

PR #1103, opened 2024-04-23
Implement Stealer::will_steal_from

james7132:
Currently it's not possible to tell if a Worker and a Stealer point to the same underlying queue. This make it difficult to tell which stealer to remove from a list and you have a local Worker on hand, or skip stealing from when given a global list of stealers. The implementation just forwards to Arc::ptr_eq on the inner buffer. Added a doctest as well.

### DIFF
--- a/crossbeam-deque/src/deque.rs
+++ b/crossbeam-deque/src/deque.rs
@@ -296,8 +296,8 @@ impl<T> Worker<T> {
     /// let w_1 = Worker::<i32>::new_lifo();
     /// let w_2 = Worker::<i32>::new_fifo();
     ///
-    /// let s_1 = w.stealer();
-    /// let s_2 = w.stealer();
+    /// let s_1 = w_1.stealer();
+    /// let s_2 = w_2.stealer();
     ///
     /// assert!(w_1.is_same_as(&s_1));
     /// assert!(w_2.is_same_as(&s_2));
--- a/crossbeam-deque/src/deque.rs
+++ b/crossbeam-deque/src/deque.rs
@@ -286,6 +286,28 @@ impl<T> Worker<T> {
         }
     }
 
+    /// Checks if the worker and the provided stealer are pointing to the same underlying queue.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use crossbeam_deque::Worker;
+    ///
+    /// let w_1 = Worker::<i32>::new_lifo();
+    /// let w_2 = Worker::<i32>::new_fifo();
+    ///
+    /// let s_1 = w.stealer();
+    /// let s_2 = w.stealer();
+    ///
+    /// assert!(w_1.is_same_as(&s_1));
+    /// assert!(w_2.is_same_as(&s_2));
+    /// assert!(!w_1.is_same_as(&s_2));
+    /// assert!(!w_2.is_same_as(&s_1));
+    /// ```
+    pub fn is_same_as(&self, stealer: &Stealer<T>) -> bool {
+        Arc::ptr_eq(&self.inner, &stealer.inner)
+    }
+
     /// Resizes the internal buffer to the new capacity of `new_cap`.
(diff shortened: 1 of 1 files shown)


crossbeam-deque/src/deque.rs
```diff
@@ -284,6 +284,28 @@ impl<T> Worker<T> {
         }
     }
 
+    /// Checks if the worker and the provided stealer are pointing to the same underlying queue.
```
taiki-e, line 287: I feel it would be easier to understand documentation something like "returns true if the given stealer was derived from this worker".

crossbeam-deque/src/deque.rs
```diff
@@ -284,6 +284,28 @@ impl<T> Worker<T> {
...
+    ///
+    /// let s_1 = w_1.stealer();
+    /// let s_2 = w_2.stealer();
+    ///
+    /// assert!(w_1.is_same_as(&s_1));
+    /// assert!(w_2.is_same_as(&s_2));
+    /// assert!(!w_1.is_same_as(&s_2));
+    /// assert!(!w_2.is_same_as(&s_1));
+    /// ```
+    pub fn is_same_as(&self, stealer: &Stealer<T>) -> bool {
```
taiki-e, line 305: I don't think this method name is easy to understand, but I'm not sure what the good name would be.

taiki-e (commented):
Thanks for the PR! I'm positive about adding such a method.

crossbeam-deque/src/deque.rs
james7132:
```suggestion
    /// Returns true if the given stealer was derived from this worker.
```

crossbeam-deque/src/deque.rs
james7132: Not sure if `Worker::ptr_eq` is better. I can't seem to come up with a better name here either.

notgull: How about `Worker::donates_to`?

james7132: Went ahead and renamed the function as @notgull requested.

james7132:
@taiki-e, went ahead with the suggested rename.

crossbeam-deque/src/deque.rs
```diff
@@ -623,6 +623,30 @@ impl<T> Stealer<T> {
...
+    ///
+    /// let s_1 = w_1.stealer();
+    /// let s_2 = w_2.stealer();
+    ///
+    /// assert!(!s_1.will_steal_fromq(&w_1));
+    /// assert!(!s_2.will_steal_fromq(&w_2));
+    /// assert!(s_1.will_steal_fromq(&w_2));
+    /// assert!(s_2.will_steal_fromq(&w_1));
+    /// ```
+    pub fn will_steal_fromq(&self, worker: &Worker<T>) -> bool {
```
taiki-e, line 646:
```suggestion
    /// assert!(!s_1.will_steal_from(&w_1));
    /// assert!(!s_2.will_steal_from(&w_2));
    /// assert!(s_1.will_steal_from(&w_2));
    /// assert!(s_2.will_steal_from(&w_1));
    /// ```
    pub fn will_steal_from(&self, worker: &Worker<T>) -> bool {
```

crossbeam-deque/src/deque.rs
```diff
@@ -623,6 +623,30 @@ impl<T> Stealer<T> {
         b.wrapping_sub(f).max(0) as usize
     }
 
+    /// Checks if the stealer will steal from the provided worker.
+    ///
+    /// If both are pointing to the same underlying queue, this will return false.
```
taiki-e, line 628: Isn't that the reverse?
- `steal` steals from the underlying queue.
- `steal_batch` steals from the underlying queue and pushes to other worker.

taiki-e (changes requested):

Issue #997, opened 2023-06-17
crossbeam::channel::Receiver::try_recv can block forever if sending thread is blocked


benhansen-io:
Using unbounded channels, try_recv can block indefinitely if the sender is stopped at a particular point. This behavior is also exhibited in the std mpsc and in the issue there (rust-lang/rust#112723) I was asked to verify the issue also exists in crossbeam (it does). I'll update the backtraces and code for crossbeam below.

Running the following code (using the crossbeam feature):

```rust
use std::{
    sync::Arc,
    time::{Duration, Instant},
};

use thread_priority::ThreadPriority;

#[cfg(not(feature = "crossbeam"))]
use std::sync::mpsc::channel;
#[cfg(not(feature = "crossbeam"))]
use std::sync::mpsc::TryRecvError;

#[cfg(feature = "crossbeam")]
use crossbeam::channel::unbounded as channel;
#[cfg(feature = "crossbeam")]
use crossbeam::channel::TryRecvError;

fn main() {
    const PINNED_CORE: usize = 2;

    if cfg!(feature = "crossbeam") {
        println!("Using crossbeam::unbounded channels");
    } else {
        println!("Using std::sync::mpsc channels");
    }

    let (sender, receiver) = channel::<usize>();

    std::thread::Builder::new()
        .name("sending".to_owned())
        .spawn(move || {
            thread_priority::set_current_thread_priority(ThreadPriority::Min).unwrap();
            core_affinity::set_for_current(core_affinity::CoreId { id: PINNED_CORE });

            loop {
                sender.send(42).unwrap();
            }
        })
        .unwrap();

    let num_received = Arc::new(std::sync::atomic::AtomicUsize::new(0));

    std::thread::Builder::new()
        .name("receiving".to_owned())
        .spawn({
            let num_received = num_received.clone();
            move || {
                thread_priority::set_current_thread_priority(ThreadPriority::Max).unwrap();
                core_affinity::set_for_current(core_affinity::CoreId { id: PINNED_CORE });

                loop {
                    let start = Instant::now();
                    let try_receive_result = receiver.try_recv();
                    let elapsed = start.elapsed();
                    if elapsed > Duration::from_secs(1) {
                        println!("try_recv blocked for {:.2} seconds", elapsed.as_secs_f32());
                    }
                    match try_receive_result {
                        Ok(_) => {
                            num_received.fetch_add(1, std::sync::atomic::Ordering::SeqCst);
                        }
[...]

PR #1105, opened 2024-05-02
Fallback to Parking in Bounded and Unbounded Channels

### LINKED ISSUE #366: Yielding in crossbeam-channel

ibraheemdev:
Currently, the queues used by the bounded and unbounded channels (also `ArrayQueue` and `SegQueue`) are not lock-free. The linearization point of a send or receive is moving the tail or head index respectively, meaning that if a sender or receiver sees the head/tail has moved but the value has not been written yet, they must spin until the corresponding receive or send completes that frees up space in the channel. Blocking in such a case is not possible, as sends might notify a receiver even though they are not visible due to a lagging sender holding up the channel, or vice versa, leading to missed wakeups.

The current solution of unbounded spinning can lead to issues, especially on platforms with a single core or custom thread priorities. One way to allow falling back to blocking is if waiters that see more values in the channel after being woken up notify any other waiters, making up for any missed notifications. However, this ends up being quite expensive and can lead to many unnecessary wakeups. A solution to this is wakeup throttling as implemented [by tokio](https://tokio.rs/blog/2019-10-scheduler#reducing-cross-thread-synchronization) and other userspace schedulers, which avoids notifications if there is already an active thread, the "waker thread". That thread will then notify another waiter if it finds a message, creating a new waker thread. This PR changes the bounded and unbounded channels to fallback to blocking after observing a linearizability condition, using wakeup throttling to makeup for missed notifications without creating too many unnecessary wakeups.

Thanks to @kprotty for pointing me in the right direction for the wakeup throttling algorithm. Hopefully paired with https://github.com/crossbeam-rs/crossbeam/pull/1038 this should resolve all of the spinning issues in crossbeam/std.

Should resolve https://github.com/crossbeam-rs/crossbeam/issues/366 and https://github.com/crossbeam-rs/crossbeam/issues/997 (as well as https://github.com/rust-lang/rust/issues/114851 and https://github.com/rust-lang/rust/issues/112723 when upstreamed to std).

### DIFF
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -14,6 +14,7 @@ use crate::err::{
 };
 use crate::flavors;
 use crate::select::{Operation, SelectHandle, Token};
+use crate::waker::BlockingState;
 
 /// Creates a multi-producer multi-consumer channel of unbounded capacity.
 ///
@@ -1358,6 +1359,14 @@ impl<T> fmt::Debug for IntoIter<T> {
 }
 
 impl<T> SelectHandle for Sender<T> {
+    fn start(&self) -> Option<BlockingState<'_>> {
+        match &self.flavor {
+            SenderFlavor::Array(chan) => chan.sender().start_ref(),
+            SenderFlavor::List(chan) => chan.sender().start_ref(),
+            SenderFlavor::Zero(chan) => chan.start(),
+        }
+    }
+
     fn try_select(&self, token: &mut Token) -> bool {
         match &self.flavor {
             SenderFlavor::Array(chan) => chan.sender().try_select(token),
@@ -1370,11 +1379,11 @@ impl<T> SelectHandle for Sender<T> {
         None
     }
 
-    fn register(&self, oper: Operation, cx: &Context) -> bool {
+    fn register(&self, oper: Operation, cx: &Context, state: Option<&BlockingState<'_>>) -> bool {
         match &self.flavor {
-            SenderFlavor::Array(chan) => chan.sender().register(oper, cx),
-            SenderFlavor::List(chan) => chan.sender().register(oper, cx),
-            SenderFlavor::Zero(chan) => chan.sender().register(oper, cx),
+            SenderFlavor::Array(chan) => chan.sender().register(oper, cx, state),
+            SenderFlavor::List(chan) => chan.sender().register(oper, cx, state),
+            SenderFlavor::Zero(chan) => chan.sender().register(oper, cx, state),
         }
     }
 
@@ -1402,11 +1411,11 @@ impl<T> SelectHandle for Sender<T> {
         }
     }
 
-    fn watch(&self, oper: Operation, cx: &Context) -> bool {
+    fn watch(&self, oper: Operation, cx: &Context, state: Option<&BlockingState<'_>>) -> bool {
         match &self.flavor {
-            SenderFlavor::Array(chan) => chan.sender().watch(oper, cx),
-            SenderFlavor::List(chan) => chan.sender().watch(oper, cx),
-            SenderFlavor::Zero(chan) => chan.sender().watch(oper, cx),
+            SenderFlavor::Array(chan) => chan.sender().watch(oper, cx, state),
+            SenderFlavor::List(chan) => chan.sender().watch(oper, cx, state),
+            SenderFlavor::Zero(chan) => chan.sender().watch(oper, cx, state),
         }
     }
 
@@ -1420,6 +1429,17 @@ impl<T> SelectHandle for Sender<T> {
 }
 
 impl<T> SelectHandle for Receiver<T> {
+    fn start(&self) -> Option<BlockingState<'_>> {
+        match &self.flavor {
+            ReceiverFlavor::Array(chan) => chan.receiver().start_ref(),
+            ReceiverFlavor::List(chan) => chan.receiver().start_ref(),
+            ReceiverFlavor::Zero(chan) => chan.start(),
+            ReceiverFlavor::At(chan) => chan.start(),
+            ReceiverFlavor::Tick(chan) => chan.start(),
+            ReceiverFlavor::Never(chan) => chan.start(),
+        }
+    }
+
     fn try_select(&self, token: &mut Token) -> bool {
         match &self.flavor {
             ReceiverFlavor::Array(chan) => chan.receiver().try_select(token),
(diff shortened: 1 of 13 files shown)


ibraheemdev:
Falling back to parking for `wait_write` and` wait_next` in the unbounded channel still needs to be implemented.

ibraheemdev:
We probably also want to introduce a non-linearizable version of `try_recv`, because this doesn't actually fix the issue of `try_recv` taking longer than expected.

ibraheemdev:
Here are the benchmark results before and after. Note that all of the fallback cases are very unlikely to be hit on most setups, so this is just measuring the performance difference of new wakeup algorithm, which seems to be pretty mixed.

```diff
- test bounded_1::create    ... bench:          91 ns/iter (+/- 0)
+ test bounded_1::create    ... bench:          82 ns/iter (+/- 3)
- test bounded_1::mpmc      ... bench:  10,576,639 ns/iter (+/- 1,012,158)
+ test bounded_1::mpmc      ... bench:  11,206,415 ns/iter (+/- 2,787,440)
- test bounded_1::mpsc      ... bench:  21,551,625 ns/iter (+/- 975,344)
+ test bounded_1::mpsc      ... bench:  21,975,220 ns/iter (+/- 390,863)
- test bounded_1::oneshot   ... bench:         116 ns/iter (+/- 1)
+ test bounded_1::oneshot   ... bench:         104 ns/iter (+/- 3)
- test bounded_1::spmc      ... bench:  22,192,680 ns/iter (+/- 760,426)
+ test bounded_1::spmc      ... bench:  22,268,657 ns/iter (+/- 559,440)
- test bounded_1::spsc      ... bench:  23,097,047 ns/iter (+/- 76,889)
+ test bounded_1::spsc      ... bench:  22,753,449 ns/iter (+/- 143,048)
- test bounded_n::mpmc      ... bench:   2,387,457 ns/iter (+/- 313,821)
+ test bounded_n::mpmc      ... bench:   2,734,135 ns/iter (+/- 315,206)
- test bounded_n::mpsc      ... bench:   5,909,059 ns/iter (+/- 81,411)
+ test bounded_n::mpsc      ... bench:   6,635,209 ns/iter (+/- 134,351)
- test bounded_n::par_inout ... bench:   6,835,077 ns/iter (+/- 234,552)
+ test bounded_n::par_inout ... bench:   6,583,975 ns/iter (+/- 210,953)
- test bounded_n::spmc      ... bench:   5,649,130 ns/iter (+/- 127,809)
+ test bounded_n::spmc      ... bench:   5,748,161 ns/iter (+/- 130,954)
- test bounded_n::spsc      ... bench:   1,126,902 ns/iter (+/- 32,449)
+ test bounded_n::spsc      ... bench:   1,194,757 ns/iter (+/- 15,458)
- test unbounded::create    ... bench:          64 ns/iter (+/- 1)
+ test unbounded::create    ... bench:          57 ns/iter (+/- 1)
- test unbounded::inout     ... bench:          36 ns/iter (+/- 0)
+ test unbounded::inout     ... bench:          44 ns/iter (+/- 0)
- test unbounded::mpmc      ... bench:   1,181,564 ns/iter (+/- 204,692)
+ test unbounded::mpmc      ... bench:   1,200,760 ns/iter (+/- 174,606)
- test unbounded::mpsc      ... bench:   2,900,684 ns/iter (+/- 196,933)
+ test unbounded::mpsc      ... bench:   2,144,163 ns/iter (+/- 199,247)
- test unbounded::oneshot   ... bench:         103 ns/iter (+/- 0)
+ test unbounded::oneshot   ... bench:         113 ns/iter (+/- 0)
- test unbounded::par_inout ... bench:   4,799,151 ns/iter (+/- 487,975)
+ test unbounded::par_inout ... bench:   6,485,203 ns/iter (+/- 407,363)
- test unbounded::spmc      ... bench:   2,366,594 ns/iter (+/- 46,477)
+ test unbounded::spmc      ... bench:   2,014,760 ns/iter (+/- 28,476)
- test unbounded::spsc      ... bench:   1,061,444 ns/iter (+/- 14,589)
+ test unbounded::spsc      ... bench:   1,392,352 ns/iter (+/- 11,792)
```

ibraheemdev:
There are some changes I made on the rust-lang PR that have to be applied to this one as well.

I believe there's a large optimization here than can be made which is only notifying a waiter if the current thread sees remaining data on the channel.

the8472:
Since some of the issues indicate priority inversions I wonder if parking is the right choice or using inheritance-aware synchronization would be better on platforms that support it.

ibraheemdev:
> Since some of the issues indicate priority inversions I wonder if parking is the right choice or using inheritance-aware synchronization would be better on platforms that support it.

I think the root cause here are the spin locks, and those should definitely be removed. Do you mean like using a PI futex? that would still be parking, no?

the8472:
>  Do you mean using something like a PI-futex? 

yes. Though using them probably should be opt-in since at least on linux [they have lower throughput](https://github.com/rust-lang/rust/pull/131584#issuecomment-2492438131)

> I'm not sure how that would apply here, we have threads that are directly dependent on an operation being performed by another thread

I assume the cases where users are seeing deadlocks are a high priority thread waiting on a lower-priority thread. So even if spinning were replaced with parking that wouldn't tell the OS scheduler that for the high-priority thread to make progress it must wake up the lower-priority one but instead wakes up a medium-priority one, that would then be a priority inversion.

But yeah, the primary issue is spinning. I guess avoiding PI situations is a secondary issue that users might end up seeing once it's fixed.

PR #1106, opened 2024-05-02
Speed Up `concurrent_insert_get_same_key` Under Miri

ibraheemdev:
CI is now taking 3 hours under Miri after https://github.com/crossbeam-rs/crossbeam/pull/1101.


taiki-e (approved):
Thanks!

PR #1115, opened 2024-05-27
Initial version of `ArrayStack`

al8n:
Hi, this PR implements an `ArrayStack` inspired by the `ArrayQueue` implementation.


tomtomjhj:
This implementation is wrong. When running pop, after decrementing top from `t` to `t-1`, another thread can concurrently push and write at `t-1`. This races with the popping thread's read of `t-1`. Even if the accesses are made atomic, it won't be linearizable.

PR #1118, opened 2024-06-10
Add `CachePadded` variants with capped max alignment

Jules-Bertholet:
These are useful for ensuring that values *cross* as few cache lines as possible, without restricting whether they can *share* one.

### DIFF
--- a/crossbeam-utils/src/cache_padded.rs
+++ b/crossbeam-utils/src/cache_padded.rs
@@ -1,215 +1,250 @@
 use core::fmt;
 use core::ops::{Deref, DerefMut};
 
-/// Pads and aligns a value to the length of a cache line.
-///
-/// In concurrent programming, sometimes it is desirable to make sure commonly accessed pieces of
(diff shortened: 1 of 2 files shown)


Jules-Bertholet:
My phrasing of "don't cross a cache line" was suboptimal; I've changed it to "cross as few cache lines as possible", which is what I really want. My use case is that I have a large value (a lookup table), and I want it to take up as few cache lines as possible.

Jules-Bertholet:
> Given how we set our aarch64 and x86_64's alignments, I don't think this implementation is sufficient.

Could you elaborate?

taiki-e:
> > Given how we set our aarch64 and x86_64's alignments, I don't think this implementation is sufficient.
> 
> Could you elaborate?

On some CPUs, cache line is 64-byte, and on some, it is 128-byte. We choice 128-byte to avoid false sharing in either case, but if you wish to use it for a purpose other than avoiding false sharing, the opposite choice may be necessary. (By the way, on AArch64, there is CPU with 256-byte cache line (A64FX), but we ignore it at this time since it is very rare.)

PR #1120, opened 2024-06-20
chore: fix some comments

rustco:


taiki-e:
"deque" is "double-ended queue", not a typo of "dequeue (de-queue)".

taiki-e:
Closing as the current ones are correct, but thanks anyway for the PR.

PR #1114, opened 2024-05-22
Introduce disconnection state for channels

ryoqun:
(hopefully) much improved pr than https://github.com/crossbeam-rs/crossbeam/pull/959 ....

Currently, `crossbeam-channel` completely overlaps the channel connected status with rust sender/receiver instance aliveness. While this is uncontroversial and covers most use cases, it prevents advanced channel management.

Namely:

- graceful receiver-initiated shutdown (notify disconnection from receiver to sender and drain all sent-not-received messages): see description of #959
- temporal non-blocking faults (make {receiver,sender} disconnected until new {sender,receiver} is needed): see #750 
- off-loaded channel destruction (drop retired receivers outside performance sensitive code-path, which possibly could contain large number of not-received buffered messages while notifying disconnection to senders): see this comment: https://github.com/crossbeam-rs/crossbeam/pull/959#issuecomment-2099543477

While these seem to be unrelated to each other, the underlying problem is the lack of explicit channel disconnection state as described above.

Thereby, this adds the following new methods:

```rust
impl Sender {
    fn disconnect(&self) -> bool;
    fn connect(&self) -> bool;
}
impl Receiver {
    fn disconnect(&self) -> bool;
    fn connect(&self) -> bool;
}
```

Fortunately, the implementation is quite simple by decoupling the already-existing disconnected bit (`MARK_BIT` in `tail`) into the two: one for `::drop()` as was before, one for new `::disconnect()`, although this pr touches the heart of the impl unlike #959 (i.e. more intrusive....). Also, the almost all of existing code is already robust enough to introduce the 2 kinds of disconnected status. Come to think of it, {receivers, senders} should be able to cope with disconnected {senders, receivers} at any given moment of time due to `drop`s.

As for performance hit, it should be negligible because there's no additional operation at all. just `<< 2` instead of `<< 1` in the performance sensitive code-path. Also, no space increse. There were design considerations.

PR #1121, opened 2024-06-26
fix: port fix discard all message on receiver droped

### LINKED ISSUE #1102: Potential deadlock and resource leak when `Receiver` is dropped in `crossbeam_channel`'s bounded channel
I have encountered a situation where the code gets stuck indefinitely and the content in the channel's buffer is not dropped when the receiver is dropped. 

I'm not sure if this behavior is intended or if it represents a bug in the `crossbeam_channel` library. The situation is causing a potential deadlock and resource leaks in my code. I would appreciate further investigation by the upstream maintainers to clarify this behavior.

Here is the minimal reproducing code:

```rust
use crossbeam_channel::Sender;
use std::io::Error;

pub struct Request<A, R> {
    pub responder: Sender<R>,
    pub arguments: A,
}

impl<A, R> Request<A, R> {
    /// Call the service with the arguments and wait for the response.
    pub fn call(sender: &Sender<Request<A, R>>, arguments: A) -> Option<R> {
        let (responder, response) = crossbeam_channel::bounded(0);
        sender
            .send(Request {
                responder,
                arguments,
            })
            .unwrap();
        response.recv().ok()
    }
}

#[derive(Debug)]
struct Block {
    b: usize,
}

impl Drop for Block {
    fn drop(&mut self) {
        println!("dropping block {}", self.b);
    }
}

type ProcessBlockRequest = Request<Block, Result<(), Error>>;

struct ChainController {
    process_block_sender: Sender<ProcessBlockRequest>,
}

fn chain_service() -> ChainController {
    let (ct, cx) = crossbeam_channel::bounded::<ProcessBlockRequest>(10);

    std::thread::spawn(move || {
        let _cx_clone = cx.clone();
        return;
    });

    ChainController {
        process_block_sender: ct,
    }
}

impl ChainController {
    fn process_block(&self, b: Block) {
        Request::call(&self.process_block_sender, b);
    }
}

fn main() {
    let chain_controller = chain_service();
    chain_controller.process_block(Block { b: 1 });
}
```

I expect that `fn main` will not hang, but in reality it does hang.
If I comment out `let _cx_clone = cx.clone();`, then it does not hang.

driftluo:
fix #1102 

I think [this fix](https://github.com/rust-lang/rust/pull/108164) should be done on the crossbeam channel

### DIFF
--- a/ci/miri.sh
+++ b/ci/miri.sh
@@ -21,7 +21,7 @@ case "${group}" in
     # -Zmiri-ignore-leaks is needed because we use detached threads in tests in tests/golang.rs: https://github.com/rust-lang/miri/issues/1371
     MIRIFLAGS="${MIRIFLAGS} -Zmiri-ignore-leaks" \
       cargo miri test --all-features \
-      -p crossbeam-channel --test golang 2>&1 | ts -i '%.s  '
+      -p crossbeam-channel --test golang --test array 2>&1 | ts -i '%.s  '
     ;;
   others)
     cargo miri test --all-features \
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -675,7 +675,7 @@ impl<T> Drop for Sender<T> {
     fn drop(&mut self) {
         unsafe {
             match &self.flavor {
-                SenderFlavor::Array(chan) => chan.release(|c| c.disconnect()),
+                SenderFlavor::Array(chan) => chan.release(|c| c.disconnect_senders()),
                 SenderFlavor::List(chan) => chan.release(|c| c.disconnect_senders()),
                 SenderFlavor::Zero(chan) => chan.release(|c| c.disconnect()),
             }
@@ -1185,7 +1185,7 @@ impl<T> Drop for Receiver<T> {
     fn drop(&mut self) {
         unsafe {
             match &self.flavor {
-                ReceiverFlavor::Array(chan) => chan.release(|c| c.disconnect()),
+                ReceiverFlavor::Array(chan) => chan.release(|c| c.disconnect_receivers()),
                 ReceiverFlavor::List(chan) => chan.release(|c| c.disconnect_receivers()),
                 ReceiverFlavor::Zero(chan) => chan.release(|c| c.disconnect()),
                 ReceiverFlavor::At(_) => {}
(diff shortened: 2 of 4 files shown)


taiki-e:
Thanks for the PR. It seems that Miri reports memory leak: https://github.com/crossbeam-rs/crossbeam/actions/runs/12221429705/job/34090460906?pr=1121#step:4:168

driftluo:
> Thanks for the PR. It seems that Miri and sanitizer reports memory leak: https://github.com/crossbeam-rs/crossbeam/actions/runs/12221429705/job/34090460906?pr=1121#step:4:168 https://github.com/crossbeam-rs/crossbeam/actions/runs/12221429705/job/34090462020?pr=1121#step:5:186


The memory leak reported by miri is the panic on drop test case, which seems to be the expected behavior, because https://github.com/rust-lang/rust/pull/108164/files#diff-ed6b99777af1301c859ee5a593fca195c952216d861d87f6ac7cb44b82fdc280R482 The comment here indicates that if there is a panic, a memory leak will occur

I think std seems to assume that this behavior is expected. Do we need to make some extra efforts for this?

ryoqun:
@taiki-e hi, is there some timeline for merging this bug-fix pr? is there some blocker? my pr #1114 is touching same area of code and i'd rather to rework my pr once after this pr landed...

taiki-e:
@driftluo 

> The memory leak reported by miri is the `panic_on_drop` test case, which seems to be the expected behavior, because https://github.com/rust-lang/rust/blob/master/library/std/src/sync/mpmc/array.rs#L487-L489 The comment here indicates that if there is a panic, a memory leak will occur
> 
> I think std seems to assume that this behavior is expected. Do we need to make some extra efforts for this?

Assuming this PR is based on the rust-lang/rust PR I reviewed, the implementation should be fine with this. (But if so, please make that clear when submitting a patch based on someone else's work.)

That said, we cannot merge PRs that have not passed CI, so you will need to adjust the failing test. (Perhaps a similar way to https://github.com/crossbeam-rs/crossbeam/commit/2d3430f74d91b2e339411f32b92b84408813ad90 's MIRI_LEAK_CHECK=1 approach is fine.)

driftluo:
@taiki-e I use `-Zmiri-ignore-leaks` to disable Miri test on the channel test, but on sanitizer test, I can't find any flag to ignore it

driftluo:
@taiki-e ok, I ignore `panic_on_drop` test on sanitizer, now ci is pass

taiki-e (approved):

PR #1123, opened 2024-07-04
Fix all warnings from the latest Clippy

powergee:
The Clippy from the stable channel now produces new warnings, which this PR addresses. Specifically, those warnings are:

* [`declare_interior_mutable_const`](https://rust-lang.github.io/rust-clippy/master/index.html#/declare_interior_mutable_const): For efficient `Block` initialization, a constant `Slot` is used, but it has interior mutability, which Clippy warns about. I believe that using the inline-const expression in `Block::new` is ideal, but it is experimental in the current MSRV (1.61). Other solutions either have performance overheads (e.g., `array::map`, `array::from_fn`) or require nightly features (e.g., `MaybeUninit::uninit_array`).

* [`doc_lazy_continuation`](https://rust-lang.github.io/rust-clippy/master/index.html#/doc_lazy_continuation): Some indentations are missing in the documentation.

* [`zero_repeat_side_effects`](https://rust-lang.github.io/rust-clippy/master/index.html#/zero_repeat_side_effects): When there are no explicit cases for the `select!` macro, it creates a zero-sized array with initialization that calls a function. Clippy is concerned if the function has side effects. In our case, it has no side effects and can be trivially resolved.

### DIFF
--- a/crossbeam-channel/src/select_macro.rs
+++ b/crossbeam-channel/src/select_macro.rs
@@ -686,10 +686,9 @@ macro_rules! crossbeam_channel_internal {
     ) => {{
         const _LEN: usize = $crate::crossbeam_channel_internal!(@count ($($cases)*));
         let _handle: &dyn $crate::internal::SelectHandle = &$crate::never::<()>();
-        let _entry = (_handle, 0, ::std::ptr::null());
 
         #[allow(unused_mut)]
-        let mut _sel = [_entry; _LEN];
+        let mut _sel = [(_handle, 0, ::std::ptr::null()); _LEN];
 
         $crate::crossbeam_channel_internal!(
             @add
--- a/crossbeam-channel/src/select_macro.rs
+++ b/crossbeam-channel/src/select_macro.rs
@@ -686,9 +686,10 @@ macro_rules! crossbeam_channel_internal {
     ) => {{
         const _LEN: usize = $crate::crossbeam_channel_internal!(@count ($($cases)*));
         let _handle: &dyn $crate::internal::SelectHandle = &$crate::never::<()>();
+        let _entry = (_handle, 0, ::std::ptr::null());
 
         #[allow(unused_mut)]
-        let mut _sel = [(_handle, 0, ::std::ptr::null()); _LEN];
+        let mut _sel = [_entry; _LEN];
 
         $crate::crossbeam_channel_internal!(
             @add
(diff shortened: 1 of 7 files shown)


powergee:
> I would prefer ignoring this lint at workspace level [like my other projects](https://github.com/taiki-e/atomic-maybe-uninit/blob/3629d156f32cf7c1d736b670197ac22d2d91d856/Cargo.toml#L92).

I agree with your suggestion. I will add another commit to ignore the `declare_interior_mutable_const` warning at the module level. I found that there is [another line](https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-utils/src/atomic/atomic_cell.rs#L975) that ignores the same warning 🙃.

> I feel it is a clippy bug that this lint warns of a code that has no side effects...

It also seems strange to me because `crossbeam_channel::never` clearly does not have any side effects, and even changing that function to `const` does not resolve the warning. It might be worth looking into this issue more deeply if I have time.

powergee:
I pushed a commit and all CI checks are passed.

crossbeam-channel/src/select_macro.rs
```diff
@@ -686,9 +686,10 @@ macro_rules! crossbeam_channel_internal {
     ) => {{
         const _LEN: usize = $crate::crossbeam_channel_internal!(@count ($($cases)*));
         let _handle: &dyn $crate::internal::SelectHandle = &$crate::never::<()>();
+        let _entry = (_handle, 0, ::std::ptr::null());
 
         #[allow(unused_mut)]
-        let mut _sel = [(_handle, 0, ::std::ptr::null()); _LEN];
+        let mut _sel = [_entry; _LEN];
```
taiki-e, line 692: I still think this change is unneeded and clippy lint's false positive.

powergee: Thanks, I reverted this change and manually ignored `zero_repeat_side_effects` in that case.

crossbeam-channel/src/select_macro.rs
```diff
@@ -687,7 +687,7 @@ macro_rules! crossbeam_channel_internal {
         const _LEN: usize = $crate::crossbeam_channel_internal!(@count ($($cases)*));
         let _handle: &dyn $crate::internal::SelectHandle = &$crate::never::<()>();
 
-        #[allow(unused_mut)]
+        #[allow(unused_mut, zero_repeat_side_effects)]
```
taiki-e, line 690:
```suggestion
        #[allow(unused_mut, clippy::zero_repeat_side_effects)]
```

taiki-e (approved):

Issue #1122, opened 2024-07-01
Document the semantics of SkipMap::compare_insert


blinsay:
I was trying to figure out from the documentation if `SkipMap::compare_insert` is atomic and didn't see anything explicitly mentioned. Based on a toy program that tries to use a `SkipMap<&str, u64>` as a set of atomic counters, it seems like it probably has the same semantics to `get_or_insert_with` and allows two concurrent callers to race. It'd be nice to reproduce that note on `compare_insert` (and maybe also `get_or_insert`) as well.

Happy to attach my toy program if it's useful. Thank you!

Tianion:
SkipMap is the lock-free skip list, so it naturally allows two callers to race. It is atomic when actually replacing data(after I fixed it #1101), using compare_exchange. I will update the document. The implementation of compare_insert was done before I fixed the concurrency issues, which has led to some inaccuracies in the current documentation.

blinsay:
Thank you!

PR #1124, opened 2024-07-13
docs(skiplist): update insert function document

Tianion:
"it will be removed before inserting the new one", it looks like there are two steps involved in the process described.
#1122


taiki-e:
Thanks. It seems to me that “using CAS” is a description of implementation details and does not explain what specific behavior is provided. (I guess you want to document the semantics that no intermediate state is observed during the update.)

Issue #540, opened 2020-08-06
SkipList Memory Release Issue


Jarves42:
I am facing a strange memory release issue on the use of SkipList.

So I have used skiplist to the cache layer in my system. On the start of the service, I am creating rust std library thread for cache refresh and eviction. I am doing this both operation in single function call only. And there is a wait of 2 minutes between to refresh cache call. **I have created same cache structure for multiple partitions(~200 partitions).** And I am refreshing all partitions in single thread in single function call.

**Addition of new keys in cache would be in parallel of cache refresh cycle.**

**And we have about 1M keys on average in skiplist.**

Problem I am facing that, even after removing keys from cache it is not releasing memory. And memory continuouly increasing till OOM. And if I increase refresh interval to 10 or more minutes I am able to see stready memory in my system which I can't as it will directly affect user experience.

**I have even tried using threadpool to refresh partition instead of doing all in one thread. That also didn't work.**

I can't post exact function as it is part of very big code base. But trying to give structure overview.

```
struct Cache {
   inner: SkipList<Vec<u8>,CacheValue>
}
```

```
struct CacheValue {
    pub updated: i64,

    pub last_accessed: AtomicI64,

    pub access_count: AtomicUsize,

    pub refresh_run: usize,

    pub data: BitPageVec,
}
```
```
struct Partition {
   id: String,
   cache: Cache
}
```

So basically spawned thread on start have access of Vec<Partition>. It iterator over it at every 2 minutes and removed expired keys and refreshed remaining keys.

No explicit impl of Drop trait for any struct mentioned above. 

Any lead would be really helpful. Thanks.

stjepang:
So the issue with `SkipList` is that it deallocates memory lazily using the epoch-based GC. If you remove an entry from a skiplist, it will be stashed away on some kind of garbage list. Deallocation happens periodically after every 128 skiplist operations, assuming only one thread is accessing it. Deallocation could be delayed for longer if multiple threads are accessing it.

This mechanism is necessary for efficiency under concurrency. The core idea of epoch-based GC is that we pessimistically assume another thread is accessing the skip list while we're removing entries, so deallocation must be delayed a bit.

Jarves42:
@stjepang Thanks for your comment. Can you give more details about **Deallocation happens periodically after every 128 skiplist operations part** you mentioned. Like all op or any specific op.

[...]

PR #1126, opened 2024-08-01
Add failing memory leak tests for SkipMap

bojanserafimov:
Hi,

I noticed a memory leak in SkipMap and wrote two tests that reproduce it. I don't know where the bug is, or whether I'm misusing the library. I'd be happy to help fix the issue if any of the authors can give me some pointers.

To write these tests I copied the [`drops()` test from base.rs](https://github.com/bojanserafimov/crossbeam/blob/276c9a1aba753099d48e4a54942ad31a2f884e40/crossbeam-skiplist/tests/base.rs#L857) and changed `SkipList` to `SkipMap`.

I will open this draft for review if we end up fixing the issue and getting the tests to pass

### DIFF
--- a/crossbeam-skiplist/tests/map.rs
+++ b/crossbeam-skiplist/tests/map.rs
@@ -942,3 +942,92 @@ fn concurrent_insert_get_same_key() {
     }
     handle.join().unwrap()
 }
+
+#[test]
+fn drops_after_remove() {
+    use std::sync::atomic::{AtomicUsize, Ordering};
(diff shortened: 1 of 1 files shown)


bojanserafimov:
Adding the following at the end of the test fixes it:
```
    crossbeam_epoch::pin().flush();
    crossbeam_epoch::pin().flush();
```

Is that the best way to clean up after a `SkipMap`? Also why are two flushes necessary?

taiki-e:
I guess this is the same issue as https://github.com/crossbeam-rs/crossbeam/issues/540.

PR #1135, opened 2024-09-13
Add force_send method to channel Sender

### LINKED ISSUE #400: Lossy Channels
Is there any current work for adding lossy bounded channels? Or any strong reason to not add them? If not, I might open a PR in the next few weeks with an implementation.

bergkvist:
### Add a force_push method to the Sender part of a channel. 

In particular, this is useful for bounded channels of non-zero size.

Motivation:
- https://github.com/crossbeam-rs/crossbeam/issues/400

Design goals:
- It is fine to lose messages, but messages should never appear in a different order from which they were sent
- For bounded (non-zero-sized) channels, force_send should never fail (except when the receiver has been dropped)
- Should not be blocking

```rs
use crossbeam_channel::{bounded, ForceSendError};

let (s, r) = bounded(3);

assert_eq!(s.force_send(0), Ok(None));
assert_eq!(s.force_send(1), Ok(None));
assert_eq!(s.force_send(2), Ok(None));
assert_eq!(s.force_send(3), Ok(Some(0)));

assert_eq!(r.recv(), Ok(1));

assert_eq!(s.force_send(4), Ok(None));

assert_eq!(r.recv(), Ok(2));
assert_eq!(r.recv(), Ok(3));

assert_eq!(s.force_send(5), Ok(None));
assert_eq!(s.force_send(6), Ok(None));
assert_eq!(s.force_send(7), Ok(Some(4)));
assert_eq!(s.force_send(8), Ok(Some(5)));

assert_eq!(r.recv(), Ok(6));
assert_eq!(r.recv(), Ok(7));
assert_eq!(r.recv(), Ok(8));

drop(r);

assert_eq!(s.force_send(9), Err(ForceSendError(9)));
```

Closes https://github.com/crossbeam-rs/crossbeam/issues/400

### DIFF
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -1,6 +1,5 @@
 //! The channel interface.
 
-use core::sync::atomic::AtomicUsize;
 use std::fmt;
 use std::iter::FusedIterator;
 use std::mem;
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -1,5 +1,6 @@
 //! The channel interface.
 
+use core::sync::atomic::AtomicUsize;
 use std::fmt;
 use std::iter::FusedIterator;
 use std::mem;
@@ -10,7 +11,8 @@ use std::time::{Duration, Instant};
 use crate::context::Context;
 use crate::counter;
 use crate::err::{
-    RecvError, RecvTimeoutError, SendError, SendTimeoutError, TryRecvError, TrySendError,
+    ForceSendError, RecvError, RecvTimeoutError, SendError, SendTimeoutError, TryRecvError,
+    TrySendError,
 };
 use crate::flavors;
 use crate::select::{Operation, SelectHandle, Token};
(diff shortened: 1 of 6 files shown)


crossbeam-channel/src/channel.rs
```diff
@@ -410,6 +411,45 @@ impl<T> Sender<T> {
...
     }
 
+    /// Sends a message to a channel without blocking. It will only fail if the channel is disconnected.
+    ///
+    /// This method will either send a message into the channel immediately,
+    /// overwrite and return the last value in the channel or return an error if
+    /// the channel is full. The returned error contains the original message.
+    ///
+    /// If called on a zero-capacity channel, this method will send the message only if there
+    /// happens to be a receive operation on the other side of the channel at the same time.
```
proski, line 421: It's not clear from the description what would happen if there is no receive operation. Would the call fail (despite what the start of the comment says)? Would the data be returned to the caller?

proski:
The "lossy channel" proposal is about removal the oldest element rather than the latest:

> By "lossy channel", I mean one that discards old data when the buffer is full

Removing the latest element makes it impossible to force send multiple messages unless the reader reads a message at the right time. That's very limiting and unreliable. The "force" is one element deep.

Figuratively speaking, users expect a bulldozer, and you give them a chisel.

taiki-e (changes requested):
Thanks for the PR. I would prefer that `force_push` in `crossbeam_channel` have the same semantics as `force_push` in `crossbeam_queue::ArrayQueue` (https://github.com/crossbeam-rs/crossbeam/pull/789):

```rs
let q = crossbeam_queue::ArrayQueue::new(2);

assert_eq!(q.force_push(10), None);
assert_eq!(q.force_push(20), None);
assert_eq!(q.force_push(30), Some(10));
```

As said in https://github.com/crossbeam-rs/crossbeam/issues/400#issuecomment-1837125914, it should be able to be implemented by porting https://github.com/crossbeam-rs/crossbeam/pull/789.

bergkvist:
I agree that removing the oldest entry is more desirable than removing the most recent one.

I've been putting some more thought into this, and force_send seems to be tricky compared to try_send.

Consider the following, a `bounded(6)` channel (H=head (read index), T=tail (write index)):
```
step  next state        op
       
        H
                    T
1     [ 0, 1, 2, 3, _, _ ]   force_send <- 3
      
        H
                       T
2     [ 0, 1, 2, 3, 4, _ ]   force_send <- 4
      
           H
                       T
3     [ _, 1, 2, 3, 4, _ ]   recv -> 0
      
           H
        T              
4     [ _, 1, 2, 3, 4, 5 ]   force_send <- 1
              
           H
           T            
5     [ 6, 1, 2, 3, 4, 5 ]   force_send <- 6
              
              H
              T            
6     [ 6, 7, 2, 3, 4, 5 ]   force_send <- 7 (removes/returns 1)
      
                 H
              T            
7     [ 6, 7, _, 3, 4, 5 ]   recv -> 2
```

Notice that `force_send` only has different behavior from `try_send` in step 6, where `try_send` will fail, but `force_send` should overwrite the oldest value, updating both H and T indices at the same time.

And here lies the problem, H and T both are atomic usizes, meaning we can ensure atomicity when updating one of them using `compare_exchange_weak(..)`, but if we need to update both of them at the same time, we have a problem.

sbarral:
> [...] but since we need to update both of them at the same time, we have a problem (on x86 we have https://docs.rs/mwcas/latest/mwcas/).

I had a cursory look at the implementation in #789 and I couldn't really see what makes this issue different. What mandates an atomic update of both head and tail for a channel when a non-atomic update is acceptable for `ArrayQueue`?

Issue #1138, opened 2024-09-29
CachePadded doesn't work correctly (Mac M1)


xpepermint:
I noticed that CachePadded doesn't work correctly on my MacBook M1, and this applies not just to AtomicBool but also to any atomic type (e.g., AtomicUsize, AtomicBool, AtomicU8, etc.). The issue arises in some strange multi-threaded cases, leading to unexpected behavior.

Initially, I thought the problem was caused by something in my code - possibly due to incorrect atomic ordering or a logic bug. After investigating further, I replaced CachePadded with the atomic type directly (e.g., AtomicBool or AtomicUsize), and false reading (load returning some strange number) effects disappeared. This suggests that the issue might not be with atomic operations themselves but with how CachePadded interacts with them on certain architectures.

I also experimented with the alignment of CachePadded. When I used #[repr(align(64))] to explicitly align CachePadded to a 64-byte boundary, the issue disappeared. This leads me to believe that the problem might be related to overalignment (e.g., aligning to 128 bytes instead of 64 bytes on certain architectures like ARM or Apple Silicon). 

https://github.com/crossbeam-rs/crossbeam/blob/abf24e1f31a76ef2590688d0c2bb55f82c7410a9/crossbeam-utils/src/cache_padded.rs#L83

I'd appreciate any clarification on whether CachePadded should be aligned to 128 bytes for specific architectures like aarch64 (Apple M1), or if this alignment is unnecessary and potentially problematic. If padding atomic types like AtomicBool is required or recommended for preventing false sharing, I'd also be interested in understanding why aligning to 64 bytes seems to fix the issue.

Thanks for your help!

xpepermint:
Btw, for x86_64 and powerpc64, I think we should align to their actual cache line size as well, which is typically 64 bytes for both architectures. The use of 128 bytes for alignment in certain contexts (like prefetching or spatial prefetcher behavior on Intel's Sandy Bridge) is an optimization for specific CPU behaviors, but this does not mean the cache line size itself is 128 bytes. So ... Any thoughts?

taiki-e:
Thanks for the report! Could you tell us more specific information? (e.g., how to reproduce the problem or on what code exactly was tried and what happened.)

> I replaced CachePadded with the atomic type directly (e.g., AtomicBool or AtomicUsize), and false reading (load returning some strange number) effects disappeared.

[...]

PR #1139, opened 2024-09-29
CachePadded aligns aarch64 to 64 instead of 128

xpepermint:
Ref: https://github.com/crossbeam-rs/crossbeam/issues/1138

### DIFF
--- a/crossbeam-utils/src/cache_padded.rs
+++ b/crossbeam-utils/src/cache_padded.rs
@@ -13,7 +13,7 @@ use core::ops::{Deref, DerefMut};
 ///
 /// Cache lines are assumed to be N bytes long, depending on the architecture:
 ///
-/// * On x86-64, aarch64, and powerpc64, N = 128.
+/// * On x86-64, and powerpc64, N = 128.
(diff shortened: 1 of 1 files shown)


taiki-e:
If we revert align of AArch64, we need to address the following comment explaining why we are currently using 128 in AArch64.

https://github.com/crossbeam-rs/crossbeam/blob/abf24e1f31a76ef2590688d0c2bb55f82c7410a9/crossbeam-utils/src/cache_padded.rs#L70-L73

taiki-e:
Considering that the cache line size of the Apple M1 actually seems to be 128-bytes, your guess at https://github.com/crossbeam-rs/crossbeam/issues/1138 regarding the problem you encountered is probably wrong and doing this does not seem right.

```console
$ sysctl machdep.cpu.brand_string
machdep.cpu.brand_string: Apple M1 (Virtual)

$ sysctl hw.cachelinesize
hw.cachelinesize: 128
```

```console
$ sysctl machdep.cpu.brand_string
machdep.cpu.brand_string: Apple M1 Pro

$ sysctl hw.cachelinesize
hw.cachelinesize: 128
```

PR #1141, opened 2024-10-11
Explicitly annotate lifetime of entry methods

thynson:
The `.key()` and `.value()` should return a borrow whose lifetime aligns with the container, instead of the entry itself.


taiki-e (approved):
Thanks!

thynson:
Hello @taiki-e ,  could you ship a new release for this PR? Sorry for disturbing but sometimes it's inconvenient for us to use a custom fork.

taiki-e:
See #1162.

PR #1144, opened 2024-10-26
Clearly document SegQueue is FIFO

tisonkun:
Although a "queue" is typically FIFO, I found it unclear when I first read the doc and I have to read the code to verify it.


taiki-e (approved):
Thanks.

PR #1146, opened 2024-11-07
channel: Create `list::Block` directly on the heap

cuviper:
The list channel's `Block::new` was causing a stack overflow because it
held 32 item slots, instantiated on the stack before moving to
`Box::new`. The 32x multiplier made modestly-large item sizes untenable.

That block is now initialized directly on the heap.

References from the `std` channel implementation:
* https://github.com/rust-lang/rust/issues/102246
* https://github.com/rust-lang/rust/pull/132738


taiki-e (approved):
Thanks!

Issue #1145, opened 2024-11-07
Non-macro biased Select?


SimonSapin:
https://github.com/crossbeam-rs/crossbeam/pull/1040 added a `select_biased!` macro. Some internal functions gained an `is_biased: bool` parameter to support it, but the non-macro public API around the `Select` type does not expose this functionality at all. Some of those internal functions are only ever called with `false`. Biased functionality could be useful to add to the `Select` type but I’m not sure what the API should look like exactly. ~Doubling the number of methods does not feel great.

For context: I’m building a priority queue on top of an array of crossbeam channels. My receiver should block until any of the inner receivers are ready, but then consume one item from the highest-priority channel rather than arbitrary one. (Starvation is mitigated by "ageing" some low-priority items into a one-level-higher priority channel.) As a work-around I call `Select::ready()` then an explicit loop with `Receiver::try_recv()`, but this looks like it does redundant work compared to an hypothetical `Select::biased_select()`

taiki-e:
Added in crossbeam-channel 0.5.14 (#1150).

PR #1150, opened 2024-11-26
feat: add new_biased constructor for biased channel selection

JabobKrauskopf:
I added a new `with_bias` constructor to the `Select` struct to allow for selecting channels with a bias.

This should make it possible to use a biased select without the `select_biased!` macro.

Looking for feedback on naming or if I missed "implementing" something

CC https://github.com/crossbeam-rs/crossbeam/issues/1145

Edit:

The `with_bias` function was renamed to `new_biased`

Closes https://github.com/crossbeam-rs/crossbeam/issues/1145


crossbeam-channel/src/select.rs
```diff
@@ -633,6 +636,28 @@ impl<'a> Select<'a> {
...
+    ///
+    /// let mut sel = Select::with_bias();
+    ///
+    /// // The list of operations is empty, which means no operation can be selected.
+    /// assert!(sel.try_select().is_err());
+    /// ```
+    pub fn with_bias() -> Self {
+        Self {
+            biased: true,
+            ..Default::default()
```
taiki-e, line 660: Why are you doing things differently here than `new`? (about `handles` field)

taiki-e (changes requested):
Thanks. As for naming, I think `new_biased` is a better name here.

crossbeam-channel/src/select.rs
JabobKrauskopf: If you e.g. decide that the handles field should be initialized differently when a new `Select` is created, it would only need to be edited in the `new` function and not in both. `new_biased` would use the same values for all fields except for the `biased` field.

I can change it if you want :)

taiki-e: Ah, sorry, I missed Default::default just calls Select::new.

taiki-e (approved):
LGTM

taiki-e:
Published in crossbeam-channel 0.5.14.

PR #1156, opened 2024-12-08
Relax MSRV of utils to 1.38

taiki-e:
Implements https://github.com/crossbeam-rs/crossbeam/pull/1075#issuecomment-2039050221.

This is not a PR intended to be merged as is, and I think it is fine to relax MSRV to 1.56, 1.52, or worse, 1.43, but I'm quite negative with relaxing to 1.42 or lower because it decreases efficiency (see crossbeam-utils/src/sync/once_lock.rs).

cc @mgeier

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,7 @@ name = "crossbeam"
 # - Update README.md (when increasing major or minor version)
 # - Run './tools/publish.sh crossbeam <version>'
 version = "0.8.4"
-edition = "2021"
+edition = "2018"
 rust-version = "1.61"
 license = "MIT OR Apache-2.0"
 repository = "https://github.com/crossbeam-rs/crossbeam"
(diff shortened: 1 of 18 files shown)


taiki-e:
Relax to 1.56, which considered non-controversial, has been done in https://github.com/crossbeam-rs/crossbeam/pull/1157.

mgeier:
Thanks for the heads-up!

I appreciate any lowering of the MSRV, but I also fully agree that it shouldn't be done if there are real disadvantages (like a performance hit).

I still think that it is a problem to combine different independent tools with different MSRVs into a single crate.

Even if this PR would be merged as is, it would still not avoid any future MSRV bumps affecting all files of the crate even if it wouldn't be needed for a specific file.

For the time being, I will continue using my hack to grab `cache_padded.rs` without being restricted by the MSRV (https://github.com/mgeier/rtrb/pull/115). I would prefer to have it as a proper `Cargo.toml` dependency, but IMHO it's not worth accepting the unnecessary MSRV increase.

PR #1158, opened 2024-12-11
skiplist: extract `equivalent` mod to a crate

al8n:
Hi, this PR extracts `equivalent` mod in `crossbeam-skiplist` to [`equivalent-flipped`](https://crates.io/crates/equivalent-flipped) crate so that users can implement `Comparable` and `Equivalent` without depending on `crossbeam-skiplist`.


al8n:
`indexmap` directly uses `equivalent: v1` to avoid break-changing or compatible issues.
I published `equivalent-flipped` version `1.0.0`, and I think there should not be more changes in `equivalent-flipped`, so it should be ok for `crossbeam-skiplist` to use `equivalent-flipped: v1`.

al8n:
Hi @taiki-e, any suggestions on this PR?

thynson:
This looks reasonable from my eye. 

However, since the desired next release version is 0.1.4, breaking change is should be avoided, so I would like to suggest to  implemented it in this way:

```rust
#[cfg(feature="equivalent_flipped")]
impl <T> Equivalent for T where T: equivalent_flipped::Equivalent { /* ** */ }

#[cfg(feature="equivalent_flipped")]
impl <T> Comparable for T where T: equivalent_flipped::Comparable { /* ** */ }
```

thynson:
After carefully considering, I believe it will be a breaking change anyway.

Either re-exports or blanket impls break the following code, rustc consider the following trait impls conflicted.

```rust

struct SomeType;

impl equivalent_flipped::Equivalent for SomeType { /* ** */ }

impl crossbeam_skiplist::Equivalent for SomeType { /* ** */ }
```

al8n:
> After carefully considering, I believe it will be a breaking change anyway.
> 
> 
> 
> Either re-exports or blanket impls break the following code, rustc consider the following trait impls conflicted.
> 
> 
> 
> ```rust
> 
> 
> 
> struct SomeType;
> 
> 
> 
> impl equivalent_flipped::Equivalent for SomeType { /* ** */ }
> 
> 
> 
> impl crossbeam_skiplist::Equivalent for SomeType { /* ** */ }
> 
> ```
> 
> 
> 
> 

I do not think this is a breaking change, because the current 0.1.3 still using Borrow, the Equivalent and Comparable version has not been released.

thynson:
> Equivalent and Comparable version has not been released, so this will not influence the crates which are using 0.1.3...

Okay, so is there any concern blocking the release of v0.1.4? @taiki-e

taiki-e:
The re-export compatibility issue mentioned in https://github.com/crossbeam-rs/crossbeam/pull/1132 is why this is a blocker that should be resolved before release.
And the blocker against merging this is the supply chain issue.

al8n:
Hi @taiki-e, in the latest commit, I use future proof style dependency to avoid re-export compatibility issue, any suggestion or direciton for this?

thynson:
I don't think it's a good practice. For given crate, cargo merge all the features required by different dependents.
So, it's likely to break crates that use `Equivalent` but does not enable feature `equivalent-flipped-1` when another crate enabled `equivalent-flipped-1`.

taiki-e:
As for comparison trait, there is also another alternative idea (https://github.com/crossbeam-rs/crossbeam/issues/1180), so it would be better not to depend on a specific crate for now. And in any case I'm not so positive about adding dependencies that are not under the control of crossbeam or its primary users' maintainers.

So, I'm going to close this, but thanks anyway for the PR!

Issue #1180, opened 2025-02-14
Custom comparators for SkipList


matthew-mcallister:
It would be useful to be able to specify a custom C++-style comparator function to override `SkipList` key order at runtime, specifically for an ordered K/V store like [fjall](https://github.com/fjall-rs/fjall).

The implementation would be easy. Unfortunately, Rust has no standard API for custom comparators, so there's no obvious, future-proof design to go with. The closest reference points I'm aware of are the [`compare` crate](https://docs.rs/compare/latest/compare/) and C++ [`std::map`](https://en.cppreference.com/w/cpp/container/map).

taiki-e:
Is this the same as the already merged https://github.com/crossbeam-rs/crossbeam/pull/1132? Or are you requesting something different?

matthew-mcallister:
Different. The idea is essentially to add a type parameter `C` and a field `comparator: C` to `SkipList` that implements a trait similar to (pseudocode)
```rust
trait Comparator<K> {
   fn compare(&self, left: &K, right: &K) -> std::cmp::Ordering;
}
```
When comparing keys, instead of `key1.compare(key2)`, you write `self.comparator.compare(key1, key2)`. The default comparator would be to do the same thing that `SkipList` already does:
```rust
impl<K: Ord> Comparator<T> for DefaultComparator {
    fn compare(&self, left: &K, right: &K) -> Ordering {
        Ord::cmp(left, right)
    }
}
```
But the use case I had in mind was to use `Box<dyn Comparator>` as the comparator to support an arbitrary or dynamic comparison function on `Vec<u8>` keys.

That said, fjall seems to be moving towards a custom skiplist implementation, at which point the feature could be implemented without support from crossbeam.

al8n:
> Different. The idea is essentially to add a type parameter `C` and a field `comparator: C` to `SkipList` that implements a trait similar to (pseudocode)
> 
> trait Comparator<K> {
>    fn compare(&self, left: &K, right: &K) -> std::cmp::Ordering;
> }
> When comparing keys, instead of `key1.compare(key2)`, you write `self.comparator.compare(key1, key2)`. The default comparator would be to do the same thing that `SkipList` already does:
> 
> impl<K: Ord> Comparator<T> for DefaultComparator {
>     fn compare(&self, left: &K, right: &K) -> Ordering {
>         Ord::cmp(left, right)
>     }
> }
[...]

PR #1182, opened 2025-02-24
Add custom comparators

matthew-mcallister:
Closes https://github.com/crossbeam-rs/crossbeam/issues/1180.

The diff is large due to having to add a type parameter `C` and update constraints in numerous places, but the logic changes are not large.

This trait adds the `Comparator` and `Equivalator` traits, which generalize the `Comparable` and `Equivalent` traits by taking a `self` parameter in addition to the two keys to compare, making it possible to change the comparison function used without changing the key type.

## Summary

- Adds `trait Comparator<L, R>` and `trait Equivalator<L, R>`.
- Adds `struct OrdComparator`, which is a zero-sized type that implements `Comparator` using `Borrow` and `Ord`.
- Adds a type parameter `C` to `SkipList`, `SkipMap`, and `SkipSet`.
  - `C` is defaulted to `OrdComparator`. This preserves the existing behavior of functions like `get()`, `insert()`, etc.
- Adds `SkipList::with_comparator()`, `SkipSet::with_comparator()`, and `SkipMap::with_comparator()` constructors.
  - `SkipList::new()`, `SkipMap::new()`, and `SkipSet::new()` always use `C = OrdComparator`. It would be nice if it worked for any `C` that implements `Default`, but that would break type inference. This is exactly the reason [`std::collections::HashMap::new()`](https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.new) always uses `S = RandomState` for its defaulted parameter even if `S` implements `Default`.
- Adds a test using a wrapper around `Box<dyn Fn(&[u8], &[u8]) -> Ordering>` as the comparator.

## Hashing

Because the goal of these traits is to support totally different comparison operations from the default for the underlying key type, `Equivalator` explicitly does not require two equivalent keys to have the same hash. If someone, someday wants to support `Equivalator` on their hash map, they will have to generalize the `Hash` trait as well.

## Safety

My one minor question is whether `unsafe impl<Q, R, K, V, C> Send for RefRange<'_, Q, R, K, V, C>` should require `C: Sync`. I don't think so because it doesn't require `K` or `V` to be sync, but it was enough to make me second-guess myself.


crossbeam-skiplist/tests/base.rs
```diff
@@ -935,23 +938,35 @@ fn comparable_get() {
...
+        fn compare(&self, foo: &Foo, key: &FooRef<'a>) -> std::cmp::Ordering {
             let a = u64::from_be_bytes(key.data[..8].try_into().unwrap());
             let b = u32::from_be_bytes(key.data[8..].try_into().unwrap());
-            Foo { a, b }.cmp(self)
+            Foo { a, b }.cmp(foo)
         }
     }
 
-    let s = SkipList::new(epoch::default_collector().clone());
+    let s = SkipList::with_comparator(epoch::default_collector().clone(), FooComparator);
```
taiki-e, line 969: This seems like an unfortunate increase in complexity for users who want to make stateless comparisons.

matthew-mcallister: It's about a dozen lines of new boilerplate. Regardless, it would be very easy to support `Equivalent`/`Comparable` on the default comparator. Rename `OrdComparator` to `BasicComparator` and define it like this:
```rust
struct BasicComparator;

impl<K: ?Sized, Q: ?Sized> Equivalator<K, Q> for BasicComparator
where
    K: Equivalent<Q>,
{
    #[inline]
    fn equivalent(&self, lhs: &K, rhs: &Q) -> bool {
        <K as Equivalent<Q>>::equivalent(lhs, rhs)
    }
}

impl<K: ?Sized, Q: ?Sized> Comparator<K, Q> for BasicComparator
where
    K: Comparable<Q>,
{
    #[inline]
    fn compare(&self, lhs: &K, rhs: &Q) -> Ordering {
        <K as Comparable<Q>>::compare(lhs, rhs)
    }
}
```

In some sense this is the best of both worlds. The only downside would be that there are now two sets of traits (`Equivalent`/`Comparable`, `Equivalator`/`Comparator`), which might be mildly confusing at first sight.

al8n: IMO, two sets of traits are reasonable. Actually, it is just one set of traits for stateless comparison, because, for stateless comparison, users just use the `BasicComparator` (the default one) and only need to implement `Equivalent` and `Comparable`. They do not need to care about `Equivalentor` and `Comparator`.

matthew-mcallister:
Restored the `Equivalent` and `Comparable` traits. Now there is no change in behavior compared to master unless you explicitly define your own comparator type.

crossbeam-skiplist/src/map.rs
```diff
@@ -116,11 +134,99 @@ where
...
+    pub fn get_or_insert_with<F>(&self, key: K, value_fn: F) -> Entry<'_, K, V, C>
+    where
+        F: FnOnce() -> V,
+    {
+        let guard = &epoch::pin();
+        Entry::new(self.inner.get_or_insert_with(key, value_fn, guard))
+    }
+
+    /// Returns an iterator over all entries in the map,
+    /// sorted by key.
```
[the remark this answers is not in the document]
matthew-mcallister: Should also update `SkipSet` and change all method docs to use the term "equivalent" (instead of equal).

PR #1184, opened 2025-03-12
replace int_to_ptr_with_provenance by map_addr-style function

RalfJung:
When https://github.com/crossbeam-rs/crossbeam/pull/796 was added, `int_to_ptr_with_provenance` was a new idea and looked like a hack. Now that strict provenance is stable, this is called `with_addr` and looks like a very normal API. ;)  And in fact, there's an even nicer API on top of it, `map_addr`. We can't use it here yet since the MSRV of Rust 1.61 does not have the strict provenance API, but we can still update code and comments to match the state of the art and to make it easier to eventually use the standard APIs.


taiki-e (approved):
Thanks!

RalfJung:
That clippy failure is in a file I did not even change... seems like clippy isn't pinned so PRs can fail unrelated reasons?

PR #1187, opened 2025-04-08
crossbeam-channel: prevent double free on `Drop`

petrosagg:
This PR is fixing a regression introduced by #1084 that can lead to a double free when dropping the channel.

The method `Channel::discard_all_messages` has the property that if it observes `head.block` pointing to a non-null pointer it will attempt to free it.

The same property holds for the `Channel::drop` method and so it is critical that whenever `head.block` is freed it must also be set to a null pointer so that it is freed exactly once.

Before #1084 the behavior of `discard_all_messages` ensured `head.block` was `null` after its execution due to the atomic store right before exiting [1].

After #1084 `discard_all_messages` atomically swaps the current value of `head.block` with a null pointer at the moment the value is read instead of waiting for the end of the function.

The problem lies in the fact that `dicard_all_messages` contained two paths that could lead to `head.block` being read but only one of them would swap the value. This meant that `dicard_all_messages` could end up observing a non-null block pointer (and therefore attempting to free it) without setting `head.block` to null. This would then lead to `Channel::drop` making a second attempt at dropping the same pointer.

The bug is similar to the one previously fixed by #972 and the double free can be reproduced by reverting the reproduction commit from that PR [2].

As with #972 it is quite difficult to trigger this bug without introducing artificial sleeps in critical points so this PR does not include a test.

[1] https://github.com/crossbeam-rs/crossbeam/blob/crossbeam-channel-0.5.11/crossbeam-channel/src/flavors/list.rs#L625
[2] https://github.com/crossbeam-rs/crossbeam/pull/972/commits/2d2262823de47492fdfa9555a47cfd999772f396

### DIFF
--- a/crossbeam-channel/src/flavors/list.rs
+++ b/crossbeam-channel/src/flavors/list.rs
@@ -613,7 +613,7 @@ impl<T> Channel<T> {
             // In that case, just wait until it gets initialized.
             while block.is_null() {
                 backoff.snooze();
-                block = self.head.block.load(Ordering::Acquire);
+                block = self.head.block.swap(ptr::null_mut(), Ordering::AcqRel);
             }
         }
 


taiki-e (approved):
Thanks!

taiki-e:
Published in crossbeam-channel 0.5.15.

ijackson:
FTR, I'm preparing a RUSTSEC advisory.

djc:
@taiki-e I mentioned you in the [advisory PR](https://github.com/rustsec/advisory-db/pull/2277) because we like to get sign-off from a maintainer before merging an advisory. Not sure if you've seen it?

taiki-e:
Approved.

> we like to get sign-off from a maintainer before merging an advisory

By the way, I love this. If NVD CPE team had such a policy, we would not have had problems like https://github.com/github/advisory-database/issues/5064#issuecomment-2761833506...

mfindra:
Hello, is there a CVE id reserved for this flaw?

taiki-e:
No. In the past we indeed have used a workflow to assign CVEs via GitHub Advisory, but **all** CVEs for this repository are in fact incorrectly assigned and only help the detection tools to detect false positives.

Furthermore, NVD CPE team who made incorrect assignments has ignored requests for fixes from the maintainer and GitHub has done nothing, saying it is out of scope, so until they address the issue I will not be submitting an advisory by any way other than RUSTSEC. (Even if submitted, it will be assigned incorrectly and will not reach the people who actually need it.)

See https://github.com/github/advisory-database/issues/5064 for more info.

PR #1191, opened 2025-04-18
add push_mut and pop_mut for queues

### LINKED ISSUE #1179: push_mut and pop_mut for queues
It would be nice if you could avoid the extra cost of atomic operations + the extra checks done for the lock-free algorithms when you have a mutable reference to the queue (for example during single-threaded initialization of a queue which is then passed off to workers).

jafar75:
Closes #1179 
This PR attempts to provide a simplified version of push and pop for crossbeam-queue data structures. By having an exclusive reference, there is no need to do CAS operations and some other checks.

- compare_exchange operations are removed
- all atomic load and store operations are done with Relaxed ordering.
- No need for backoff


jafar75:
About destroying block inside `pop_mut`, I think it is possible to do as below:

```
if offset + 1 == BLOCK_CAP {
    for i in 0..BLOCK_CAP - 1 {
        let slot = (*block).slots.get_unchecked(i);
        slot.state.fetch_or(DESTROY, Ordering::Relaxed);
    }
    drop(Box::from_raw(block));
} else {
    slot.state.fetch_or(READ, Ordering::Relaxed);
}
```

But I don't sure it is better than using `Block::destroy()` or not.

taiki-e (changes requested):
Thanks! Could you use `get_mut` instead of `load(Relaxed)`/`store(Relaxed)`? 

Atomics are known to prevent some optimizations even with relaxed ordering. e.g, https://godbolt.org/z/WaK57sec9

crossbeam-queue/src/seg_queue.rs
```diff
@@ -387,6 +439,79 @@ impl<T> SegQueue<T> {
...
+                new_head |= HAS_NEXT;
+            }
+        }
+
+        self.head.index.store(new_head, Ordering::Relaxed);
+
+        unsafe {
+            // If we've reached the end of the block, move to the next one.
+            if offset + 1 == BLOCK_CAP {
+                let next = (*block).wait_next();
```
taiki-e: I think this can simply be replaced with `get_mut` as `&mut self` means there are no other threads to wait for.

taiki-e (approved):

taiki-e:
Published in crossbeam-queue 0.3.13.

Issue #752, opened 2021-11-10
About the Windows environment： no_atomic.rs file


rockit-ba:
On Windows, the contents of the no_atomic.rs file '.. /no_atomic.rs' cannot compile. how i do?

taiki-e:
You probably need to enable `core.symlinks` in git. (e.g., `git config --global core.symlinks true`)

(If you are using a crate uploaded to crates.io, the symlink should have been converted to a regular file, so I don't think you will have this problem.)

rockit-ba:
yes，you are right

PR #1192, opened 2025-04-24
please help compiling on windows

jmikedupont2:
Hi all, I had a bunch of errors like 
```
 --> crossbeam-epoch\src\sync\once_lock.rs:1:1
  |
1 | ../../../crossbeam-utils/src/sync/once_lock.rs
  | ^^ expected item

```

Assuming they are some autoconfig that that failed on windows,
Here are my patches to compile, a hack. let me know if you know how to fix this
thanks,
mike

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,14 @@
+{
+    "files.exclude": {
+        "**/*.vo": true,
+        "**/*.vok": true,
+        "**/*.vos": true,
+        "**/*.aux": true,
+        "**/*.glob": true,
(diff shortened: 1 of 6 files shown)


taiki-e:
Duplicate of https://github.com/crossbeam-rs/crossbeam/issues/752

> You probably need to enable `core.symlinks` in git. (e.g., `git config --global core.symlinks true`)
>
> (If you are using a crate uploaded to crates.io, the symlink should have been converted to a regular file, so I don't think you will have this problem.)

PR #1195, opened 2025-06-04
Move WaitGroup count outside Mutex

orlp:
By moving the `WaitGroup` counter outside of the `Mutex` the `Clone` and `Drop` implementations no longer have to lock at all except for the final `Drop` which wakes the others.


orlp:
CI failures seem unrelated?

taiki-e (approved):
Thanks!

taiki-e:
Published in crossbeam-utils 0.8.22.

PR #1196, opened 2025-06-07
Fix CI failure

taiki-e:
The current code is not a problem with TB, the problem with SB is known and not caused by the one pointed by this lint and already has a PR to resolve it.


0xllx0:
Looks like there needs to be a MSRV bump to `1.74`, also.

PR #1197, opened 2025-06-08
fixes #946 -- return both the old and new values in `Atomic::compare_exchange`

### LINKED ISSUE #946: Confusing `compare_exchange` api in `crossbeam_epoch::Atomic`
The current api for `compare_exchange` in `crossbeam_epoch::Atomic` is confusing because it differs from other implementations of compare-and-swap operations. Traditionally, on success, we expect the previously stored value to be returned. This is not the case with the current implementation.

`crossbeam_epoch` api (from the [docs](https://docs.rs/crossbeam-epoch/latest/crossbeam_epoch/struct.Atomic.html#method.compare_exchange)):

> The return value is a result indicating whether the new pointer was written. On success the pointer that was written is returned. On failure the actual current value and new are returned.

`std` api (from the [docs](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicPtr.html#method.compare_exchange))

> The return value is a result indicating whether the new value was written and containing the previous value. On success this value is guaranteed to be equal to current.

alex:


alex:
I decided returning both made more sense than trying to just mirror the stdlib's API, because of the extreme risk existing code wouldn't notice the behavior change unless types also changed.

alex:
the CI failures appear to be the ones fixed by #1196. I can rebase after that merges.

taiki-e (approved):
Thanks! It's interesting that this also revealed mistakes in our code (fetch_update) and example (compare_exchange_weak).

alex:
Thanks!

PR #1201, opened 2025-08-29
epoch: Allow deferred destruction of unsized types

### LINKED ISSUE #1200: epoch: `defer_destroy` cannot be used with unsized types due to missing `?Sized` bound
`crossbeam-epoch` supports unsized types through the `Pointable` trait. This allows for the creation of pointers to dynamically sized types, e.g., `Shared<[MaybeUninit<T>]>`.

However, it's not possible to pass these `Shared` pointers of unsized types to `Guard::defer_destroy`, as the function is missing a `?Sized` bound on its generic type `T`. For example:

```rust
use std::mem::MaybeUninit;
use crossbeam::epoch::{Owned, pin};

fn main() {
    let guard = &pin();
    let ptr = Owned::<[MaybeUninit<usize>]>::init(42).into_shared(guard);

    unsafe {
        guard.defer_destroy(ptr); // compile error
        guard.defer_unchecked(|| ptr.into_owned()); // OK
    }
}
```

`guard.defer_destroy` generates a compile error:

```
error[E0277]: the size for values of type `[MaybeUninit<usize>]` cannot be known at compilation time
   --> src/main.rs:9:29
    |
9   |         guard.defer_destroy(ptr); // compile error
    |               ------------- ^^^ doesn't have a size known at compile-time
    |               |
    |               required by a bound introduced by this call
    |
    = help: the trait `Sized` is not implemented for `[MaybeUninit<usize>]`
note: required by an implicit `Sized` bound in `Guard::defer_destroy`
   --> /home/hyeon/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/crossbeam-epoch-0.9.18/src/guard.rs:268:33
    |
268 |     pub unsafe fn defer_destroy<T>(&self, ptr: Shared<'_, T>) {
    |                                 ^ required by the implicit `Sized` requirement on this type parameter in `Guard::defer_destroy`
```

This is because the signature for `defer_destroy` does not include `T: ?Sized`, causing Rust to add an implicit `T: Sized` bound by default.

This seems like a mistake, as the deallocation for `Shared` pointers is handled by the `Pointable::into_owned`, which does not depends on whether they are sized or not.

kim-jhyeon:
Fixes #1200

### DIFF
--- a/crossbeam-epoch/src/guard.rs
+++ b/crossbeam-epoch/src/guard.rs
@@ -1,7 +1,7 @@
 use core::fmt;
 use core::mem;
 
-use crate::atomic::Shared;
+use crate::atomic::{Pointable, Shared};
 use crate::collector::Collector;
 use crate::deferred::Deferred;
 use crate::internal::Local;
@@ -267,7 +267,7 @@ impl Guard {
     /// }
     /// # unsafe { drop(a.into_owned()); } // avoid leak
     /// ```
-    pub unsafe fn defer_destroy<T>(&self, ptr: Shared<'_, T>) {
+    pub unsafe fn defer_destroy<T: ?Sized + Pointable>(&self, ptr: Shared<'_, T>) {
         unsafe { self.defer_unchecked(move || ptr.into_owned()) }
     }
 


taiki-e (approved):
Thanks!

taiki-e:
Published in crossbeam-epoch 0.9.19.

PR #1204, opened 2025-10-06
array_queue: add core implementation

ghost:
Adds a `no_std`, `no_alloc` implementation of the `ArrayQueue`.


0xllx0:
After porting the `ArrayQueue` implementation to `heapless`, I thought upstream might also want a `no-alloc` implementation.

Most of the diffs are code motion of moving the existing `alloc` implementation into submodules (for `src` + `tests`). The rest is the `core` implementation, which is mostly a copy of the original, with some tweaks to remove the `alloc` dependency.

PR #1206, opened 2025-10-18
Port Golang select3.go Test

shota-tsuji:
Port of the [select3.go](https://github.com/golang/go/blob/master/test/chan/select3.go) test.

Part of https://github.com/crossbeam-rs/crossbeam/issues/201

Some tests are omitted:
- Empty `select` statements (they cause a compile error)
- Receiving with the `ok` flag
- Panic on select send to a closed channel (it blocks instead)

### DIFF
--- a/crossbeam-channel/tests/golang.rs
+++ b/crossbeam-channel/tests/golang.rs
@@ -1011,21 +1011,6 @@ mod select3 {
             }
         });
     }
-
-    //#[test]
-    fn main4() {
-        // selects with closed channels behave like ordinary operations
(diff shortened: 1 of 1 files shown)


taiki-e (changes requested):
Thanks! Could you resolve CI failures?

As for Miri errors, I think it can be fixed by avoiding leak checks as follows.

https://github.com/crossbeam-rs/crossbeam/blob/7a9c10de17f165d0a4ae842e55198cf0adaa8eb7/crossbeam-channel/tests/golang.rs#L1107-L1112

PR #1210, opened 2025-12-06
Implement `Unparker::unpark_with_result()`

### LINKED ISSUE #1209: Having a version of `Unpark::unpark()` that returns a status
Hi `crossbeam-rs` maintainers,

It is sometimes useful to know the status of an unpark operation, for example in a `rayon`-style thread pool where we need to heuristically wake a thread for incoming tasks:

```rs
// We have a new task, so we call this function to wake a thread to do the task if appropriate.
fn wake_one(&self) {
  let (idle_count, asleep_count) = self.counter.get();
  // If there are no awake idle threads to do the task, but there are asleep ones, wake one up.
  if idle_count == 0 && asleep_count > 0 {
    for unparker in self.unparkers {
      let old_state = unparker.unpark_returning(); // Imaginary API
      if old_state == ParkState::Parked { break; }
    }
  }
}
```

Having an `Unparker::unpark_returning() -> ParkState` here would be helpful since it avoids the alternatives:
- Having to call `unpark()` on all threads, introducing more wakeups and swaps than necessary.
- Have some other atomic variable external to `Unparker` that tracks sleep states, complicating the code.

coclique:
Fixes #1209.

`Unparker::unpark_with_result()` is like `Unpark::unpark()`, but also returns the previous state of the token. This change also introduces one new public enum `UnparkResult`.

Some other questions to think about:

- This is deliberately a non-breaking change. If we think that in the long term, it is better to migrate `unpark` to returning a result altogether, then making that breaking change here and now could be *less* painful than adding an alternative method, migrate the main method, and finally deprecate the alternative method, which is *two* breaking changes.
- Naming: `unpark_with_result` sounds a bit awkward. I can think of `unpark_returning` as an alternative, but it sounds as awkward.

### DIFF
--- a/crossbeam-utils/src/sync/mod.rs
+++ b/crossbeam-utils/src/sync/mod.rs
@@ -11,7 +11,7 @@ mod parker;
 mod sharded_lock;
 mod wait_group;
 
-pub use self::parker::{Parker, UnparkReason, Unparker};
+pub use self::parker::{Parker, UnparkReason, UnparkResult, Unparker};
 #[cfg(not(crossbeam_loom))]
 pub use self::sharded_lock::{ShardedLock, ShardedLockReadGuard, ShardedLockWriteGuard};
 pub use self::wait_group::WaitGroup;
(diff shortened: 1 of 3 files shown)


taiki-e (changes requested):
Thanks! I would like to implement this as a change to unpark. (Even if we avoid making this a breaking change here, I have no plan to backport it at this time.)

PR #1216, opened 2026-01-06
Replace iterative capacity growth with `next_power_of_two`

SaiCharanMarrivada:
Replace the capacity-doubling [loop](https://github.com/crossbeam-rs/crossbeam/blob/master/crossbeam-deque/src/deque.rs#L337-L340) with `(reserve_cap + len).next_power_of_two()`. Given the existing invariant that capacity is a power of two, the result is identical.

### DIFF
--- a/crossbeam-deque/src/deque.rs
+++ b/crossbeam-deque/src/deque.rs
@@ -333,7 +333,7 @@ impl<T> Worker<T> {
 
             // Is there enough capacity to push `reserve_cap` tasks?
             if cap - len < reserve_cap {
-                // Keep doubling the capacity as much as is needed.
+                // Ensure capacity for reserve_cap + len, rounded up to the next power of 2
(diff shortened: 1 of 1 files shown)


SaiCharanMarrivada:
To ensure that replacing the iterative capacity growth with `next_power_of_two()` maintains exact semantic parity and handles edge cases correctly, I have verified the transformation using **Alive2**.

* **LLVM-IR Generation:** I’ve isolated the logic in [Compiler Explorer](https://godbolt.org/z/9h8P14b1c) to generate the IR for both the original loop and the new implementation.
* **Alive2 Proof:** You can view the formal equivalence proof here: [Alive2 Verification](https://alive2.llvm.org/ce/z/Z7ZA39).

**Note on Verification Scope:**
I used `u32` for the proof to ensure the solver completes in a reasonable time.

taiki-e (approved):
Thanks!

PR #1241, opened 2026-02-23
Add watch channel with latest-value semantics

### LINKED ISSUE #942: Channel that only stores latest message, blocking on Recv, but not Send
Is it possible to implement a channel that only stores only the latest value and returns the owned latest value on `.recv()`? This would be roughly similar to:

* The async [tokio::sync::watch](https://docs.rs/tokio/latest/tokio/sync/watch/index.html), but would be owned rather than borrowed and non-async
* `bounded(1)`, except senders would never block, since sent messages would just replace the latest message, if there is one
* `(Mutex<T>, CondVar)`, except senders and receivers are separate like they are in channels, so senders cannot read the value, etc

## Current Solution

```rust
// A similar effect can be achieved by draining a bounded/unbounded channel, but this channel would be more efficient
while let Ok(mut msg) = chan.recv() {
  // Drain all channel messages, and keep only the last
  while let Ok(m) = chan.try_recv() { msg = m; }

  // Do work with the latest message `msg`
}
```

## Example API

```rust
// Some new channel type
let (tx, mut rx) = unbounded_watch_channel::<usize>();

// Set latest value to 0.
tx.send(0);
// Set latest value to 1. This won't block since it's updating the latest value only (would deadlock if using bounded(1))
tx.send(1);

assert_eq!(rx.recv(), Ok(1));

std::thread::spawn(move || {
  tx.send(2);
});

// This will block until the thread above sends 2
assert_eq!(rx.recv(), Ok(2));

// This will sit and wait forever (unlike the tokio::sync::watch api, which does not move values)
x.recv();
```

## Example use-case

If doing a non-interruptable refresh task, and `n>1` `RefreshRequest` messages arrive in the channel during the refresh, the refresh task only runs once more instead of `n` times more.

If this is something that could conceivably be added to crossbeam, I could try starting development on it.

veeceey:
Adds a `watch` channel type that stores only the latest value. Sending never blocks (overwrites the previous value), and receiving blocks until a new value is available.

This is useful for scenarios like refresh signals where you only care about the most recent value - if multiple sends happen while the receiver is busy, the receiver processes just one iteration instead of N.

The implementation uses `Mutex<Option<T>> + Condvar` as suggested in #942. Key points:

- `WatchSender` and `WatchReceiver` are both `Clone`
- Each receiver independently tracks which version it has seen
- Disconnection is properly detected (recv returns `RecvError` when all senders are dropped, send returns `SendError` when all receivers are dropped)
- `try_recv` is available for non-blocking checks

```rust
let (tx, mut rx) = crossbeam_channel::watch::<i32>();
tx.send(1);
tx.send(2);
assert_eq!(rx.recv(), Ok(2)); // only latest value
```

Closes #942

### DIFF
--- a/crossbeam-channel/src/lib.rs
+++ b/crossbeam-channel/src/lib.rs
@@ -364,6 +364,8 @@ mod select_macro;
 mod utils;
 #[cfg(feature = "std")]
 mod waker;
+#[cfg(feature = "std")]
+mod watch;
 
 /// Crate internals used by the `select!` macro.
 #[doc(hidden)]
@@ -384,4 +386,5 @@ pub use crate::{
         SendTimeoutError, TryReadyError, TryRecvError, TrySelectError, TrySendError,
     },
     select::{Select, SelectedOperation},
+    watch::{watch, WatchReceiver, WatchSender},
 };
--- a/crossbeam-channel/src/lib.rs
+++ b/crossbeam-channel/src/lib.rs
@@ -386,5 +386,5 @@ pub use crate::{
         SendTimeoutError, TryReadyError, TryRecvError, TrySelectError, TrySendError,
     },
     select::{Select, SelectedOperation},
-    watch::{watch, WatchReceiver, WatchSender},
+    watch::{WatchReceiver, WatchSender, watch},
 };
--- a/crossbeam-channel/src/watch.rs
+++ b/crossbeam-channel/src/watch.rs
@@ -1,0 +1,215 @@
+//! A watch channel that only stores the latest value.
+//!
+//! Sending never blocks (it overwrites the latest value), while receiving
+//! blocks until a new value is available.
+
+use alloc::sync::Arc;
+use core::fmt;
+use std::sync::{Condvar, Mutex};
+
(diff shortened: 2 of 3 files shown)


veeceey:
Test results:

```
running 12 tests
test watch_clone_receiver ... ok
test watch_clone_sender ... ok
test watch_multiple_sends_between_recvs ... ok
test watch_overwrites_value ... ok
test watch_recv_error_on_disconnect ... ok
test watch_send_error_when_no_receivers ... ok
test watch_send_recv_basic ... ok
test watch_multithreaded ... ok
test watch_try_recv_disconnected ... ok
test watch_try_recv_empty ... ok
test watch_try_recv_value ... ok
test watch_recv_blocks_until_send ... ok

test result: ok. 12 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

Full existing test suite also passes (74 doc+unit tests).

veeceey:
checking in on this - the watch channel is a pretty commonly requested primitive. open to design feedback

veeceey:
friendly bump on this - happy to make any design changes if needed

PR #1245, opened 2026-04-05
Updated recv_deadline documentation

### LINKED ISSUE #975: recv_deadline with past instant is ambiguous
Hello,

Thanks for your work on this crate.
I find `recv_deadline` on a channel a bit ambiguous when we pass an `Instant` that is in the past. This behavior isn't documented and, in my first intuition was: it will not even try to `recv` as the deadline is already pass. After a test it shows me that even if the deadline is in the past but there is already a message in the channel he will be received.

Do you think the behavior should change or is it possible to document it ?

ErisianArchitect:
I noticed that the behavior of `recv_deadline` is undocumented for the edge case where the channel is non-empty and the deadline has already been reached already. It has a bit of an unexpected behavior in that it ignores the deadline and returns the next message in the channel. I assume this behavior is intentional since you can check the deadline upstream, so I have updated the documentation to reflect the behavior.

This is related to issue #975 

Closes #975

### DIFF
--- a/crossbeam-channel/src/channel.rs
+++ b/crossbeam-channel/src/channel.rs
@@ -904,7 +904,9 @@ impl<T> Receiver<T> {
     ///
     /// If the channel is empty and not disconnected, this call will block until the receive
     /// operation can proceed or the operation times out. If the channel is empty and becomes
-    /// disconnected, this call will wake up and return an error.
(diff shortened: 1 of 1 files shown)


taiki-e (approved):
Thanks.

PR #1247, opened 2026-04-14
Added lossy channel variety based on bounded

joe-saronic:
Full bounded channels drop the newest item instead of the oldest when used in non-blocking mode.

This adds a `lossy_send` method that pushes out the oldest elements. Semantics are defined for all the existing types of queues. I'm not re-using the existing error semantics for `try_send` to avoid breaking existing code or creating a breaking or non-additive change to the API.

A `lossy` convenience constructor is provided to check that capacity is non-zero and immediately delegate to `bounded`.

Unit tests are made to provide similar coverage to the original `bounded` queue as well as verify basic semantics for non-`bounded` queues.

Benchmarks show that the lossy version is much faster than regular `bounded` because it never has to wait:

```
test lossy::create        ... bench:         122.43 ns/iter (+/- 8.14)
test lossy::inout         ... bench:          16.72 ns/iter (+/- 2.48)
test lossy::mpmc          ... bench:   1,076,605.90 ns/iter (+/- 331,346.17)
test lossy::mpsc          ... bench:   6,635,463.50 ns/iter (+/- 2,494,536.68)
test lossy::spsc          ... bench:     600,372.20 ns/iter (+/- 309,960.31)
                                                                                                                        
test bounded_1::create    ... bench:         125.86 ns/iter (+/- 5.14)
test bounded_1::mpmc      ... bench:  17,134,194.30 ns/iter (+/- 5,030,452.40)
test bounded_1::mpsc      ... bench:  86,669,462.20 ns/iter (+/- 11,068,690.83)
test bounded_1::oneshot   ... bench:         147.53 ns/iter (+/- 17.90)
test bounded_1::spmc      ... bench:  87,127,345.00 ns/iter (+/- 14,460,539.76)
test bounded_1::spsc      ... bench:   9,895,681.25 ns/iter (+/- 1,179,280.38)
```

Unbounded is a bit faster for MPMC, much faster for MPSC, but slower for everything else that involves potential buffer resizing but does not involve eviction:

```
test unbounded::create    ... bench:         107.42 ns/iter (+/- 15.82)
test unbounded::inout     ... bench:          31.88 ns/iter (+/- 4.13)
test unbounded::mpmc      ... bench:     870,335.36 ns/iter (+/- 385,854.63)
test unbounded::mpsc      ... bench:   2,846,018.20 ns/iter (+/- 1,642,916.66)
test unbounded::oneshot   ... bench:         172.70 ns/iter (+/- 40.10)
test unbounded::par_inout ... bench:   3,990,006.80 ns/iter (+/- 1,924,910.08)
test unbounded::spmc      ... bench:   4,544,291.70 ns/iter (+/- 13,056,468.18)
test unbounded::spsc      ... bench:   1,533,016.00 ns/iter (+/- 926,932.93)
```

Summary:

```
                  lossy          bounded_1       unbounded
create            122 ns         126 ns          107 ns
inout              17 ns         —                32 ns
spsc              600 µs         9,896 µs        1,533 µs
mpsc            6,635 µs        86,669 µs        2,846 µs
mpmc            1,077 µs        17,134 µs          870 µs
```

### DIFF
--- a/crossbeam-channel/src/flavors/array.rs
+++ b/crossbeam-channel/src/flavors/array.rs
@@ -21,7 +21,7 @@ use crossbeam_utils::{Backoff, CachePadded};
 
 use crate::{
     context::Context,
-    err::{RecvTimeoutError, SendTimeoutError, TryRecvError, TrySendError},
+    err::{LossySendError, RecvTimeoutError, SendTimeoutError, TryRecvError, TrySendError},
     select::{Operation, SelectHandle, Selected, Token},
     waker::SyncWaker,
 };
@@ -89,23 +89,11 @@ pub(crate) struct Channel<T> {
 
     /// Receivers waiting while the channel is empty and not disconnected.
     receivers: SyncWaker,
-
-    /// If true, `try_send` on a full channel evicts the oldest message instead of dropping the incoming one.
-    lossy: bool,
 }
 
 impl<T> Channel<T> {
     /// Creates a bounded channel of capacity `cap`.
     pub(crate) fn with_capacity(cap: usize) -> Self {
-        Self::new(cap, false)
-    }
-
-    /// Creates a lossy bounded channel of capacity `cap`.
-    pub(crate) fn with_capacity_lossy(cap: usize) -> Self {
-        Self::new(cap, true)
-    }
-
-    fn new(cap: usize, lossy: bool) -> Self {
         assert!(cap > 0, "capacity must be positive");
 
         // Compute constants `mark_bit` and `one_lap`.
@@ -137,7 +125,6 @@ impl<T> Channel<T> {
             tail: CachePadded::new(AtomicUsize::new(tail)),
             senders: SyncWaker::new(),
             receivers: SyncWaker::new(),
-            lossy,
         }
     }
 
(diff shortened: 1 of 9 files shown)


crossbeam-channel/src/flavors/array.rs
```diff
@@ -321,11 +334,91 @@ impl<T> Channel<T> {
...
+                Err(TrySendError::Full(msg))
+            }
+        }
+    }
+
+    /// Sends a message, evicting the oldest if the channel is full.
+    ///
+    /// Returns `Ok(())` if there was space, `Err(TrySendError::Full(evicted))` if the oldest
+    /// message was displaced, or `Err(TrySendError::Disconnected(msg))` if disconnected.
+    fn force_send(&self, msg: T) -> Result<(), TrySendError<T>> {
```
sid-baral: @joe-saronic TrySendError::Full semantic is getting changed here — Standard crossbeam: TrySendError::Full(T) means the item T we tried to send was rejected because the channel was full. but lossy channel TrySendError::Full(T) means an old item T was evicted to make room for the new item which was actually sent successfully.
This is kind of dangerous semantic flip, the error variant now contains the evicted item instead of the rejected one. Any code pattern matching on TrySendError::Full expecting to handle the rejected item will silently process the wrong data.

joe-saronic: @sid-baral This is a good point. The current approach is a bit of a shortcut. My thought process here was that this is justifiable because whoever opened the channel deliberately called `lossy` to do so. At the same time, maybe they did it in a conditional or so, where the result is not always a lossy channel.

One alternative is to add another enum variant, say `Evicted`. It would be more proper, but the downside is that any existing code with an exhaustive match would break in the new version. I am not convinced that this change is worth a major bump, but perhaps.

Yet another alternative is to add an entirely new flavor of channel and try to reuse most of the code in `array`. The return value would be its own enum at that point (`LossySendError`?), and would not break existing code. This would however require a larger refactor.

The functionality in all three cases would be identical. Personally, I am leaning to the third option, but happy to hear suggestions and advice.

sid-baral:
"Yet another alternative is to add an entirely new flavor of channel" - I like that.

crossbeam-skiplist/tests/base.rs
joe-saronic, line 1: This is clippy fixes that are unrelated to my code, but are required to make CI pass. They are minor enough not to warrant a separate PR IMO

joe-saronic:
> "Yet another alternative is to add an entirely new flavor of channel" - I like that.

Implemented. It actually turned out to be the best balance of an interface that captures solutions to all the issues we discussed without breaking backwards compatibility in existing features in any way. I'm not sure it even took more code to do. Most of this PR by LOC is unit and benchmark tests.

joe-saronic:
Related to https://github.com/crossbeam-rs/crossbeam/issues/400, but added the flavor.

joe-saronic:
Closing this down:

1. I did not have the context of #400 when I first opened this, so the approach of adding a new channel is exactly what the maintainers did not want to do. This is entirely my failure.
2. Since #789 has merged, the functionality already present in a much more ergonomic form.
3. Pretty sure there is an actual synchronization bug in my code since the check and write are not surrounded by an encompassing lock.

PR #1250, opened 2026-04-18
Make `never()` a `const` function

Cyberboss:
This is useful to allow the following scenario which is not possible today:
```
static REQUEST_SYSTEM: Mutex<Receiver<Model>> = Mutex::new(crossbeam_channel::never());
```


taiki-e (approved):
Thanks.

taiki-e:
Published in crossbeam-channel 0.5.16.

PR #1252, opened 2026-05-04
skiplist: prevent Range iterator from rewinding after exhaustion

### LINKED ISSUE #1142: `crossbeam_skiplist::Range` rewinds to the head after it was exhausted
Hi, I found that the implementation of the `Iterator` trait for `Range` will rewind to the head after it is exhausted. The test case below can reproduce this behavior. Is it an expected behavior or should this be a bug?

```rust
#[test]
fn range() {
    use crate::Bound::*;
    let guard = &epoch::pin();
    let s = SkipList::new(epoch::default_collector().clone());
    let v = (0..10).map(|x| x * 10).collect::<Vec<_>>();
    for &x in v.iter() {
        s.insert(x, x, guard).release(guard);
    }

    let mut it = s.range(..=5, guard);
    for _ent in &mut it {}

    let ent = it.next();
    assert!(ent.is_none(), "{ent:?}"); // panic, but why? it should be none, but yield the first element Entry(0, 0) again
}
```

cong-or:
Fixes #1142 — Range iterators rewound to the first element after exhaustion.                                                                                                                
                                                                
  When the bounds check failed, `head` was set to `None`. Next call interpreted                                                                                                               
  that as "haven't started yet" and searched from `start_bound()` again.                                                                                                                      
                                                                                                                                                                                              
  Added a `finished` flag to `Range`/`RefRange`, set only when bounds are          
  exceeded. Iter/RefIter left unchanged — their rewind on concurrent inserts
  is intentional.

  Also added `FusedIterator` impls for Range in base, map, and set.

### DIFF
--- a/crossbeam-skiplist/src/base.rs
+++ b/crossbeam-skiplist/src/base.rs
@@ -4,6 +4,7 @@ use alloc::alloc::handle_alloc_error;
 use core::{
     alloc::Layout,
     cmp, fmt,
+    iter::FusedIterator,
     marker::PhantomData,
     mem,
     ops::{Bound, Deref, RangeBounds},
@@ -684,6 +685,7 @@ where
             range,
             guard,
             _marker: PhantomData,
+            finished: false,
         }
     }
 
@@ -701,6 +703,7 @@ where
             head: None,
             tail: None,
             _marker: PhantomData,
+            finished: false,
         }
     }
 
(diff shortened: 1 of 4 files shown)


crossbeam-skiplist/src/base.rs
```diff
@@ -1988,6 +1991,7 @@ where
     range: R,
     guard: &'g Guard,
     _marker: PhantomData<fn() -> Q>, // covariant over `Q`
+    finished: bool,
```
taiki-e, line 1994: Hmm. Unless it's necessary for correctness, I think it's generally preferable to use fuse rather than "adding an extra flag just to provide FusedIterator's behavior".

PR #1267, opened 2026-06-13
Add cargo-fuzz target for OSS-Fuzz integration

canolgun:
Adds cargo-fuzz target for continuous fuzzing via OSS-Fuzz.

### DIFF
--- a/fuzz/fuzz_targets/fuzz_parse.rs
+++ b/fuzz/fuzz_targets/fuzz_parse.rs
@@ -1,0 +1,6 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+fuzz_target!(|data: &[u8]| {
+    let s = std::str::from_utf8(data).unwrap_or("");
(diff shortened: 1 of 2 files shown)


fuzz/fuzz_targets/fuzz_parse.rs
```diff
@@ -0,0 +1,6 @@
+#![no_main]
+use libfuzzer_sys::fuzz_target;
+fuzz_target!(|data: &[u8]| {
+    let s = std::str::from_utf8(data).unwrap_or("");
+    let _ = s.parse::<String>();
+});
```
taiki-e, line 6: This is a fuzzing for std::str::from_utf8, not crossbeam's API.

taiki-e (commented):
Closing per https://github.com/google/oss-fuzz/pull/15743#issuecomment-4699878530. I welcome PRs that actually fuzzing our code, but this and your other PRs are nothing like that.

PR #1270, opened 2026-06-21
make CachePadded<T> have C repr to allow casting to and from T

nahla-nee:
While working on a current project which uses CachePadded, I noticed that the struct doesn't have a C repr. My project casts pointers to CachePadded<T> to T pretty frequently and I think many others assume that they can do the same. To make matters worse, this isn't something that Miri catches. While it is unlikely to cause problems, I believe that it can as by default the internal layout of any struct is undefined. This change makes CachePadded have a C repr so that casting pointers to T can work as expected.

### DIFF
--- a/crossbeam-utils/src/cache_padded.rs
+++ b/crossbeam-utils/src/cache_padded.rs
@@ -107,19 +107,19 @@ use core::{
         target_arch = "sparc",
         target_arch = "hexagon",
     ),
-    repr(align(32), C)
+    repr(align(32))
 )]
 // m68k has 16-byte cache line size.
 //
 // Sources:
 // - https://github.com/torvalds/linux/blob/3516bd729358a2a9b090c1905bd2a3fa926e24c6/arch/m68k/include/asm/cache.h#L9
-#[cfg_attr(target_arch = "m68k", repr(align(16), C))]
+#[cfg_attr(target_arch = "m68k", repr(align(16)))]
 // s390x has 256-byte cache line size.
 //
 // Sources:
 // - https://github.com/golang/go/blob/3dd58676054223962cd915bb0934d1f9f489d4d2/src/internal/cpu/cpu_s390x.go#L7
 // - https://github.com/torvalds/linux/blob/3516bd729358a2a9b090c1905bd2a3fa926e24c6/arch/s390/include/asm/cache.h#L13
-#[cfg_attr(target_arch = "s390x", repr(align(256), C))]
+#[cfg_attr(target_arch = "s390x", repr(align(256)))]
 // x86, wasm, riscv, and sparc64 have 64-byte cache line size.
 //
 // Sources:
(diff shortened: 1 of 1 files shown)


nahla-nee:
I'm not entirely sure what caused that failure? All tests pass for me locally and the output doesn't specify what test failed.

taiki-e (changes requested):
Thanks!

`#[repr]` can be applied separately, so I would prefer to add a single `#[repr(C)]` rather than adding `, C` to every `#[repr(align(..))]`.

(armv5te CI is currently flaky, so the failure is unrelated.)

taiki-e:
Also, could you document this guarantee?

I hope that by properly documenting guarantees when they exist, we can reduce the number of users who they rely on guarantees that are neither documented nor actually in place.

nahla-nee:
hopefully this is sufficient.

taiki-e:
Oh, sorry, I should have been a little clearer. 

"documenting guarantees" doesn't mean writing code comments; it means, as [std did](https://doc.rust-lang.org/stable/std/num/struct.Wrapping.html#layout-1), leaving doc comments about guarantees about the layout so that they are visible to users reading the documentation. 

In this case, something like:

```rs
/// # Layout
///
/// Since crossbeam-utils 0.9.0, this type is `#[repr(C)]` and is guaranteed that the pointer to `CachePadded<T>` has the same address as the pointer to the underlying `T`.
```

One more thing to add: Users should not make any assumptions about layouts that do not include such explicit guarantees.

nahla-nee:
Oh, sorry about that! I totally misunderstood what you said, I inferred wrong by looking at the comments citing alignment values haha. I added the comment you suggested.

> One more thing to add: Users should not make any assumptions about layouts that do not include such explicit guarantees.

That is very fair, it was a silly assumption I made while prototyping, but I figured it'd be useful hence this PR. Third time's the charm, but let me know if you need any other changes made.

I did keep the comment citing the standard in regards to the validity of the change. I wasn't sure if it should be removed or not.

crossbeam-utils/src/cache_padded.rs
```diff
@@ -147,6 +152,25 @@ use core::{
...
+//  field. If the current offset is not a multiple of the field’s alignment, then add padding bytes
+//  to the current offset until it is a multiple of the field’s alignment. The offset for the field
+//  is what the current offset is now. Then increase the current offset by the size of the field."
+//
+// Since the value is the first an only item in the struct, it is placed in the 0th byte, since 0 is
+// a multiple of every integer, this its alignment wouldn't need to be padded.
+//
+// Sources:
+// - https://github.com/rust-lang/reference/blob/264aec433acc7503c48a8c4a2c59075c955f3497/src/type-layout.md?plain=1#L207C1-L208C1
+//
```
taiki-e:
```suggestion
```

I think layout doc is enough.

crossbeam-utils/src/cache_padded.rs
```diff
@@ -31,6 +31,11 @@ use core::{
 ///
 /// The alignment of `CachePadded<T>` is the maximum of N bytes and the alignment of `T`.
 ///
+/// # Layout
+///
+/// Since crossbeam-utils 0.9.0, this type is `#[repr(C)]` and is guaranteed that the pointer to
```
taiki-e:
```suggestion
/// Since crossbeam-utils 0.8.22, this type is `#[repr(C)]` and is guaranteed that the pointer to
```

I will include this to https://github.com/crossbeam-rs/crossbeam/pull/1274.

taiki-e (approved):

taiki-e:
Published in crossbeam-utils 0.8.22.

PR #1279, opened 2026-07-06
Implement a recycle method for Select

CenTdemeern1:
Reduce, reuse, recycle!
This PR adds a method to crossbeam-channel's `Select` that recycles the `Select` and its internal buffer (akin to [`Vec::recycle`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.recycle)) with a new lifetime, and thus allows the referenced senders and receivers to be dropped, for example. (This is something I needed for a personal project)

First time contributing to crossbeam, please be kind, and let me know if there's anything I forgot to do.

### DIFF
--- a/crossbeam-channel/src/select.rs
+++ b/crossbeam-channel/src/select.rs
@@ -1208,6 +1208,7 @@ impl<'a> Select<'a> {
         // Replace with [`Vec::recycle`] when it's stable.
         let mut handles = mem::ManuallyDrop::new(handles);
         let (ptr, length, capacity) = (handles.as_mut_ptr(), handles.len(), handles.capacity());
+        debug_assert_eq!(length, 0);
         Select {
             // SAFETY: this vec has been cleared beforehand, so the lifetime has ended.
             handles: unsafe { Vec::from_raw_parts(ptr.cast(), length, capacity) },
--- a/crossbeam-channel/src/select.rs
+++ b/crossbeam-channel/src/select.rs
@@ -1170,6 +1170,48 @@ impl<'a> Select<'a> {
             Some(index) => Ok(index),
         }
     }
+
+    /// Recycles this `Select` and its internal allocated buffer, clearing the `Select` and giving it a new lifetime.
+    ///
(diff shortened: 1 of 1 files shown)


crossbeam-channel/src/select.rs
```diff
@@ -1170,6 +1170,51 @@ impl<'a> Select<'a> {
...
+            next_index: _,
+            biased,
+        } = self;
+        handles.clear();
+        // FIXME: Unsafe but stable version of [`Vec::recycle`].
+        // Replace with [`Vec::into_raw_parts`] when MSRV hits 1.93.
+        // Replace with [`Vec::recycle`] when it's stable.
+        let mut handles = mem::ManuallyDrop::new(handles);
+        let (ptr, length, capacity) = (handles.as_mut_ptr(), handles.len(), handles.capacity());
+        Select {
```
taiki-e, line 1212: Per https://doc.rust-lang.org/stable/src/alloc/vec/mod.rs.html#3467.

```suggestion
        debug_assert_eq!(length, 0);
        Select {
```

taiki-e (approved):
Thanks!

crossbeam-channel/src/select.rs
```diff
@@ -1170,6 +1170,52 @@ impl<'a> Select<'a> {
...
+    pub fn recycle<'b>(self) -> Select<'b> {
+        let Select {
+            mut handles,
+            next_index: _,
+            biased,
+        } = self;
+        handles.clear();
+        // FIXME: Unsafe but stable version of [`Vec::recycle`].
+        // Replace with [`Vec::into_raw_parts`] when MSRV hits 1.93.
+        // Replace with [`Vec::recycle`] when it's stable.
```
taiki-e:
```suggestion
        // FIXME: Unsafe but stable version of `Vec::recycle`.
        // Replace with `Vec::into_raw_parts` when MSRV hits 1.93.
        // Replace with `Vec::recycle` when it's stable.
```

[``] is only valid on doc comment.

PR #1306, opened 2026-08-08
tsan: Enable Adaptive Delay

taiki-e:
https://github.com/llvm/llvm-project/blob/llvmorg-23.1.0-rc2/clang/docs/ThreadSanitizer.md#adaptive-delay


taiki-e:
It seems that the workaround for the lack of support for fence by https://github.com/crossbeam-rs/crossbeam/pull/998 isn't sufficient...

PR #1307, opened 2026-08-10
docs(epoch): removed redundant safety warning details in `Shared::as_ref` documentation

HuaGu-Dragon:


taiki-e:
This doesn't seem correct.

```rs
let shared = atomic.load(Ordering::Acquire, guard);
shared.into_owned(); // drop
shared.as_ref(); // UAF but no data race
```

HuaGu-Dragon:
> This doesn't seem correct.
> 
> ```rust
> let shared = atomic.load(Ordering::Acquire, guard);
> shared.into_owned(); // drop
> shared.as_ref(); // UAF but no data race
> ```

I’m not very sure; I’m still learning this crate.

First, in the single-threaded case, `into_owned` takes ownership, so this may not compile.

Second, `into_owned` is `unsafe`, and the caller needs to make sure that the pointer is valid and that nobody else is holding a reference to the same object.
https://github.com/crossbeam-rs/crossbeam/blob/7de845aa13ce5d80084ce0889e014ae0aa0d19ce/crossbeam-epoch/src/atomic.rs#L1440

And in the multi-threaded case, does the `guard` guarantee that the data will not be dropped in the logical sense?

taiki-e:
> First, in the single-threaded case, `into_owned` takes ownership, so this may not compile.

`Shared` implements `Copy`, so that example is compilable and causes:

```
error: Undefined Behavior: constructing invalid value of type &i32: encountered a dangling reference (use-after-free)
    --> crossbeam-epoch/src/atomic.rs:1412:27
     |
1412 |             Some(unsafe { &*T::as_ptr(raw) })
     |                           ^^^^^^^^^^^^^^^^ Undefined Behavior occurred here
     |
     = help: this indicates a bug in the program: it performed an invalid operation, and caused Undefined Behavior
     = help: see https://doc.rust-lang.org/nightly/reference/behavior-considered-undefined.html for further information
     = note: this is on thread `atomic::tests::compare_exchange_success`
     = note: stack backtrace:
             0: atomic::Shared::<'_, i32>::as_ref
                 at crossbeam-epoch/src/atomic.rs:1412:27: 1412:43
             1: atomic::tests::compare_exchange_success
                 at crossbeam-epoch/src/atomic.rs:1667:13: 1667:23
             2: atomic::tests::compare_exchange_success::{closure#0}
                 at crossbeam-epoch/src/atomic.rs:1638:34: 1638:34

note: some details are omitted, run with `MIRIFLAGS=-Zmiri-backtrace=full` for a verbose backtrace

error: aborting due to 1 previous error
```

> Second, `into_owned` is `unsafe`, and the caller needs to make sure that the pointer is valid and that nobody else is holding a reference to the same object.

This condition is met at the time `into_owned` is called.

HuaGu-Dragon:
> `Shared` implements `Copy`, so that example is compilable and causes:
> This condition is met at the time `into_owned` is called.

Ohh.. this makes sense now. Sorry for my misunderstanding, and thanks for your answer.

PR #1327, opened 2026-09-07
channel: Implement Display and Error for TryReadyError and ReadyTimeoutError

MaxFreedomPollard:
`TryReadyError` and `ReadyTimeoutError` are public, re-exported from `crossbeam-channel/src/lib.rs:383`, and are what `Select::try_ready`, `Select::ready_timeout` and `Select::ready_deadline` return. They are the only error types in `crossbeam-channel/src/err.rs` that implement neither `fmt::Display` nor `std::error::Error`, so the select side works and the ready side does not:

```rust
// fine
let _: Box<dyn std::error::Error> = Box::new(crossbeam_channel::SelectTimeoutError);
// error[E0277]: the trait bound `ReadyTimeoutError: std::error::Error` is not satisfied
let _: Box<dyn std::error::Error> = Box::new(crossbeam_channel::ReadyTimeoutError);
```

In practice that means `sel.ready_timeout(d)?` does not compile in a function returning `Box<dyn Error>` or `anyhow::Result`, while `sel.select_timeout(d)?` does, and `println!("{err}")` on either ready error does not compile at all. Today only `Debug` is available, from the `derive` on the two structs.

This adds the two `Display` impls and the two empty `error::Error` impls at the end of `err.rs`, which keeps the impls in the order the types are declared. The messages are the ones already used by the select counterparts at `crossbeam-channel/src/err.rs:340-354`. `TrySelectError` and `TryReadyError` carry the same doc line, "Failed because none of the channel operations were ready.", so both say "all operations in select would block"; `SelectTimeoutError` and `ReadyTimeoutError` carry the same doc line, "Failed because none of the channel operations became ready before the timeout.", so both say "timed out waiting on select".

`error_impls` in `crossbeam-channel/tests/ready.rs` boxes each error as `Box<dyn Error>` and asserts the rendered text. It runs on one thread with no sleeps and no timing assumptions, so it is fine under Miri and on every CI target. `ready_timeout(ms(0))` reaches the deadline branch of `run_ready` deterministically, because `ready_timeout` computes the deadline as `Instant::now()` before the readiness scan and the scan is what advances the clock past it. The sender stays bound as `_s` so the channel is connected and never ready.

Verified on macOS aarch64 with cargo 1.95.0:

- Without the `err.rs` change, `cargo test -p crossbeam-channel --test ready error_impls` fails to build with two `error[E0277]`, one per type.
- With it, `test error_impls ... ok`.
- `cargo test -p crossbeam-channel`: 400 passed, 0 failed, 1 ignored across 15 test binaries, plus 74 doctests passed.
- `cargo clippy -p crossbeam-channel --lib --tests -- -D warnings`: clean. `--all-targets` cannot be used on stable here because `benches/crossbeam.rs` opens with `#![feature(test)]`; that fails identically on an unmodified checkout.
- `cargo fmt --all -- --check`: clean.

Nothing public is removed or changed, so the addition is backwards compatible.


taiki-e (approved):
Thanks!

PR #1328, opened 2026-09-15
Optimize buffer loads in Stealer::steal/steal_batch* with load_consume

mohakgoel1:
## Use `load_consume` for the buffer pointer in `Stealer::steal`/`steal_batch*`

Replaces `self.inner.buffer.load(Ordering::Acquire, guard)` with `self.inner.buffer.load_consume(guard)` at the three sites where `buffer` is immediately dereferenced (`steal`, `steal_batch_with_limit`, `steal_batch_with_limit_and_pop`). Each of these loads `buffer` and then immediately dereferences it (`buffer.deref().read(f)`) a genuine data dependency, which is exactly the case `load_consume` is for. On weak-memory architectures like AArch64 this compiles to a relaxed load plus a compiler fence instead of a hardware acquire barrier, with no effect on x86. `crossbeam-skiplist` already uses this same pattern elsewhere in the workspace (`pred.get_level(0).load_consume(guard)`); this applies it to `crossbeam-deque`'s steal path too. The buffer re-check comparisons a few lines below each of these sites stay `Acquire`, since they're equality checks, not dependent dereferences, and don't qualify.

### Correctness
- Full `crossbeam-deque` test suite (`fifo`, `lifo`, `injector`, `steal`, doctests) passes under Miri.
- `crossbeam_epoch::Atomic<T>::load_consume` is existing, documented public API this PR doesn't touch `crossbeam-epoch`.
- Consume ordering is inherently harder to tool-validate than acquire/release (Miri and loom both model it as `Acquire`, per crossbeam-utils' own comments on `AtomicConsume`), so this rests on the ordering being a strict weakening of `Acquire` at a spot with a genuine address dependency, not on tool-verified proof beyond what's achievable for consume ordering in general.

### Benchmarks
Measured on aarch64 (Graviton-class instance), 3 runs unless noted; `contended_steal_4_stealers` re-run at n=10 given its higher variance under contention.

| Benchmark | Baseline | With `load_consume` | Δ |
|---|---|---|---|
| contended_steal_4_stealers (median, n=10) | 11.59 ms | 9.43 ms | **−18.6%** |
| single_thread_push_pop (control) | 1.9403 ms | 1.9308 ms | flat, as expected |
| injector_mpmc_4x4 | 11.70 ms | 11.74 ms | flat, as expected |

### Notes
`single_thread_push_pop` never calls `Stealer`, so its flatness confirms the change is correctly scoped. `injector_mpmc_4x4` is dominated by unrelated `Injector` machinery. `contended_steal_4_stealers` directly exercises the modified functions under contention and shows a repeatable improvement once sampled enough to separate from host-level noise (raw per-run data available on request).

This PR is a joint contribution by:

- Abhishek Jain(@abhijain1204fujitsu)
- Mohak Goel(@mohakgoel1)


mohakgoel1:
Hi @taiki-e,
Could you please take a look at this PR and review it when you get a chance?

mohakgoel1:
Hi @taiki-e,
Just a gentle reminder on PR #1328. When you get a chance, could you please take a look and review it?
Thank you!

